PR #8, opened 2016-05-27
Conda recipes with linked ipopt

leethargo:
I got everything working, more or less.

Instead of reusing some BLAS and Lapack from a conda package, I now compile everything with IPOPT.

Also, `ipoptlib` is now *required* for both `scipoptlib` and (implicitely) `pyscipopt`.

The `make scipoptlib` now requires a user to press [Enter] once :sob:, but I guess that's acceptable, since the end users will use precompiled packages anyway.


leethargo:
Once this is merged, I can build new binary packages for all three.

PR #11, opened 2016-06-29
Clean up

mattmilten:
- removed the two unsupported plugin classes `reader` and `nodeselector`
- move Python base class definition above cdefs of callback wrappers
- improve `setup.py` and `conda_setup.py`


conda_setup.py
```diff
@@ -6,10 +6,9 @@
 from Cython.Build import cythonize
 
 extensions = [Extension('pyscipopt.scip', ['pyscipopt/scip.pyx'],
-                        extra_compile_args=['-UNDEBUG'],
```
leethargo: What's the reason for this? I don't remember why I put it there, but I assume that it was necessary.

setup.py
```diff
@@ -10,26 +10,26 @@
...
-usesharedlib = True
-pathToScipoptsuite = os.path.abspath('../../../')
-if usesharedlib:
-    libscipopt = 'lib/libscipopt.so'
-else:
-    libscipopt = 'lib/libscipopt.a'
-
-linkincludescip = 'lib/scip-src'
+pathToSCIPOpt = os.path.abspath('../../../')
+pathToSCIPsrc = os.path.abspath('../../src')
```
leethargo: Do these relative paths still make sense, now that PySCIPOpt is no longer in the scip repository?

mattmilten: We still want to publish the interface together with a regular SCIP
release, so these relative paths are still valid for this case.

conda_setup.py
mattmilten: I suppose this was copied from the other setup.py and there I put for debugging purposes. I don't think we want to run with all asserts and debug lines activated so I removed it.

leethargo: Sure. I thought that `UNDEBUG` would actually turn *off* the debug flags, but it seems that I'm wrong.

PR #12, opened 2016-07-05
link to lib dir and not lib itself

mattmilten:
Tthis preserves the libraries dependencies. If the libscipopt.so was built using $ORIGIN inside the rpath (which is the preferred way) than copying or symlinking this lib will change the ORIGIN and the lib won't find its dependencies on its own.

The conda installation routine would need to be adapted as well.


leethargo:
I'm not sure if I understand the implications of this change. Is there now the hard-coded assumption that the library to be linked is called `libscipopt.so`? What other `.so` files would be linked from the same directory, `libipopt.so`? I'd like to avoid the situation where the `setup.py` requires the specific directory structure that results from running `make` in the `scipoptsuite-xxx` dir.

Regarding `conda`: I'd prefer to use a single `setup.py` to adapting both. The need for a separate `conda_setup.py` is unclear to me right now, but I believe it resulted from a too strong assumption.

leethargo:
Maybe we can sit together tomorrow and try to update the conda recipes using your new `setup.py`.

mattmilten:
Yes, we should talk about this. I just told @xmunoz to have a look at conda.

xmunoz:
I won't have time to look at it today, but I should tomorrow.

leethargo:
I'm going to move the conda recipes into a separate repository now. First, they might be useful beyond PySCIPOpt. Second, the recipe for PySCIPOpt is setup to always build from master, which brings some cyclic dependencies in when developing setup.py and the recipe.

xmunoz:
So I was able to upload something to conda, here is how the admin looks:
[
screenshot 2016-07-06 12 02 42
](url)

There is a download count.

mattmilten:
Good! I guess, if there is a way to link to or include the ZIB license we're fine.

xmunoz:
Looks like @leethargo [got started](https://anaconda.org/leethargo/pyscipopt) on that, but the link to "ZIB Academic License" is dead.

leethargo:
Well, I did not specify this link, only the name of the license. This must be some anaconda-magic.

mattmilten:
Now, we check for the env var SCIPOPTDIR and set the links to `lib` and `scip/src` accordingly. If the links are not present or don't point to the correct locations installation will be aborted.

PR #15, opened 2016-07-07
allow to retrieve obj values per variable or as Expr

mattmilten:


mattmilten:
Joao asked me how to get the objective function coefficients and apparently it wasn't possible to get them.
I also added a method to return the entire obj function as Expr. This comes in handy when you want to restore a previously saved one.

fserra:
nice

you should tell Joao to open tickets in the project

pyscipopt/expr.pxi
```diff
@@ -135,7 +132,12 @@ cdef class Expr:
         '''turn it into a constraint'''
         return _expr_richcmp(self, other, op)
 
+    def normalize(self):
```
leethargo: Why did you move the method down?
I think you can't rename it, because self.expr._normalize() is still called further down. Hopefully, some test should fail now.

leethargo:
Unrelated, but still: It seems like I can't currently build PySCIPOpt. Do I need a special and new `libscipopt.so` for the updated `setup.py`?

leethargo:
Could you rebase this branch on master? I think it does not compile because of the identation problem (that you already fixed).

mattmilten:
It works fine for me...

```
>> ldd lib/libscipopt.so 
        linux-vdso.so.1 =>  (0x00007ffef338e000)
        libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f2fc85b3000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f2fc82ad000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f2fc8097000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f2fc7cd2000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f2fc9295000)

>> make SHARED=true GMP=false READLINE=false ZLIB=false VERBOSE=true scipoptlib -j8
```

leethargo:
I reopened the [issue in GitLab](https://git.zib.de/integer/scipoptsuite/issues/1) and moved my comment there.

pyscipopt/expr.pxi
```diff
@@ -135,7 +132,12 @@ cdef class Expr:
...
+        '''remove terms with coefficient of 0'''
+        self.terms =  {t:c for (t,c) in self.terms.items() if c != 0.0}
+
     def __repr__(self):
+        self.normalize()
```
leethargo: Maybe this should be reverted, too, since it's redundant now and I don't like side-effects :-)

mattmilten:
Sure!

leethargo:
So, the tests still crash for me, but I believe it's unrelated to these changes :confused:
If they pass for you, we can merge this.

PR #19, opened 2016-07-18
add some missing features

mueldgog:


fserra:
Hey @mueldgog, does it make sense to have a method `printBestSol` which calls `SCIPprintBestSol` and use that in `optimize` instead? Do you think somebody would like to print the best sol after optimize?

mattmilten:
That's also my only complaint with this pull request. I don't think it's
necessary. I would consider printing the objective value or returning the
optimal solution instead of printing it.

mueldgog:
I have decided on adding a function `printBestSol()`. Make things easier if you just want to have a quick look on the solution.

fserra:
I think the line
`self._bestSol = Solution.create(SCIPgetBestSol(self._scip))`
belonged to `optimze`... though I would guess having that line there doesn't really make sense (what happens if there is no solution at all?), I have no idea whether moving that line breaks something.. do the tests pass?

mueldgog:
Oh, you are right. I removed it but I don't get what you mean by 

> do the tests pass?

Of course, there are no tests using `printBestSol()` and also optimize was never called with `True`. Anyway, the `printBestSol()` can be called even if no solution has been found by SCIP (SCIPprintBestSol() checks for the case and prints a warning message).

mattmilten:
We have a bunch of tests that are supposed to run using py-test from the main directory or within the tests, like so:
```py.test```

mueldgog:
@mattmilten I get it but I just added some functionalities, so which tests should fail?

fserra:
I meant, since you changed the line where `_bestSol` was set, maybe there
was some methods that used `_bestSol` which could fail now because it was
not defined anymore

PR #20, opened 2016-07-19
add getRedcost and getBasisInds to lp.pxi

mueldgog:


mattmilten:
I don't like this. I see that you need this functionality but I suppose that implementing a wrapper for `SCIPgetVarRedcost()` makes much more sense and is more intuitive for the typical user.

mattmilten:
OK, my comment was wrong.

PR #22, opened 2016-07-19
starting new implementation of Expr

leethargo:
redo pull request from branch on central repo.

supersedes #21.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,9 +49,17 @@ After following the previous two steps, it is then possible to create functions 
 Gotchas
 =======
 
+Ranged constraints
+------------------
+
 While ranged constraints of the form
 
     lhs <= expression <= rhs
 are supported, the Python syntax for [chained comparisons](https://docs.python.org/3.5/reference/expressions.html#comparisons) can't be hijacked with operator overloading. Instead, parenthesis must be used, e.g.,
 
     lhs <= (expression <= rhs)
+
+Variable objects
+----------------
+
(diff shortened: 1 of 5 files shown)


leethargo:
I just rebased it on master. Sorry for the mess.

fserra:
just to document what seemed to be the issue: the code just before 0311b1c said
```python
cdef create(SCIP_VAR* scipvar):
    var = Variable()
    var.var = scipvar
    return var
def __init__(self):
    Expr.__init__(self, {Term(self) : 1.0})
```
So at the moment when `Expr.__init__` is called, `var.var` hasn't been assigned, hence its `hash` was 0 and if we tried to print it we got the segfault. This happened, since `Expr.__init__` tried to access `var.var` at some point

leethargo:
@mattmilten Can you add the example from the bug report as a test now?

mattmilten:
The example that triggered the bug is running considerably slower now. Below is a excerpt from a cProfile run with i=500, i.e. 250000 terms to add.

```
1500499/1250500    4.814    0.000    1.886    0.000 expr.pxi:47(__hash__)
       16370098   13.368    0.000   15.513    0.000 expr.pxi:50(__eq__)
```

leethargo:
Relevant [stackoverflow](http://stackoverflow.com/questions/38434300/why-is-quicksum-very-slow-in-scip-python-interface) that started this change.

leethargo:
I'm currently adding the example from SO as a test and have to say that even with 1000 variables, it still takes quite some time (87s on my notebook). Maybe it's not reasonable to create constraints that are that large, but it's still longer than I thought after the profiling yesterday :disappointed:

leethargo:
Anyway, as a bugfix this pull request is valid and complete, so if nobody complains I'll merge it now.

PR #25, opened 2016-07-21
replace scipErrorMessage decorator with expl. PY_SCIP_CALL

### LINKED ISSUE #24: `SCIP_CALL` decorator
The `scipErrorHandler` decorator destroys our docstrings.

There are two ways around this:

1. using `functools.wraps` to preserve the docstring
2. use `PY_SCIP_CALL` explicitly instead of the decorator.

What do you think, @leethargo @fserra @xmunoz?

mattmilten:
fixes #24

### DIFF
--- a/pyscipopt/scip.pyx
+++ b/pyscipopt/scip.pyx
@@ -163,7 +163,6 @@ def PY_SCIP_CALL(SCIP_RETCODE rc):
         raise Exception('SCIP: maximal branching depth level exceeded!')
     else:
         raise Exception('SCIP: unknown return code!')
-    return rc
 
 cdef class Column:
     """Base class holding a pointer to corresponding SCIP_COL"""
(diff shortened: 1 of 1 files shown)


pyscipopt/scip.pyx
```diff
@@ -362,31 +357,26 @@ cdef class Model:
         # http://docs.cython.org/src/reference/extension_types.html#finalization-dealloc
         PY_SCIP_CALL( SCIPfree(&self._scip) )
 
-    @scipErrorHandler
     def create(self):
         """Create a new SCIP instance"""
-        return SCIPcreate(&self._scip)
+        return PY_SCIP_CALL(SCIPcreate(&self._scip))
```
leethargo: Actually, you should also get rid of the `return`.

`PY_SCIP_CALL` doesn't return anything, but would raise an exception if necessary.

PR #27, opened 2016-07-25
add freeTransform() call to setObjective()

mattmilten:
`setObjective()` won't work if `freeTransform()` not called after an `optimize()`. So we should put it in directly.

### DIFF
--- a/pyscipopt/scip.pyx
+++ b/pyscipopt/scip.pyx
@@ -433,6 +433,7 @@ cdef class Model:
         sense -- the objective sense (default 'minimize')
         """
         assert isinstance(coeffs, Expr)
+        self.freeTransform()
         for term, coef in coeffs.terms.items():
             var = <Variable>term[0]
             PY_SCIP_CALL(SCIPchgVarObj(self._scip, var.var, coef))


fserra:
There are plenty of stuff that won't work after a `optimize()`. We had this very same issue in [CSIP](https://github.com/leethargo/CSIP) and we decided to call `freeTransform()` right after `optimize`:
```C
    SCIP_in_CSIP(SCIPsolve(model->scip));
    model->status = getStatus(model);
    SCIP_in_CSIP(SCIPfreeTransform(model->scip));
```
Now, storing *only* the status is far from correct and I guess we don't really want to the same here.

As far as I see, there are two solution:
1 We call `freeTransform` at the beginning of any function that will not work after a `SCIPsolve`
2 We have a method, say something like, `reset` which calls `freeTransform` and then allows the user to modify the problem and solve again.

Thoughts?

mattmilten:
The second approach is kind of in line with Gurobi's `update()` method. If I remember correctly, we had a `freeTransform()` in the `optimize()` at some point and then decided that it should not be called there.

leethargo:
I don't think Gurobi's `update` is comparable the the `reset` that Felipe is mentioning. In my understanding, Gurobi uses `update` to actually "flush" a queue of changes (add variable etc.) so that later changes to the model (add constraint etc.) can be done consistently.

In our case, the `reset` would have to be called *before* any changes may be done, not *after* the changes (to make them have an effect). Also, it's only needed after a call of `optimize`.

Having the user call `reset` between solves is easier for PySCIPOpt, because it means we don't have to store copies of results, such as status, bounds or solutions, which would become invalid with `freeTransform`. It also means that we should catch errors because of the missing `freeTransform` and provide a nice error message.

fserra:
i agree with Robert. It doesn't make sense to implicitly call
freeTransform. It would be a lot of bookkeeping for us, just to save a
`reset` from the user. It's just not worth it

leethargo:
Issue #31 related.

leethargo:
For what it's worth, we decided to change the approach for CSIP and SCIP.jl and don't call `freeTransform` directly after the solve (because then we would have to store too many results).
Instead we go the other way and call `freeTransform` whenever the problem is changed.
This also invalidates the results, but we might be able to warn the user (wrt stages).

fserra:
I would suggest here to force the user to call `freeTransform` whenever the
user wants to modify the problem.

We do it implicitly in CSIP to satisfy JuMP, here we do not need to satisfy
anybody

mattmilten:
I agree.

PR #43, opened 2017-01-11
move TSP example to 'finished'

mattmilten:
Copy of read_tsp.py is necessary to be able to read actual TSP instances.


fserra:
> Copy of read_tsp.py is necessary to be able to read actual TSP instances.

so you mean we have to have one in unfinished until all the tsp examples in unfinished go to finished?

fserra:
1. `addcut2` is defined but never used.
2. this is not "idiomatic" SCIP. SCIP is being used just to solve the model modified externally. In other words, every time a new subtour elimination constraint is generated, SCIP is called to solve the problem from _scratch_.

Do we want this?

PR #47, opened 2017-02-10
Update INSTALL.md

blengerich:
I encountered a few difficulties while installing, so I added a little bit to the INSTALL documentation. Feel free to modify as the maintainers of this repo see fit.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -10,11 +10,11 @@ from the root of the SCIP Optimization Suite directory. This will result in the 
 
 From within the `PySCIPOpt` directory, please execute the following command:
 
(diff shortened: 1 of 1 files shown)


fserra:
Hey

thanks for the PR. In which system did you install?

blengerich:
Mac OSX El Capitan (Version 10.11.6), with Python 2.7.10.

mattmilten:
I just read [a bit about environment variables](https://askubuntu.com/questions/205688/whats-the-difference-between-set-export-and-env-and-when-should-i-use-each) and it seems that using your suggestion is actually slightly safer - although they should do exactly the same thing. I guess that the simple variant might cause issues if the ´python´  command opens another environment.

PR #56, opened 2017-03-17
update package configuration

mattmilten:
- move all sources into `src/pyscipopt` (this allows to `import pyscipopt` also from the root directory of the package)
- additionally export Cython sources into distribution


mattmilten:
fixes #55

PR #63, opened 2017-05-29
update Travis config to use precompiled SCIPOpt library

mattmilten:


mattmilten:
This does not work because the provided .so library of SCIP has a dependency to IPOPT which is not supposed to be installed in every Travis build.

PR #65, opened 2017-06-01
new methods to change lhs/rhs of linear and quadratic constraints

### LINKED ISSUE #64: allow changing rhs and lhs of existing constraints
see https://stackoverflow.com/questions/44308879/update-rhs-on-a-constraint-in-scip-using-python

mattmilten:
fixes #64

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ See [INSTALL.md](INSTALL.md) for instructions.
 How to build a model using PySCIPOpt
 ====================================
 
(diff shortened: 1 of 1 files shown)


mattmilten:
Is there any way to modify the sides of a nonlinear constraint?

fserra:
not really. I wonder why though. @svigerske is there any reason why the nonliner constraint handler doesn't have methods to set or change the sides?

svigerske:
No special reason, I believe. Probably just noone needed it so far.

PR #73, opened 2017-06-10
Correct dual value

fserra:


fserra:
this is your code btw :)

mattmilten:
I think this looks good ;-)

PR #74, opened 2017-06-12
New methods del cons and del cons local

### LINKED ISSUE #67: removing and adding constraints
I want to solve successive optimization problems, by removing and adding a set of constraints. I'm not sure how this may best be achieved. In the C-API there is a method to disable/enable a constraint, would this be the way to go? If it is, would it be possible to extend the python interface to include this functionality?

mattmilten:
fixes #67

### DIFF
--- a/tests/test_lp.py
+++ b/tests/test_lp.py
@@ -31,6 +31,19 @@ def test_lp():
     assert round(s.getVal(x)) == 5.0
     assert round(s.getVal(y)) == 0.0
 
+    s.freeProb()
+    s = Model()
+    x = s.addVar("x", vtype = 'C', obj = 1.0)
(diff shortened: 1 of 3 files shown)


tests/test_lp.py
```diff
@@ -31,6 +31,19 @@ def test_lp():
...
+    y = s.addVar("y", vtype = 'C', obj = 2.0)
+    c = s.addCons(x + 2 * y >= 1.0)
+    s.setMaximize()
+
+    s.delCons(c)
+
+    s.optimize()
+
+    assert s.getStatus() == 'unbounded'
+
```
fserra: I guess the constraint should be `<=` ! otherwise the test passes with or without a correct implementation of `delCons`

Issue #79, opened 2017-06-15
migrate code for getting dual values into SCIP itself


mattmilten:
Do the necessary trnasformations on the SCIP side instead of in PySCIPOpt.

mattmilten:
This is likely not going to happen due to how SCIP handles dual values internally.

PR #80, opened 2017-06-15
fix dualsol method and allow access to rhs and lhs of constraints

### LINKED ISSUE #78: properly handle dual values of bound constraints

mattmilten:
fix #78

### DIFF
--- a/src/pyscipopt/scip.pyx
+++ b/src/pyscipopt/scip.pyx
@@ -268,6 +268,10 @@ cdef class Variable(Expr):
         """Returns current objective value of variable"""
         return SCIPvarGetObj(self.var)
 
+    def getLPSol(self):
+        """Returns the current LP solution value of variable"""
+        return SCIPvarGetLPSol(self.var)
+
 
 cdef class Constraint:
     cdef SCIP_CONS* cons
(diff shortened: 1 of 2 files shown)


mattmilten:
This duplicates the behaviour of SCIP's internal implementation and is supposed to be removed with the next SCIP release #79

src/pyscipopt/scip.pyx
```diff
@@ -1138,13 +1170,30 @@ cdef class Model:
...
+                PY_SCIP_CALL(SCIPgetConsVars(self._scip, transcons.cons, _vars, _nvars, &_success))
+                LPsol = SCIPvarGetLPSol(_vars[0])
+                rhs = SCIPgetRhsLinear(self._scip, transcons.cons)
+                lhs = SCIPgetLhsLinear(self._scip, transcons.cons)
+                if (LPsol == rhs) or (LPsol == lhs):
+                    dual = SCIPgetVarRedcost(self._scip, _vars[0])
+                    print(LPsol, rhs, lhs)
+                else:
+                    dual = 0
+
```
fserra: mmm There is a function `SCIPgetVarsLinear` that gives you the vars right away (so you don't need the malloc)

Also, should `SCIPgetVarRedcost` be 0 when `LPsol != rhs` and `LPsol != lhs` ?

The print should be removed

src/pyscipopt/scip.pyx
```diff
@@ -1138,13 +1170,31 @@ cdef class Model:
...
                 dual = SCIPgetDualsolLinear(self._scip, transcons.cons)
             else:
-                dual = SCIPgetDualsolLinear(self._scip, cons.cons)
+                _vars = SCIPgetVarsLinear(self._scip, transcons.cons)
+                LPsol = SCIPvarGetLPSol(_vars[0])
+                rhs = SCIPgetRhsLinear(self._scip, transcons.cons)
+                lhs = SCIPgetLhsLinear(self._scip, transcons.cons)
+                if (LPsol == rhs) or (LPsol == lhs):
+                    dual = SCIPgetVarRedcost(self._scip, _vars[0])
+
```
fserra: I still don't get why you have to check if the constraint is tight... if it is not tight the reduce cost should be 0 anyway..

mattmilten:
Yeah, I guess you're right.

src/pyscipopt/scip.pyx
mattmilten: I just checked this and it actually isn't 0 for some reason. So we really need to do this extra check.

mattmilten:
@fserra could you please accept the pull request?

fserra:
I knew there was something i needed to do, but couldn't remember what it
was. thanks!

PR #81, opened 2017-06-16
UX improvements

mattmilten:
Some small user experience improvements

### DIFF
--- a/src/pyscipopt/scip.pyx
+++ b/src/pyscipopt/scip.pyx
@@ -618,7 +618,7 @@ cdef class Model:
 
         Keyword arguments:
         var -- the variable
-        lb -- the lower bound (default None)
+        lb -- the lower bound (default -infinity)
         """
         if lb is None:
            lb = -SCIPinfinity(self._scip)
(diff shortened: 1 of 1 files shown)


mattmilten:
Hey @fserra, any reason not to merge this?

fserra:
Oh i didn't check the code before merging. Is it really a good idea that by default chgrhs chglhs set the bound to infinity?? If somebody calls chgrhs and doesn't provide a rhs I would expect and error, because it is most likely a bug

mattmilten:
I did this on purpose. If you change the side and don't set one, i.e. leave it empty, you obviously don't want to set a side, hence +/- infinity. No?

fserra:
mmm i wouldn't think so, i would imagine you forgot to set a side. I feel
that those type of this make the code much less readable. Furthermore, that
is what I'd expect from a function called, removelhs or removerhs

mattmilten:
I see. Let's do like this: Leave the behaviour but remove the default value. So if you set a bound to `None` it will be infinity. But you always have to set one.

fserra:
ok. i agree

mattmilten:
This is committed in f039dd75

PR #89, opened 2017-09-06
additional functionality for writing problem and getting depth

stephenjmaher:
Just some small additions to the interface. I added the possibility of writing the transformed problem. Also, the user can access the depth of the current node from the python interface.

### DIFF
--- a/src/pyscipopt/scip.pyx
+++ b/src/pyscipopt/scip.pyx
@@ -541,6 +541,7 @@ cdef class Model:
 
         Keyword arguments:
         filename -- the name of the file to be used (default 'origprob.cip')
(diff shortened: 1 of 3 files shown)


src/pyscipopt/scip.pyx
```diff
@@ -532,7 +536,7 @@ cdef class Model:
             self.setIntParam("propagating/maxrounds", 0)
 
     # Write original problem to file
-    def writeProblem(self, filename='origprob.cip'):
+    def writeProblem(self, filename='origprob.cip', trans=False):
```
mattmilten: Would you please also add the new parameter to the comment?

Issue #92, opened 2017-09-25
zero optimal dual variables in test_pricer


AlexeyPV:
Hi,

I am trying to implement column generation using pyscipopt and have a problem with optimal dual values. In test_pricer.py example I get zeros from self.model.getDualsolLinear(c) independent of the input. This results in a solution, that uses only the initial variables, and actually prohibits the generation of new ones.

E.g. for a simple instance
```
# item widths
widths = [1, 2, 3, 4]
# width demand
demand = [1, 1, 1, 1]
# roll length
rollLength = 10
```

I get an optimal objective value 1,133333 from test_pricer.py instead of 1 (by packing everything 1+2+3+4 = 10 in a single roll).

Best,
Alexey

P.S.:
Adding the following lines didn't help.
```
model.setPresolve(pyscipopt.SCIP_PARAMSETTING.OFF)
model.setHeuristics(pyscipopt.SCIP_PARAMSETTING.OFF)
model.disablePropagation()
```

PR #94, opened 2017-09-27
get the dual solution also for singleton constraints

mattmilten:


mattmilten:
fixes #92

src/pyscipopt/scip.pyx
```diff
@@ -1276,9 +1276,8 @@ cdef class Model:
                 transcons = <Constraint>self.getTransformedCons(cons)
             else:
                 transcons = cons
-            if _nvars > 1:
-                dual = SCIPgetDualsolLinear(self._scip, transcons.cons)
-            else:
+            dual = SCIPgetDualsolLinear(self._scip, transcons.cons)
+            if dual == 0.0 and _nvars == 1:
```
fserra: I don't quite get the change. Looks very cosmetic to me. Also, why would we trust `SCIPgetDualsolLinear` if it returns nonzero when the constraint contains only one variable?

tests/test_pricer.py
```diff
@@ -132,15 +132,15 @@ def test_cuttingstock():
...
+            outline = 'Pattern_' + str(i) + ':\t' + str(solValue) + '\t\tCuts:\t '
             for j in range(len(widths)):
                 rollUsage += pricer.data['patterns'][i][j]*widths[j]
                 widthOutput[j] += pricer.data['patterns'][i][j]*solValue
                 outline += str(pricer.data['patterns'][i][j]) + '\t'
             outline += 'Usage:' + str(rollUsage)
             print(outline)
 
-    print('\t\t\tTotal Output:','\t'.join(str(e) for e in widthOutput))
+    print('\t\t\tTotal Output:\t', '\t'.join(str(e) for e in widthOutput))
```
fserra: The test is still not testing anything. Could you add some asserts of what is expected?

mattmilten: Yes, good idea! Will do that.

src/pyscipopt/scip.pyx
mattmilten: Apparently, sometimes a bound constraint appears as a constraint in the LP and we actually get a proper dual value while at other times, it gets converted into variable bounds and we need to ask for the reduced costs. In this case, the `getDualsolLinear()` call returns 0. Yeah, it is a workaround, but until we don't properly implement this in SCIP itself, we have to deal with it somehow.

PR #97, opened 2017-10-19
Add read/write for solutions & write for statistics

penenkel:
new functions:
readSol, readSolFile,
printSol, printBestSol, 
writeSol, writeBestSol
writeStatistics

### DIFF
--- a/src/pyscipopt/scip.pyx
+++ b/src/pyscipopt/scip.pyx
@@ -4,7 +4,7 @@ import sys
 
 from cpython cimport Py_INCREF, Py_DECREF
 from libc.stdlib cimport malloc, free
-from libc.stdio cimport FILE, fopen, fdopen, fclose
+from libc.stdio cimport fdopen
 
 include "expr.pxi"
 include "lp.pxi"
--- a/src/pyscipopt/scip.pyx
+++ b/src/pyscipopt/scip.pyx
@@ -1776,6 +1776,18 @@ cdef class Model:
         """Print statistics."""
         PY_SCIP_CALL(SCIPprintStatistics(self._scip, NULL))
 
+    def writeStatistics(self, filename="origprob.stats"):
+      """Write statistics to a file.
+
+      Keyword arguments:
+      filename -- name of the output file
+      """
+      # use this doubled opening pattern to ensure that IOErrors are
+      #   triggered early and in Python not in C,Cython or SCIP.
+      with open(filename, "w") as f:
+          cfile = fdopen(f.fileno(), "w")
+          PY_SCIP_CALL(SCIPprintStatistics(self._scip, cfile))
+
     # Verbosity Methods
 
     def hideOutput(self, quiet = True):
--- a/src/pyscipopt/scip.pyx
+++ b/src/pyscipopt/scip.pyx
@@ -1532,6 +1532,18 @@ cdef class Model:
         PY_SCIP_CALL(SCIPcreateSol(self._scip, &solution.sol, _heur))
         return solution
 
+    def printSol(self, Solution solution = None):
+      pass
+
+    def readSol(self, filename):
+        """Reads a given solution file, problem has to be transformed in advance.
+
+        Keyword arguments:
+        filename -- name of the input file
+        """"
+        fn = str_conversion(filename)
+        PY_SCIP_CALL(SCIPreadSol(self._scip, fn))
+
     def setSolVal(self, Solution solution, Variable var, val):
         """Set a variable in a solution.
 
--- a/src/pyscipopt/scip.pyx
+++ b/src/pyscipopt/scip.pyx
@@ -4,6 +4,7 @@ import sys
 
 from cpython cimport Py_INCREF, Py_DECREF
 from libc.stdlib cimport malloc, free
+from libc.stdio cimport FILE, fopen, fdopen, fclose
 
 include "expr.pxi"
 include "lp.pxi"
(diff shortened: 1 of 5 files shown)


src/pyscipopt/scip.pyx
```diff
@@ -1706,6 +1776,18 @@ cdef class Model:
...
+
+      Keyword arguments:
+      filename -- name of the output file
+      """
+      # use this doubled opening pattern to ensure that IOErrors are
+      #   triggered early and in Python not in C,Cython or SCIP.
+      with open(filename, "w") as f:
+          cfile = fdopen(f.fileno(), "w")
+          PY_SCIP_CALL(SCIPprintStatistics(self._scip, cfile))
+
```
mattmilten: I would prefer to move this functionality into the existing method `printStatistics()` and use the filename as optional parameter. If it's not set, the statistics are printed to the console.

src/pyscipopt/scip.pyx
```diff
@@ -1532,6 +1533,79 @@ cdef class Model:
...
+        Keyword arguments:
+        filename -- name of the output file
+        write_zeros -- include variables that are set to zero
+        """
+        # use this doubled opening pattern to ensure that IOErrors are
+        #   triggered early and in Python not in C,Cython or SCIP.
+        with open(filename, "w") as f:
+            cfile = fdopen(f.fileno(), "w")
+            PY_SCIP_CALL(SCIPprintBestSol(self._scip, cfile, write_zeros));
```
mattmilten: This way of implementing file operations looks nice. What do you think, @fserra?

src/pyscipopt/scip.pyx
penenkel: hm, I would not recommend this as then the function name is inconsistent, although I can understand the desire to keep the interface lean.
print = output to screen
write = output to file

I have to say that I find the structure/naming conventions of SCIP itself rather convoluted. I personally would never have named a function print if it wrote to file.

mattmilten: I get your point. It's less "SCIPy", though,... but this does not necessarily have to be a bad thing ;-) @fserra what do you think? I'd be fine with a new function `writeStatistics()`.

src/pyscipopt/scip.pyx
fserra: I like it. I guess it is alright if IO errors are caught by python rather than SCIP.

src/pyscipopt/scip.pyx
fserra: I have never thought of the difference between write and print, but I have to say that I agree with @penenkel. I guess we can depart from the SCIP way of things here while keeping the interface consistent. However, in general, I wouldn't depart too much from SCIP.

penenkel:
@mattmilten @fserra Does anyone know why AppVeyor keeps failing?
After passing al tests it just says `Command exited with code -1073740791` which doesn't mean anything to me.

mattmilten:
Huh, that's weird. I couldn't figure out yet why this is happening either.

mattmilten:
I just reran the latest master commit on Appveyor and it succeeded, so there's no issue with the environment.

penenkel:
On my system it works perfectly fine with Python 3.6 but crashes with Python 2.7.

mattmilten:
The Appveyor checks still crash - strangely right after successful completion of all tests. Do you have any idea why this is the case?

penenkel:
@mattmilten No, but it definitely has something to do with the writing to file part. I tried using `fopen ` instead of `fdopen` but that did not change a thing. The weird part is how the symptoms depend on the Python version. Did you notice that it fails immediately on Python 2.7 but only after all tests on Python 3.5 and 3.6? Also it works consistently for 3.6 on my system but in 2.7 it fails. Where fail is actually a fatal crash of the entire python instance.

penenkel:
I figured out that the ScipOpt binaries we use are compiled with VC 14 (i.e Visual Studio 2015?) (at least they depend on `MSVCP140.DLL`). AFAIK any compiled Python extension must use the same compiler version as Python itself or else weird errors my crop up. As we are dealing with unspecific memory exceptions (see my earlier comment) this may very well be the cause of our troubles, at least with regards to Python 2.7. 
I still have no theory what the problem is with Python 3.5 and 3.6 especially because the later works on my desktop.

mattmilten:
That sounds very reasonable. Shall we then just switch to VS 2015 for the Appveyor builds?

mattmilten:
Yet another workaround would be to also compile SCIP within Appveyor, therefore using the same environment. But this also dramatically increases the total time for the build. 

For the Travis builds we also do it like that because of annoying C++ ABI incompatibilities between the Travis environment and the pre-built SCIP library.

penenkel:
I'm afraid it isn't as simple as switching the compiler for this project (`PySCIPOpt`). My understanding ist that we are essentially dealing with three independently compiled parts: SCIPOpt, Python and PySCIPOpt as a bridge between them. I'm fairly sure that the compiler version used for Python dictates which compiler must be used for PySCIPOpt as it contains C/Cython-Code. This is fairly well documented [here](https://wiki.python.org/moin/WindowsCompilers) and [here](https://github.com/cython/cython/wiki/CythonExtensionsOnWindows). As I stated in my previous post I suspect that this constraint may also apply to binaries that are only indirectly included via Cython (i.e. SCIPOpt).

AFAIK Visual Studio 2017 and 2015 use the same Visual Compiler VC14 which is to use with Python 3.5 and 3.6, so that should not make any difference. For Python 2.7 the recommended compiler is `Microsoft Visual C++ Compiler for Python 2.7` which is included in the 2015 AppVeyor image. So as far as PySCIPOpt goes AppVeyor should be setup correctly as I already allowed for this.

We could indeed attempt to also build SCIP within AppVeyor using the "correct" compiler versions as defined by the Python version. I don't know how difficult that would be. I don't think that total build time is that much of a problem for an automated build, but the last time I tried building SCIP under Windows (~3-5 years ago) it was a pain and I don't think I succeeded in the end.

mattmilten:
Yes, you are right. It was really quite complicated to compile SCIP in previous years. In the meantime, though, we added a CMake configuration that allows a pretty straightforward compilation/installation process also on Windows. I'm going to set this up in Appveyor until tomorrow and then we can see whether this already solves the issue.

And yes, sure, build times on a CI server are not that important - I still get a bad conscious when I think about the involved computing resources :-)

mattmilten:
I tried to modify the AppVeyor config to build SCIP from source (see #101). Unfortunately, I cannot compile PySCIPOpt anymore. For some reason there is a linking error when building the interface. Another very annoying issue is that we publish the SCIP and SoPlex source code as ´.tgz´ archive which requires additional tools to extract them.

For now, this does not seem to be a viable solution for the problem at hand :-(

PR #98, opened 2017-10-24
append default Python gitignore from GitHub

penenkel:


mattmilten:
Having a larger `gitignore` can never hurt :-)

PR #99, opened 2017-10-25
Improve appveyor

penenkel:
add python 2.7 and 3.5 to appveyor


mattmilten:
So this `stdint.h` is required for Python 2.7 on Windows? Interesting.

Looks a bit hacky to be honest, but I'm going to merge it anyway.

Thanks a lot!

penenkel:
Yeah, it's a hack. Apparently Microsoft never fully supported C99 in its C compilers. Some versions of Visual Studio include that file some don't. Visual C++ Compiler for Python 2.7 which is used nowadays for Python 2.7 does not.

PR #117, opened 2018-01-16
Penenkel feature read sol

mattmilten:
This is a duplicate of the original PR and only there for testing purposes.


mattmilten:
Hey @penenkel, what do you think of just removing the tests from your contribution? Sadly, I don't see a solution to the AppVeyor crashes and can also not reproduce them on my machine.
Still, I would like to have your interface extension in the master branch.

PR #120, opened 2018-01-29
add coverage configuration to cover interface instead of tests

mattmilten:


mattmilten:
see https://stackoverflow.com/questions/48505458/coverage-of-cython-module-using-py-test-and-coverage-py for details on the new configuration

PR #127, opened 2018-02-12
support changing objective offset

### LINKED ISSUE #124: Constant offsets in objective are not supported
Hi all,

"Constant offsets in objective are not supported!"......Is it possible to overcome this issue as i am setting the objective, without removing my constant offsets? Maybe a "tricky" way or an implementation in the near future?

Best,
Loukas

mattmilten:
fixes #124


mattmilten:
Yeah, someone might want to use those in the future. I will add them.

PR #129, opened 2018-02-13
use junitxml option to specify an output file for coverage reports

mattmilten:


mattmilten:
We can close this because we're not running coverage tests.

Issue #123, opened 2018-02-04
ReOptimization


isidian:
It would be nice to be able to use the re-optimization features of SCIP in the API.  I attempted this myself, and it was a simple enough matter to do
scip.setBoolParam("reoptimization/enable",True)
and also fairly straightforward to modify the .pyx and .pxd files to import SCIPfreeReoptSolve()

But when I tried to implement SCIPchgReoptObjective() it was not going well.  There is a difference in how it is called versus the current implementation of Model.SetObjective().  SCIPchgReoptObjective asks for all variables and coefficients all at once, but the current implementation of Model.SetObjective goes through each variable one at a time, so I cannot simply modify the existing code.

isidian:
Here's a quick hack that I'm using right now.  It borrows heavily from Model.setObjective().

I say it's a quick hack because it doesn't include any kind of error checking, which would be important if this were pushed into the repo version.

```
def chgReoptObjective(self, coeffs, sense = 'minimize'):
        """Establish the objective function as a linear expression.

        :param coeffs: the coefficients
        :param sense: the objective sense (Default value = 'minimize')

        """

        cdef SCIP_OBJSENSE objsense

        if sense == "minimize":
            objsense = SCIP_OBJSENSE_MINIMIZE
        elif sense == "maximize":
            objsense = SCIP_OBJSENSE_MAXIMIZE
        else:
            raise Warning("unrecognized optimization sense: %s" % sense)
        
        assert isinstance(coeffs, Expr)

        if coeffs.degree() > 1:
            raise ValueError("Nonlinear objective functions are not supported!")
        if coeffs[CONST] != 0.0:
            raise ValueError("Constant offsets in objective are not supported!")

        cdef SCIP_VAR** _vars
        cdef int _nvars
        _vars = SCIPgetOrigVars(self._scip)
        _nvars = SCIPgetNOrigVars(self._scip)
        _coeffs = <SCIP_Real*> malloc(_nvars * sizeof(SCIP_Real))
        
        for i in range(_nvars):
            _coeffs[i] = 0.0
        
        for term, coef in coeffs.terms.items():
            # avoid CONST term of Expr
            if term != CONST:
                assert len(term) == 1
                var = <Variable>term[0]
                for i in range(_nvars):
                    if _vars[i] == var.var:
                        _coeffs[i] = coef

[...]

PR #132, opened 2018-02-14
Added interface for re-optimization routines

isidian:
Added Model.freeReoptSolve() and Model.chgReoptObjective() with corresponding interfaces to SCIP.


mattmilten:
closes #123

PR #133, opened 2018-02-21
added more functions to modify constraint parameters

g-munoz:


src/pyscipopt/scip.pyx
```diff
@@ -1254,6 +1254,33 @@ cdef class Model:
         """
         PY_SCIP_CALL(SCIPappendVarSOS2(self._scip, cons.cons, var.var))
 
+    def chgInitial(self, Constraint cons, newInit):
```
fserra: I don't quite like that it is called `chg` when in SCIP is called `set`. What do you think @mattmilten ?

PR #134, opened 2018-03-02
add support general nonlinear cons

### LINKED ISSUE #41: Non Linear expressions
Hi,

First of all, thanks for the interface, coming from gurobipy, the transition was seamless.

However, I am looking to use non-linear expression, namely a log in the objective function. Looking inside the code and documentation, I couldn't find the correct way to do it.

Is it implemented yet?

Best,
Yoann

fserra:
closes #41

### DIFF
--- a/src/pyscipopt/expr.pxi
+++ b/src/pyscipopt/expr.pxi
@@ -216,12 +216,8 @@ cdef class Expr:
 
     def __div__(self, other):
         ''' transforms Expr into GenExpr'''
-        divisor = buildGenExprObj(other)
-        # we can't divide by 0
-        if divisor.getOp() == Operator.const:
-            assert divisor.number != 0.0
-
-        return buildGenExprObj(self) * divisor**(-1)
+        selfexpr = buildGenExprObj(self)
+        return selfexpr.__div__(other)
 
     def __rdiv__(self, other):
         ''' other / self '''
@@ -230,12 +226,12 @@ cdef class Expr:
 
     def __truediv__(self,other):
         selfexpr = buildGenExprObj(self)
-        return selfexpr.__div__(other)
+        return selfexpr.__truediv__(other)
 
     def __rtruediv__(self, other):
         ''' other / self '''
         otherexpr = buildGenExprObj(other)
-        return otherexpr.__div__(self)
+        return otherexpr.__truediv__(self)
 
     def __pow__(self, other, modulo):
         if float(other).is_integer() and other >= 0:
@@ -404,11 +400,7 @@ cdef class GenExpr:
         ''' '''
 
     def __abs__(self):
-        ans = UnaryExpr()
-        ans.op = Operator.fabs
-        ans.operatorIndex = Operator.operatorIndexDic[self.op]
-        ans.children.append(self)
-        return ans
+        return UnaryExpr(Operator.fabs, self)
 
     def __add__(self, other):
         left = buildGenExprObj(self)
@@ -528,9 +520,8 @@ cdef class GenExpr:
     def __div__(self, other):
         divisor = buildGenExprObj(other)
         # we can't divide by 0
-        if divisor.getOp() == Operator.const:
-            assert divisor.number != 0.0
-
+        if divisor.getOp() == Operator.const and divisor.number == 0.0:
+            raise ValueError("cannot divide by 0")
         return self * divisor**(-1)
 
     def __rdiv__(self, other):
@@ -539,13 +530,17 @@ cdef class GenExpr:
         return otherexpr.__div__(self)
 
     def __truediv__(self,other):
-        selfexpr = buildGenExprObj(self)
-        return selfexpr.__div__(other)
+        print("truediv of GenExpr", self, other)
+        divisor = buildGenExprObj(other)
+        # we can't divide by 0
+        if divisor.getOp() == Operator.const and divisor.number == 0.0:
+            raise ValueError("cannot divide by 0")
+        return self * divisor**(-1)
 
     def __rtruediv__(self, other):
         ''' other / self '''
         otherexpr = buildGenExprObj(other)
-        return otherexpr.__div__(self)
+        return otherexpr.__truediv__(self)
 
     def __neg__(self):
         return -1.0 * self
@@ -622,8 +617,11 @@ cdef class PowExpr(GenExpr):
 
 # Exp, Log, Sqrt Expressions
 cdef class UnaryExpr(GenExpr):
-    def __init__(self):
+    def __init__(self, op, expr):
         self.children = []
+        self.children.append(expr)
+        self.op = op
+        self.operatorIndex = Operator.operatorIndexDic[op]
     def __repr__(self):
         return self.op + "(" + self.children[0].__repr__() + ")"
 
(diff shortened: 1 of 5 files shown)


src/pyscipopt/expr.pxi
```diff
@@ -1,3 +1,47 @@
...
+# expr[Term(x)] returns 0.0
+#
+# On the other hand, when dealing with expressions more general than polynomials,
+# that is, absolute values, exp, log, sqrt or any general exponent, we use GenExpr.
+# GenExpr stores expression trees in a rudimentary way.
+# Basically, it stores the operator and the list of children.
+# We have different types of general expressions that in addition
+# to the operation and list of children stores
+# SumExpr: coefficients and constant
+# ProdExpr: constant
```
leethargo: Nice documentation!

But what is the meaning of the constant for `ProdExpr`? Is it `contant * child1 * child2 * ...` ?

src/pyscipopt/expr.pxi
```diff
@@ -11,21 +55,21 @@ def _is_number(e):
 
 def _expr_richcmp(self, other, op):
     if op == 1: # <=
-        if isinstance(other, Expr):
+        if isinstance(other, Expr) or isinstance(other, GenExpr):
```
leethargo: If `Expr` and `GenExpr` are mostly handled the same, then maybe it would be easier to use a common base class for both of them?

src/pyscipopt/expr.pxi
```diff
@@ -1,3 +1,47 @@
...
+# ProdExpr: constant
+# Constant: constant
+# VarExpr: variable
+# PowExpr: exponent
+# UnaryExpr: nothing
+# We do not provide any way of accessing the internal information of the expression tree,
+# nor we simplify common terms or do any other type of simplification.
+# The `GenExpr` is pass as is to SCIP and SCIP will do what it see fits during presolving.
+#
+# TODO: All this is very complicated, so we might wanna unify Expr and GenExpr.
```
leethargo: It seems to me like `GenExpr` is strictly more powerful than `Expr`, so `Expr` could be removed completely?
I don't know whether there is a difference in performance between these implementations...

leethargo: In any case, this step could also be done in the future.

src/pyscipopt/expr.pxi
```diff
@@ -111,6 +186,11 @@ cdef class Expr:
         elif _is_number(other):
             c = float(other)
             self.terms[CONST] = self.terms.get(CONST, 0.0) + c
+        elif isinstance(other, GenExpr):
+            # is no longer in place, might affect performance?
+            # can't do `self = buildGenExprObj(self) + other` since I get
+            # TypeError: Cannot convert pyscipopt.scip.SumExpr to pyscipopt.scip.Expr
+            return buildGenExprObj(self) + other
```
leethargo: what if we do this instead:

    return other + buildGenExprObj(self)

Would this be inplace (but for other's place)?

src/pyscipopt/expr.pxi
```diff
@@ -129,14 +209,25 @@ cdef class Expr:
...
+            return buildGenExprObj(self) * other
         else:
             raise NotImplementedError
 
+    def __div__(self, other):
+        ''' transforms Expr into GenExpr'''
+        divisor = buildGenExprObj(other)
+        # we can't divide by 0
+        if divisor.getOp() == Operator.const:
+            assert divisor.number != 0.0
```
leethargo: I think a Python programmer would expect a `ValueError` instead of an `AssertionError`, but I don't mind too much.

src/pyscipopt/expr.pxi
```diff
@@ -247,3 +343,324 @@ def quicksum(termlist):
...
+        return self.__mul__(other)
+
+    def __rsub__(self, other):
+        return -1.0 * self + other
+
+    def __richcmp__(self, other, op):
+        '''turn it into a constraint'''
+        return _expr_richcmp(self, other, op)
+
+    def degree(self):
```
leethargo: why even add this method?

src/pyscipopt/expr.pxi
```diff
@@ -247,3 +343,324 @@ def quicksum(termlist):
...
+
+    def __repr__(self):
+        return str(self.number)
+
+def exp(expr):
+    ans = UnaryExpr()
+    ans.op = Operator.exp
+    ans.operatorIndex = Operator.operatorIndexDic[ans.op]
+    ans.children.append(buildGenExprObj(expr))
+    return ans
```
leethargo: why do set all these attributes after creating an "empty" object, instead of using the `UnaryExpr.__init__` method to initialize them?

src/pyscipopt/scip.pyx
```diff
@@ -891,6 +891,8 @@ cdef class Model:
             return self._addLinCons(cons, **kwargs)
         elif deg <= 2:
             return self._addQuadCons(cons, **kwargs)
+        elif deg == float('inf'): # general nonlinear
+            return self._addGenNonlinearCons(cons, **kwargs)
         else:
             return self._addNonlinearCons(cons, **kwargs)
```
leethargo: so this is for polynomials only?

src/pyscipopt/scip.pyx
```diff
@@ -1008,6 +1010,100 @@ cdef class Model:
         free(varexprs)
         return PyCons
 
+    def _addGenNonlinearCons(self, ExprCons cons, **kwargs):
```
leethargo: I'll have to believe you on this function...

src/pyscipopt/expr.pxi
fserra: yes

src/pyscipopt/expr.pxi
fserra: yes, maybe, but I felt uncertain. The idea was to try to keep Expr as untouched as possible, given that it has been much more tested.

src/pyscipopt/expr.pxi
fserra: Well, this is not true, you can't ask information about GenExpr as you can with Expr, but if we do not care about that, then yes, it is more powerful.

I guess there are performance difference, I didn't really test this, but I would definitely do this in the future

src/pyscipopt/expr.pxi
fserra: I am guessing that inplace means self's place. But yeah, this inplace stuff is a bit complicated and would also postpone it.

src/pyscipopt/expr.pxi
fserra: just to keep it compatible with the implementation of addCons in scip.pxd or whathever the termination is

src/pyscipopt/scip.pyx
fserra: `_addNonlinearCons` is for polynomials only, yes

src/pyscipopt/scip.pyx
fserra: haha, this is basically a copy-paste of the one in CSIP

src/pyscipopt/expr.pxi
fserra: what you suggest is way more elegant, will implement after lunch ;)

leethargo: Good, so I made one constructive suggestions ;-)

mattmilten:
Shall we increase the version number to 1.4 or 2.0 after merging?

src/pyscipopt/expr.pxi
fserra: I changed this and added a test for this. thanks

fserra:
I have no idea how the versioning works

src/pyscipopt/scip.pyx
mattmilten: I would rather rename it to `_addPolyCons()` or something similar. It's quite unintuitive right now and these are only internal methods.

leethargo:
> I have no idea how the versioning works

Apparently, the version is defined in [__init__.py](https://github.com/SCIP-Interfaces/PySCIPOpt/blob/master/src/pyscipopt/__init__.py#L1)

mattmilten:
> Apparently, the version is defined in __init__.py

Haha, I guess, @fserra was referring to what the meaning of the version digits was - not how we can change the version. So, the question boils down to whether we want to consider this a major or a minor version jump. I say we call it 1.4 since we don't break the interface.

PR #135, opened 2018-03-04
add nonlinear test

fserra:
this merge request translates the tests string, circle and gasnet from SCIP to PySCIPOpt

it also allows to set `lb = None` when creating a variable for having -infinite lower bound

### DIFF
--- a/src/pyscipopt/expr.pxi
+++ b/src/pyscipopt/expr.pxi
@@ -530,7 +530,6 @@ cdef class GenExpr:
         return otherexpr.__div__(self)
 
     def __truediv__(self,other):
-        print("truediv of GenExpr", self, other)
         divisor = buildGenExprObj(other)
         # we can't divide by 0
         if divisor.getOp() == Operator.const and divisor.number == 0.0:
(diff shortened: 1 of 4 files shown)


fserra:
I am guessing this is timing out because there is no ipopt

src/pyscipopt/expr.pxi
```diff
@@ -530,7 +530,6 @@ cdef class GenExpr:
         return otherexpr.__div__(self)
 
     def __truediv__(self,other):
-        print("truediv of GenExpr", self, other)
```
mattmilten: good one! :-)

src/pyscipopt/scip.pyx
```diff
@@ -740,6 +740,8 @@ cdef class Model:
         cname = str_conversion(name)
         if ub is None:
             ub = SCIPinfinity(self._scip)
+        if lb is None:
+            lb = -SCIPinfinity(self._scip)
```
mattmilten: weird! I don't know why this was missing...

mattmilten:
The time outs are annoying...

mattmilten:
Maybe adding  a time limit might be a solution to produce a valid status for the test.

fserra:
don't get it, how would this timelimit work? I guess the best would be able to have ipopt, either by having an installer that contains it or compile it there?

mattmilten:
We could check for the right result or status time limit. This would
effectively skip the test. Sure, in a perfect world we would always have
Ipopt available...

Issue #136, opened 2018-03-16
No dual solution available for constraint in LP with presolve disabled.


Viech:
I'm having issues with LP duals (with presolve disabled). Some constraints yield the correct duals, other constraints _in the same problem and during the same optimization pass_ throw `no dual solution available for constraint` warnings. Setting the affected dual values to `0.0` is not correct; some of the affected constraints are supposed to have non-zero duals.

Unfortunately my example where some duals of tight constraints are returned while some are not is too involved for a bug report. However, here is an example with a single constraint where retrieving its dual fails even though the constraint is tight (and its dual value should be `1.0`):

```
#!/usr/bin/python
import pyscipopt
m = pyscipopt.Model()
x = m.addVar("x")
c = m.addCons(x >= 1)
m.setObjective(x, "minimize")
m.setPresolve(pyscipopt.scip.PY_SCIP_PARAMSETTING.OFF)
m.optimize()
m.getDualsolLinear(c)
```
Output on `v1.2.0`:
```
Traceback (most recent call last):
  File "src/pyscipopt/scip.pyx", line 1276, in pyscipopt.scip.Model.getDualsolLinear
  File "src/pyscipopt/scip.pyx", line 1228, in pyscipopt.scip.Model.getTransformedCons
  File "src/pyscipopt/scip.pyx", line 283, in pyscipopt.scip.Constraint.create
Warning: cannot create Constraint with SCIP_CONS* == NULL

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "./test_scip_duals.py", line 9, in <module>
    m.getDualsolLinear(c)
  File "src/pyscipopt/scip.pyx", line 1292, in pyscipopt.scip.Model.getDualsolLinear
Warning: no dual solution available for constraint c1
```
Output on current f5956e0:
```
Traceback (most recent call last):
  File "src/pyscipopt/scip.pyx", line 1577, in pyscipopt.scip.Model.getDualsolLinear
  File "src/pyscipopt/scip.pyx", line 1529, in pyscipopt.scip.Model.getTransformedCons
  File "src/pyscipopt/scip.pyx", line 405, in pyscipopt.scip.Constraint.create
Warning: cannot create Constraint with SCIP_CONS* == NULL

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "./test_scip_duals.py", line 9, in <module>
    m.getDualsolLinear(c)
[...]

PR #137, opened 2018-03-16
use SCIPgetDualSolVal to get dual values of constraints

fserra:


fserra:
So @mattmilten there are a couple of issues here

1. The name, should we rename the function to the scip one? Is a bit of a weird name.
1. using this function still returns the wrong solution: it returns `SCIP_INVALID` because `SCIPgetVarRedcost` says  ` returns SCIP_INVALID if the variable is active but not in the current LP`

So I don't know how to fix the second issue

Viech:
In my small example I posted in #136 I am also getting `SCIP_INVALID`. In my more complex example (which is generated on the fly by PICOS from metaconstraints so that I can't give you a clean copy of it) where I was previously getting the same `no dual solution available` error for some of the constraints, I am now getting `-0.0` for the same subset of constraints. I can try harder to reproduce this in a tangible example, if necessary.

PR #141, opened 2018-03-20
improve travis

fserra:


fserra:
ok apparently it is impossible to upgrade travis' glibc. the self extracting arxive is compiled with a very new gcc apparently and so it cannot work. Any ideas @mattmilten

mattmilten:
It's not that I didn't try something similar and also failed miserably...

mattmilten:
You might want to check out the test-debian branch.

https://github.com/SCIP-Interfaces/PySCIPOpt/blob/test-debian/.travis.yml

mattmilten:
I've seen somewhere that we might need to specify the `GCC` or `CC` variables to point to the new gcc. I can't find it right now, though.

PR #147, opened 2018-04-09
add setParam and getParam methods

mo271:
I suggest to add the methods `getParam` and `setParam`.

Unfortunately, I wasn't able to directly access `SCIPsetParam(SCIP* scip,  const char*  name, void* value)`, but I can't cast the python variable to void* in cython.

### DIFF
--- a/src/pyscipopt/scip.pxd
+++ b/src/pyscipopt/scip.pxd
@@ -820,6 +820,10 @@ cdef extern from "scip/scip.h":
     SCIP_RETCODE SCIPsetEmphasis(SCIP* scip, SCIP_PARAMEMPHASIS paramemphasis, SCIP_Bool quiet)
     SCIP_RETCODE SCIPresetParam(SCIP* scip, const char* name)
     SCIP_RETCODE SCIPresetParams(SCIP* scip)
+    SCIP_RETCODE SCIPsetParam(SCIP* scip,  const char*  name, void* value)
+    SCIP_PARAM* SCIPgetParam(SCIP* scip,  const char*  name)
+
+
 
     # LPI Functions
     SCIP_RETCODE SCIPgetLPI(SCIP* scip, SCIP_LPI** lpi)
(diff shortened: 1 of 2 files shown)


src/pyscipopt/scip.pxd
```diff
@@ -1114,3 +1118,24 @@ cdef extern from "scip/cons_countsols.h":
...
+    ctypedef enum SCIP_PARAMTYPE:
+        SCIP_PARAMTYPE_BOOL    = 0
+        SCIP_PARAMTYPE_INT     = 1
+        SCIP_PARAMTYPE_LONGINT = 2
+        SCIP_PARAMTYPE_REAL    = 3
+        SCIP_PARAMTYPE_CHAR    = 4
+        SCIP_PARAMTYPE_STRING  = 5
+
+    ctypedef struct SCIP_PARAM:
+        pass
```
mattmilten: Could you please move these enum definitions at the top of the file to the other ones?

src/pyscipopt/scip.pxd
```diff
@@ -820,6 +820,10 @@ cdef extern from "scip/scip.h":
     SCIP_RETCODE SCIPsetEmphasis(SCIP* scip, SCIP_PARAMEMPHASIS paramemphasis, SCIP_Bool quiet)
     SCIP_RETCODE SCIPresetParam(SCIP* scip, const char* name)
     SCIP_RETCODE SCIPresetParams(SCIP* scip)
+    SCIP_RETCODE SCIPsetParam(SCIP* scip,  const char*  name, void* value)
+    SCIP_PARAM* SCIPgetParam(SCIP* scip,  const char*  name)
```
mattmilten: We don't need `SCIPsetParam()`, right?

mo271: good catch! I just forgot to remove it, after tinkering with it before...

src/pyscipopt/scip.pxd
mo271: done

PR #152, opened 2018-04-10
warn if problem was not solved before

mo271:
When we access `getActivity` before `optimize`, we can a segmentation fault:
```
In [1]: from pyscipopt import Model
In [2]: M = Model()
In [3]: M.addCons(M.addVar()<=10)
Out[3]: c1
In [4]: M.getActivity(_)
Segmentation fault
```
This patch catches it, and gives a warning instead.


mattmilten:
This is also true for `getSlack()`, right?

mattmilten:
We need to adapt the test in `tests/test_model.py`. I suggest removing the code block about the `badsolution`.

PR #160, opened 2018-05-04
close #158 consistent number check in expr.pxi

### LINKED ISSUE #158: Constant values in Expr mult and div
Hello,

So `Expr.__mult__` check if the input is a number using `_is_number`, but `Expr.__div__` calls `buildGenExprObj` which check if the object is a number using python builtin `isinstance`:

https://github.com/SCIP-Interfaces/PySCIPOpt/blob/910beeb4b770519c55ec25538258bc3a077bb369/src/pyscipopt/expr.pxi#L115

`_is_number` tries to convert the object to a float and catches the eventual error.
`_is_number` is actually more general than `isinstance`, because it will also catch all the numpy types (np.int, np.float, np.float32...) as numbers.

Is it possible to replace the call to `isinstance(int) or isinstance(float)` with `_is_number`?

Thanks

AntoinePrv:


fserra:
thanks

PR #163, opened 2018-05-24
fix for conshdlr and adds expr

stephenjmaher:
Adds the Expr object to the public interface. This is inline with the Gurobi python interface. Also, the conshdlr enfolp callback is corrected. The parameters were incorrect.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,6 +7,8 @@ includedir = os.path.abspath(os.path.join(scipoptdir, 'include'))
 libdir = os.path.abspath(os.path.join(scipoptdir, 'lib'))
 libname = 'scip'
 
+print(scipoptdir)
+
 cythonize = True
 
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -7,6 +7,8 @@
 libdir = os.path.abspath(os.path.join(scipoptdir, 'lib'))
 libname = 'scip'
 
+print(scipoptdir)
+
```
mattmilten: Why is this necessary?

stephenjmaher: Not necessary, but I thought useful for the user to know what the environment variable is set to. I added this when I was having trouble with linking the scipoptdir.

If you don't like it, we can remove it.

mattmilten: I would rather not have it. The `setup.py` should be as minimal/clean as possible.

stephenjmaher: I have updated this now

PR #165, opened 2018-05-29
Logical connectives

ramabile:
Dear SCIP workers, greetings.

The aim of this pull request is to **introduce logical connectives**[1] AND and OR[2] in PySCIPOpt.
For the moment being, I added some test without checking the resultant's value.
Please be direct whether some does not fulfill any policy here.[3] Any comment is much appreciated.

Best wishes.

[1] about nomenclature: since variables are called "operators" in SCIP (see [here](http://scip.zib.de/doc-5.0.1/html/cons__and_8h.php)), I called AND and OR "connectives".
[2] XOR is TBI since its signature is not consistent with AND/OR.
[3] a CONTRIBUTING.md / CONTRIBUTING file could help the project.

### DIFF
--- a/src/pyscipopt/__init__.py
+++ b/src/pyscipopt/__init__.py
@@ -13,6 +13,7 @@ from pyscipopt.scip      import Prop
 from pyscipopt.scip      import Sepa
 from pyscipopt.scip      import LP
 from pyscipopt.scip      import quicksum
+from pyscipopt.scip      import quickmul
 from pyscipopt.scip      import exp
 from pyscipopt.scip      import log
 from pyscipopt.scip      import sqrt
--- a/tests/test_connectives.py
+++ b/tests/test_connectives.py
@@ -1,0 +1,95 @@
+from pyscipopt import Model
+
+################################################################################
+#
+# Testing AND/OR constraints
+#   check whether any error is raised
+# see http://scip.zib.de/doc-5.0.1/html/cons__and_8c.php
+#   for resultant and operators definition
+# CAVEAT: ONLY binary variables are allowed
+#   Integer and continous variables behave unexpectedly (due to SCIP?)
+# TBI: automatic assertion of expected resultant VS optimal resultant
+#   (visual inspection at the moment)
+# TBI: implement and test XOR constraint
+#
+################################################################################
+
+### AUXILIARY ###
+def setModel(vtype="B", name=None, imax=2):
+    if name is None: name = "model"
+    m = Model(name)
+    m.hideOutput()
+    i = 0
+    m.addVar("r", vtype)
+    while i < imax:
+        m.addVar("v%s" % i, vtype)
+        i+=1
+    return m
+
+def getVarByName(m, name):
+    try:
+        return [v for v in m.getVars() if name == v.name][0]
+    except IndexError:
+        return None
+
+def getAllVarsByName(m, name):
+    try:
+        return [v for v in m.getVars() if name in v.name]
+    except IndexError:
+        return []
+
+
+def setConss(m, vtype="B", val=0, imax=1):
+    i = 0
+    while i < imax:
+        vi = getVarByName(m,"v%s" % i)
+        m.addCons(vi == val, vtype)
+        i+=1
+    return
+
+def printOutput(m):
+    status = m.getStatus()
+    r = getVarByName(m,"r")
+    rstr = "%d" % round(m.getVal(r))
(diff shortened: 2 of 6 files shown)


fserra:
I don't understand the error in the continuous integration. I guess it is something related to py.test, which we are using for testing. Apparently, it thinks that `m` is a fixture, which is not, and then it dies...

I guess you should change the test to use `py.test`. Probably the best is to get inspired by the other tests.

BTW: I had a look at the code and everything looks nice ;). It would be nice to have some kind of assertion on the tests but apart from that, I would merge.

src/pyscipopt/__init__.py
```diff
@@ -14,6 +14,7 @@
 from pyscipopt.scip      import LP
 from pyscipopt.scip      import Expr
 from pyscipopt.scip      import quicksum
+from pyscipopt.scip      import quickmul
```
mattmilten: I would prefer the term `quickprod`

tests/test_connectives.py
```diff
@@ -0,0 +1,94 @@
...
+def printOutput(m):
+    status = m.getStatus()
+    r = getVarByName(m,"r")
+    rstr = "%d" % round(m.getVal(r))
+    vs = getAllVarsByName(m, "v")
+    vsstr = "".join(["%d" % round(m.getVal(v)) for v in vs])
+    print("Status: %s, resultant: %s, operators: %s" % (status, rstr, vsstr))
+
+### TEST ###
+def test_connective(m, connective, sense="min"):
```
mattmilten: You cannot define a function that's called `test_...()` and pass some parameters, as these functions will be called by py.test directly, so the parameters are undefined.

tests/test_connectives.py
```diff
@@ -0,0 +1,94 @@
...
+        m.setObjective(r, sense="%simize" % sense)
+        m.optimize()
+        printOutput(m)
+        return True
+    except Exception as e:
+        print("%s: %s" % (e.__class__.__name__, e))
+        return False
+
+### MAIN ###
+if __name__ == "__main__":
```
mattmilten: This function is only necessary for running the code without py.test. All `test_...()` functions must be self-contained. I am sure there is some way of defining pre-conditions that are executed before, but I would just define all necessary variables inside the tests themselves.

tests/test_quickmul.py
```diff
@@ -0,0 +1,41 @@
+from pyscipopt import Model, quickmul
```
mattmilten: Please change to `quickprod` like above.

tests/test_quickmul.py
```diff
@@ -0,0 +1,41 @@
...
+    dim = 20
+    x = [m.addVar("x_%d" % i) for i in range(dim)]
+    expr = quickmul((i+j+1)*x[i]*x[j]
+                    for i in range(dim)
+                    for j in range(dim))
+    cons = expr <= 1.0
+    #                              upper triangle,     diagonal
+    assert cons.expr.degree() == 2*dim*dim
+    m.addCons(cons)
+    # TODO: what can we test beyond the lack of crashes?
```
mattmilten: You could generate the same model the old way and compare the values of every constraint using `getTermsQuadratic()`

tests/test_connectives.py
```diff
@@ -0,0 +1,94 @@
+from pyscipopt import Model
```
mattmilten: Maybe you could change the name to `test_logical`?

tests/test_quickprod.py
```diff
@@ -0,0 +1,41 @@
...
+
+def test_quickprod_model():
+    m = Model("quickprod")
+    x = m.addVar("x")
+    y = m.addVar("y")
+    z = m.addVar("z")
+    c = 2.3
+
+    q = quickprod([x,y,z,c]) == 0.0
+    s = reduce(mul,[x,y,z,c],1) == 0.0
```
mattmilten: `reduce()` is not available in python3. The code must comply to both standards. Could you please rewrite this?

https://stackoverflow.com/a/13638960/673271

mattmilten: This is also why the python3 test does not succeed.

tests/test_quickmul.py
ramabile: Apart from the fact that the tests for `quicksum` and `quickprod` are dummies, I did not get this comment. Would you mind to rephrase it/explain it deeper? What's *the old way*?

PR #173, opened 2018-06-05
XOR constraint + misc

ramabile:
`addConsXor` as new constraint (MAIN CHANGE)
changed `test_logical.py`: two different testing functions, removed trivial skips, cosmetics
examples for and/or/xor constraints (and a workaround to get xor resultant as variable)
example to get if a integer number is even or odd (for newcomers)
ignorefix

Question: what is the boolean `rhs` in http://scip.zib.de/doc-5.0.1/html/cons__xor_8h.php ?
_right-hand side_? it is left-side...
_resultant_? it is not a variable...
I'd like to add its meaning to the docstring.


fserra:
> Question: what is the boolean rhs in http://scip.zib.de/doc-5.0.1/html/cons__xor_8h.php ?
right-hand side? it is left-side...
resultant? it is not a variable...
I'd like to add its meaning to the docstring.

See here. It is the right-hand side. I guess it would make more sense to write it as the rhs ;)
http://scip.zib.de/doc-5.0.1/html/group__CONSHDLRS.php#ga6d8ed4fefe3d67731f636c06e580750e

examples/finished/basiceven.py
```diff
@@ -0,0 +1,51 @@
...
+################################################################################
+# EVEN OR ODD?
+#
+# This example is made for newcomers and motivated by:
+# - modulus is unsupported for pyscipopt.scip.Variable and int
+# - negative integer numbers are not default
+# Based on this:
+# https://github.com/SCIP-Interfaces/PySCIPOpt/issues/172#issuecomment-394644046
+#
+################################################################################
```
fserra: can you say here what this example does please?

ramabile:
I am confused. :/
Can a right-hand side be at the left-hand side? (e.g. [here](http://scip.zib.de/doc-5.0.1/html/cons__xor_8h.php))

examples/finished/basiceven.py
```diff
@@ -0,0 +1,51 @@
...
+sdic = {0:"even",1:"odd"}
+
+def parity(number):
+    assert number == int(round(number))
+    m = Model()
+    m.hideOutput()
+
+    ### "None" means -infinity as lower bound
+    ### N.B.: 0 is the default lower bound
+    ### thus negative numbers are not allowed
```
fserra: what do you mean `thus negative numbers are not allowed` ? you can write `lb = -2`. Maybe is better to write
`m.addVar("x", vartype="I", lb = None, ub = None)` ?

examples/finished/logical.py
```diff
@@ -0,0 +1,64 @@
+from pyscipopt import Model
```
fserra: can you write a comment saying what this should do, please?

Also, wouldn't make more sense to put this as a test? I guess examples are more for applications, something like solving a sudoku or whatnot.

fserra:
in this case it doesn't really matter since `a = b` if and only if `b = a`

ramabile:
**On rhs**
Lhs and Rhs are associated to Model method getLhs and getRhs. Albeit left-hand in the formula, I would expect the boolean value rhs should be returned with getRhs if in any future XOR constraint supports getRhs/getLhs. Or am I missing something?

**On and/or/xor example**
It is a tutorial example for newcomers with a little tip if one wants to set a XOR constraint with a resultant integer variable instead of a boolean rhs. (I used the same tip in `test_logical.py`)

**On integer parity example**
1. as caveat, to show variables are non-negative by default;
2. to point there is no explicit method/error to get the minimum set of constraints for which the problem is infeasible;
3. there is no explicit modulus (%) operator, so an auxiliary variable is needed.

**On example policy**
If examples are application showcases, may I suggest a folder for tutorial examples to help newcomers? A showcase may not be the best to start and may not be the best place to stress caveats, etc.

fserra:
> Or am I missing something?

Ok, I see what you mean. The model method `getRhs` will not work for Xor constraints. SCIP has different types of constraints and each of them provides its own method to query them. We have only interfaced the behavior for linear and quadratic constraints.  
One would need to extend `getRhs` with something like
```python
        elif constype == 'quadratic':
            return SCIPgetRhsQuadratic(self._scip, cons.cons)
        elif constype == 'xor':
            return SCIPgetRhsXor(self._scip, cons.cons)
```

and this would return the boolean, even though in the documentation it is written at the left side.  
On the other hand, the method `getLhs` will never work for Xor constraints, because there is no lhs. There is no `SCIPgetLhsXor`. One could return the variables, but this would differ from SCIP.  
In any case, I am unsure whether it is a good idea to extend `getRhs` so that it also returns the boolean for Xor constraints.  

Regarding examples and tutorial, maybe we can have a directory example and another, applications, for showcases? Or inside example, we can have advanced and basic? @mattmilten any opinions?

mattmilten:
We should definitely clean up the examples directory. I really like the separation of examples/applications into beginner and advanced.

fserra:
ok, I would open a different issue for that. I guess this is good to go

Issue #176, opened 2018-06-08
clean examples


fserra:
On top of #49, we should differentiate between basic and advanced examples. The idea would be that the basic ones could serve as tutorials for how to set up something: a constraint, use re-optimization, etc.
The advanced ones would be more in the spirit of what we currently have as examples, that is, small applications showing how to solve some problem.

fserra:
Yes, you are right, I also don't like advanced and basic.    
Maybe is just best to add a directory called `tutorial` with the "basic" examples.

ramabile:
So `finished`, `unfinished` and `tutorial`?
May I update my last open pull request or should I make a new one?
(the latter means branching my fork at some point? github newcomer here...)

PR #177, opened 2018-06-08
getters for variables

ramabile:
Dear SCIPers,
Here I introduce getters of variables and their values of best known solution since I often find myself querying them. I think it is a nice little feature to slightly improve the user experience.


src/pyscipopt/scip.pyx
```diff
@@ -887,6 +887,7 @@ cdef class Model:
         if lb is None:
             lb = -SCIPinfinity(self._scip)
         cdef SCIP_VAR* scip_var
+        vtype = vtype.upper()
```
mattmilten: Nice!

tests/test_getvars.py
```diff
@@ -0,0 +1,35 @@
...
+def test_getvars(idxvar):
+    m, vs = setModel()
+    try:
+        vs_ = [vs[i] for i in idxvar]
+    except TypeError:
+        if idxvar is None:
+            vs_ = idxvar
+        else:
+            vs_ = vs[idxvar]
+    print(m.getVals(vs_))
```
mattmilten: Shouldn't there be an assert to check the correct value of the variables? Same applies to `test_getvals()` above. Or am I missing some py.test magic here?

ramabile:
#176 and #178 considered.

ramabile:
About testing, I just print variables: if something goes wrong an error is raised.
I am looking for suggestions for other tests of these new methods.

fserra:
Hi @ramabile 

I am sorry, we have failed in writing a decent contributing guide. Most of the methods here make sense and we will include them. However, we will probably change a bit the implementation given that some of the functions actually exist in SCIP (like finding a variable by name). Maybe we can sit down and write a proper contributing guide next week! I propose we also have a look at this merge request then.

ramabile:
I read this pull request with love... since I put different stuff together! which in principle would have required at least 3 branches.

So, first of all: sorry for the mess!

Secondly, #210 (merged) and #211 (pending) overlap with this pull request.
What remains out are these methods:
1- `getVarsByAttr` which `getVarsByName` is based on;
2- `getVarsByReturn` which `getVarsByVtype` and `getVarsByDegree` are based on;
3- `getVals` which works on `getVal`.
4- and their associated tests.
Whereas, after some time, I see 1) and 2) as an exercise and I do not know their practical every-day purpose, in my very humble opinion I think 3) may be newcomer's lifeline.
The very first time I met SCIP/PySCIPOpt was on [Wikipedia categories](https://en.wikipedia.org/wiki/Category:Mathematical_optimization_software) without any background. I was the python interface and installed it just out of curiosity in a period when I was trying hopelessly to solve a problem. I cannot recall exactly why, but I had a very hard time as a total newcomer to understand how to get the values of the optimized model.
So, what I am talking about is a convenience functions according to the [contributing guide](https://github.com/SCIP-Interfaces/PySCIPOpt/blob/master/CONTRIBUTING.rst#design-principles-of-pyscipopt). To me, now, it is not *doubtless beneficial*: I think there is no hurt for a list comprehension. But it could be very helpful for who puts their hands on the code for the very first time.

ramabile:
Closed by #217 through #216.

PR #181, opened 2018-06-19
String conversion in setStringParam()

gasse:
Safety check both name AND value before calling SCIPsetStringParam().


fserra:
thanks!

PR #182, opened 2018-06-20
SCIP 6.0

mattmilten:


mattmilten:
The failing appveyor test can safely be ignored since this does not pull the current SCIP 6.0 release candidate.

PR #189, opened 2018-06-30
CONTRIBUTING improvement

ramabile:
Dear SCIP workers,
here I propose to make the contributing guide clearer for strangers. The guide integrates my notes on the discussion we had at the ZIB. Feel free to edit it and make it complying that discussion.
Cheers!

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -3,15 +3,44 @@ Contributing to PySCIPOpt
 
 Code contributions are very welcome and should comply to a few rules:
 
-1. Compatibility with both Python-2 and Python-3
+0. Read PySCIPOpt design principles. (see further)
 
-#. All tests defined in the Continuous Integration setup need to pass:
+1. Compatibility with both Python-2 and Python-3.
 
-    - `.travis.yml <.travis.yml>`__
-    - `appveyor.yml <appveyor.yml>`__
+2. All tests defined in the Continuous Integration setup need to pass:
 
-#. New features should be covered by the tests, so please extend `tests <tests>`__
+    - `.travis.yml <.travis.yml>`__;
+    - `appveyor.yml <appveyor.yml>`__.
 
-#. New code should be documented in the same style as the rest of the code
+3. New features should be covered by tests *and* examples, so please extend `tests <tests>`__ and `examples <examples>`__. Tests must be implemented with pytest and examples must be as clear as possible for PySCIPOpt newcomers (even advanced examples). **N.B.**: *this is not optional!*
 
-#. New code should be `pep8-compliant <https://www.python.org/dev/peps/pep-0008/>`__. Help yourself with the `style guide checker <https://pypi.org/project/pep8/>`__.
+4. New code should be documented in the same style as the rest of the code.
+
+5. New code should be `pep8-compliant <https://www.python.org/dev/peps/pep-0008/>`__. Help yourself with the `style guide checker <https://pypi.org/project/pep8/>`__.
+
+6. Before implementing a new PySCIPOpt feature, check whether the feature exists in SCIP. If so, implement it as a pure wrapper, mimicking SCIP whenever possible. If the new feature does not exist in SCIP but it is close to an existin
+g one, consider if implementing that way is substantially convenient (e.g. Pythonic). If it does something completely different, you are welcome to pull your request and discuss the implementation.
+
+For general reference, we suggest:
+
+    - `PySCIPOpt README <README.rst>`__;
+    - `SCIP documentation <http://scip.zib.de/doc/html/>`__;
+    - `SCIP mailing list <https://listserv.zib.de/mailman/listinfo/scip/>`__ which can be easily searched with search engines (e.g. `Google <http://www.google.com/#q=site:listserv.zib.de%2Fpipermail%2Fscip>`__);
+    - `open and closed PySCIPOpt issues <https://github.com/SCIP-Interfaces/PySCIPOpt/issues?utf8=%E2%9C%93&q=is%3Aissue>`__;
+    - `SCIP/PySCIPOpt Stack Exchange <https://stackoverflow.com/questions/tagged/scip>`__.
+
+If you find this contributing guide unclear, please open an issue! :)
+
+Design principles of PySCIPOpt
+==============================
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.rst
```diff
@@ -3,15 +3,44 @@ Contributing to PySCIPOpt
 
 Code contributions are very welcome and should comply to a few rules:
 
-1. Compatibility with both Python-2 and Python-3
+0. Read PySCIPOpt design principles. (see further)
```
mattmilten: Use rst link to section
```
Read `Design principles of PySCIPOpt`_.
```

CONTRIBUTING.rst
```diff
@@ -3,15 +3,44 @@ Contributing to PySCIPOpt
...
-#. All tests defined in the Continuous Integration setup need to pass:
+1. Compatibility with both Python-2 and Python-3.
 
-    - `.travis.yml <.travis.yml>`__
-    - `appveyor.yml <appveyor.yml>`__
+2. All tests defined in the Continuous Integration setup need to pass:
 
-#. New features should be covered by the tests, so please extend `tests <tests>`__
+    - `.travis.yml <.travis.yml>`__;
+    - `appveyor.yml <appveyor.yml>`__.
```
mattmilten: Why the trailing comma/semicolon?

CONTRIBUTING.rst
```diff
@@ -3,15 +3,44 @@ Contributing to PySCIPOpt
...
 
-#. New code should be documented in the same style as the rest of the code
+3. New features should be covered by tests *and* examples, so please extend `tests <tests>`__ and `examples <examples>`__. Tests must be implemented with pytest and examples must be as clear as possible for PySCIPOpt newcomers (even advanced examples). **N.B.**: *this is not optional!*
```
mattmilten: That sounds pretty harsh. I would rather have the last sentence removed.

CONTRIBUTING.rst
```diff
@@ -3,15 +3,44 @@ Contributing to PySCIPOpt
...
 
-#. New code should be `pep8-compliant <https://www.python.org/dev/peps/pep-0008/>`__. Help yourself with the `style guide checker <https://pypi.org/project/pep8/>`__.
+4. New code should be documented in the same style as the rest of the code.
+
+5. New code should be `pep8-compliant <https://www.python.org/dev/peps/pep-0008/>`__. Help yourself with the `style guide checker <https://pypi.org/project/pep8/>`__.
+
+6. Before implementing a new PySCIPOpt feature, check whether the feature exists in SCIP. If so, implement it as a pure wrapper, mimicking SCIP whenever possible. If the new feature does not exist in SCIP but it is close to an existin
+g one, consider if implementing that way is substantially convenient (e.g. Pythonic). If it does something completely different, you are welcome to pull your request and discuss the implementation.
```
mattmilten: There is a typo with the linebreak within a word.

CONTRIBUTING.rst
```diff
@@ -3,15 +3,44 @@ Contributing to PySCIPOpt
...
+
+For general reference, we suggest:
+
+    - `PySCIPOpt README <README.rst>`__;
+    - `SCIP documentation <http://scip.zib.de/doc/html/>`__;
+    - `SCIP mailing list <https://listserv.zib.de/mailman/listinfo/scip/>`__ which can be easily searched with search engines (e.g. `Google <http://www.google.com/#q=site:listserv.zib.de%2Fpipermail%2Fscip>`__);
+    - `open and closed PySCIPOpt issues <https://github.com/SCIP-Interfaces/PySCIPOpt/issues?utf8=%E2%9C%93&q=is%3Aissue>`__;
+    - `SCIP/PySCIPOpt Stack Exchange <https://stackoverflow.com/questions/tagged/scip>`__.
+
+If you find this contributing guide unclear, please open an issue! :)
```
mattmilten: :+1:

CONTRIBUTING.rst
ramabile: I removed them. Someone likes ending items of bullet lists with semicolons and commas, and the very last item with a period.

CONTRIBUTING.rst
```diff
@@ -3,15 +3,48 @@ Contributing to PySCIPOpt
...
-#. All tests defined in the Continuous Integration setup need to pass:
+1. Compatibility with both Python-2 and Python-3.
+
+2. All tests defined in the Continuous Integration setup need to pass:
 
     - `.travis.yml <.travis.yml>`__
     - `appveyor.yml <appveyor.yml>`__
 
-#. New features should be covered by the tests, so please extend `tests <tests>`__
+3. New features should be covered by tests *and* examples. Please extend `tests <tests>`__ and `examples <examples>`__. Tests uses pytest and examples are meant to be accessible for PySCIPOpt newcomers (even advanced examples).
```
fserra: Hi

I do not know if it is clear here what "feature" means. In point 6 a distinction is made between different types of features is given. We certainly do not require tests for wrapping basic SCIP code. In the design principles is the first place where convenience functions are mentioned which are the functions that definitely need tests an examples

So I think it would make more sense to try to explicit this further. In particular, the non-optionality of test and examples for convenience functions (I know @mattmilten suggested to remove that sentence, but I think that it is ok if it is explicitly mentioned that it is for convenience functions)

mattmilten: I say, we don't put it in as mandatory and just hope that people add tests. If not we can always just ask for (more) tests in the PR.

ramabile: I look for suggestions about "features": maybe it is just wording, maybe it is not... I used the term "feature" from the discussion in Berlin.

About testing: are you sure we do not need tests for pure wrapping functions? I think these functions have an pretty defined expected behaviour (SCIP!), therefore it must be tested to comply with that behaviour.

mattmilten: I agree with @fserra that simple wrapper functions do not require testing. We should assume that SCIP works correctly and only test those parts that PySCIPOpt adds on top.

ramabile: The assumption that SCIP works correctly does not ensure PySCIPOpt pure wrapper to works as correctly as SCIP. Right?

mattmilten: True, I suppose you can never guarantee that. I still think that we do not need to enforce those tests. If people add them by themselves it's perfectly fine, of course.

PR #193, opened 2018-07-13
add methods for probing and diving

fserra:
extends the interface by, mainly adding some probing and diving functions. It also allows to check the status of the lp solve and allows to create empty rows of unspecified origin.

I think I will change this row stuff in a future pull request. The idea would be that we just give a linear expression and internally this is transformed into a SCIP_ROW


mattmilten:
Apart from the missing tests, this looks good.

src/pyscipopt/scip.pyx
```diff
@@ -1010,6 +1024,9 @@ cdef class Model:
         return [Variable.create(_vars[i]) for i in range(_nvars)]
 
     # LP Methods
+    def getLPSolstat(self):
```
mattmilten: The comment here is missing.

fserra: done

PR #196, opened 2018-07-27
adds BendersDecomp class and update lower bound function

stephenjmaher:
A Benders' decomposition class has been added to handle the functions that are directly related to the Benders' decomposition framework of SCIP. These are the functions that only take a Benders' pointer and no SCIP pointer.

In addition, the update lower bound function has been added so that a user can provide a lower bound on the auxiliary variables before solving the problem.

### DIFF
--- a/src/pyscipopt/scip.pyx
+++ b/src/pyscipopt/scip.pyx
@@ -2170,7 +2170,7 @@ cdef class Model:
                 PY_SCIP_CALL(SCIPfreeBendersSubproblem(self._scip, _benders[i],
                     j))
 
-    def updateBendersLowerbounds(self, lowerbounds, Benders benders = None):
+    def updateBendersLowerbounds(self, lowerbounds, Benders benders=None):
         """"updates the subproblem lower bounds for benders using
         the lowerbounds dict. If benders is None, then the default
         Benders' decomposition is updated
--- a/src/pyscipopt/scip.pyx
+++ b/src/pyscipopt/scip.pyx
@@ -2114,7 +2114,7 @@ cdef class Model:
         subproblems -- a single Model instance or dictionary of Model instances
         """
         cdef SCIP** subprobs
-        cdef SCIP_BENDERS* scipbenders
+        cdef SCIP_BENDERS* benders
 
         # checking whether subproblems is a dictionary
         if isinstance(subproblems, dict):
@@ -2136,8 +2136,8 @@ cdef class Model:
 
         # creating the default Benders' decomposition
         PY_SCIP_CALL(SCIPcreateBendersDefault(self._scip, subprobs, nsubproblems))
-        scipbenders = SCIPfindBenders(self._scip, "default")
-        pyBenders = BendersDecomp.create(scipbenders)
+        benders = SCIPfindBenders(self._scip, "default")
+        pyBenders = BendersDecomp.create(benders)
 
         # activating the Benders' decomposition constraint handlers
         self.setBoolParam("constraints/benderslp/active", True)
--- a/src/pyscipopt/scip.pyx
+++ b/src/pyscipopt/scip.pyx
@@ -557,25 +557,6 @@ cdef class Constraint:
         constype = bytes(SCIPconshdlrGetName(SCIPconsGetHdlr(self.cons))).decode('UTF-8')
         return constype == 'quadratic'
 
-cdef class BendersDecomp:
-    cdef SCIP_BENDERS* benders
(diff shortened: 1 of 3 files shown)


src/pyscipopt/scip.pyx
```diff
@@ -557,6 +557,25 @@ cdef class Constraint:
         constype = bytes(SCIPconshdlrGetName(SCIPconsGetHdlr(self.cons))).decode('UTF-8')
         return constype == 'quadratic'
 
+cdef class BendersDecomp:
```
mattmilten: Why not call it Benders? This would make it consistent with the C naming.

mattmilten: And there should be a doc string for this class

src/pyscipopt/scip.pyx
```diff
@@ -2095,6 +2114,7 @@ cdef class Model:
         subproblems -- a single Model instance or dictionary of Model instances
         """
         cdef SCIP** subprobs
+        cdef SCIP_BENDERS* scipbenders
```
mattmilten: This should also be called `benders`

src/pyscipopt/scip.pyx
stephenjmaher: The reason that it is not called `Benders`, is that the class wrapping around the C plugin type is called `Benders` (in benders.pxi). So there would be a clash with the naming. I am happy for a different name, that was the best that I could come up with.

mattmilten: I see. So there is a generic benders `BendersDecomp` already included in SCIP and people should have the possibility to write their own custom benders decomposition based on the plugin type `Benders`? What about calling `BendersDecomp` `BendersDefault` then?

stephenjmaher: Your comment just made me see a limitation of my approach. I created `BendersDecomp` to allow users to easily have access to public Benders functions, such as `updateLowerbounds`. Now, if the user creates their own `Benders` plugin, then they won't have access to these functions.

Would something like this work? The function `updateLowerbounds` is changed to `updateBendersLowerbounds` and in a member function of `Model`. The arguments of `updateBendersLowerbounds` is a dictionary `lowerbounds` and the `name` of the Benders' decomposition. `name` defaults to `default`, which then means that the default Benders' decomposition is called. In `updateBendersLowerbounds`, the C struct `SCIP_BENDERS` is retrieved by calling `SCIPfindBenders`.

Does that sounds reasonable?

mattmilten: So you wouldn't need a new class `BendersDecomp` at all, right? Yes, this sounds very reasonable. Still, I find it unintuitive to pass the *name* of my custom benders when I have the instance of the class readily available.

stephenjmaher: Yes, we would remove `BendersDecomp`.

It is a little unintuitive. Do have an idea on how to pass the `SCIP_BENDERS` pointer to the C function if we are using the default Benders' decomposition? If we have a custom Benders' decomposition python object, then we could request that as an input. But the default Benders' decomposition plugin doesn't have a python object.

mattmilten: Can't we generate a PyBenders from the default benders whenever the user does not specify his own custom benders?

mattmilten: Wait, do we even need to generate a PyBenders? isn't is enough to have the C pointer from `findBenders()` lying around? So either you ask for the C pointer of the default benders or you take the pointer that comes with your custom benders instance.

stephenjmaher: That makes sense. So in `updateBendersLowerbounds`, we take a `Benders` object as a parameter, which defaults to None. If `Benders == None`, then we just use the default Benders' decomposition, otherwise we use the `SCIP_BENDERS` pointer from `Benders`.

Good idea.

stephenjmaher: I have made some changes. I was unsure about setting the default parameter of `benders` to `None` in `updateBendersLowerbounds`.

mattmilten: Looks good to me. Excpet you might want to remove the spaces around the asignment in the parameters list. It seems, this is against PEP8 style.

stephenjmaher: How is this?

mattmilten: Awesome!

mattmilten:
Shouldn't we have merged this already, @fserra?

PR #199, opened 2018-08-15
Add function to redirect SCIP output via Python

stan227:
compare issue https://github.com/SCIP-Interfaces/PySCIPOpt/issues/116

### DIFF
--- a/src/pyscipopt/scip.pxd
+++ b/src/pyscipopt/scip.pxd
@@ -379,6 +379,9 @@ cdef extern from "scip/scip.h":
     ctypedef struct SCIP_MESSAGEHDLR:
         pass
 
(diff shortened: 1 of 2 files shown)


mattmilten:
Actually, I'm not sure whether it's worth the effort to put it into a separate `.pxi`. I cannot imagine someone wanting to write another message handler,

mattmilten:
BTW, while I didn't like the `from __future__ import ...` at first, it seems to be necessary even for Python-3 because Cython complains and refuses to compile the file.

stan227:
@mattmilten That's weird, for me cython does the same, although the command works in Python3. Another way to do it may be using sys.stdout.write and sys.stderr.write. If the output of these commands are visible in jupyter notebook, we can avoid the __future__ print_function.

PR #206, opened 2018-10-18
add runInstance() method for cluster tests and test file based on gomory

fabian-weg:


mattmilten:
Why did write such a long test? IMHO, they should be as short as possible, only showing the feature in question.

src/pyscipopt/scip.pyx
```diff
@@ -3185,6 +3185,53 @@ cdef class Model:
...
+                problemname = l.split()[1].split()[-1]
+                self.readProblem(problemname)
+            elif l.startswith('set'):
+                paramname = ''
+                for w in range(1, len(l.split())-2):
+                    paramname += l.split()[w] + '/'
+                paramname = paramname[:-1]
+                self.setParam(paramname, l.split()[-1])
+                if paramname == 'limits/objective':
+                    self.setHeuristics(SCIP_PARAMSETTING.OFF)
```
```suggestion
                    self.setHeuristics(SCIP_PARAMSETTING_OFF)
```

mattmilten: This refers to the actual name in SCIP. Only when using PySCIPOpt you write `SCIP_PARAMSETTING.OFF`

Spoetter:
> Why did write such a long test? IMHO, they should be as short as possible, only showing the feature in question.

I didn't really write that test myself. I copied the code from the gomory test and only adjusted the main such that it calls runInstance and can be used for the make scripts. Thought that would be a better test than doing a completely minimal example

mattmilten:
A simple model is perfectly fine since this is not what is supposed to be tested. To be honest, I'm not sure whether we need a test at all since we cannot call the function from `make test` anyway.

mattmilten:
What are we going to do with this, @fserra? I would vote to close it.

fserra:
yes i guess this is no longer needed

PR #209, opened 2018-10-19
allow constants in setObjective, esp. 0 to clear it

mattmilten:


fserra:
oh, why did the test not run?

Issue #178, opened 2018-06-12
AssertionError messages


ramabile:
Dear SCIP workers,
with the goal of a self-explaining interface, when an assertion error is raised it would be very helpful to get a meaningful assertion error. Not everyone goes through the code with a breezy attitude. :)

I am working to give an appropriate message to the ones I find and understand. I would like to keep this issue open to clarify some nasty `assert` statements.

I start with `scip.pyx` which I am more confident with. I list the assertions in increasing order of personal _understanding_:
- `assert op2idx[valuenode[0]] == SCIP_EXPR_CONST` in `_addGenNonlinearCons`;
- `assert len(node[1]) == 1` in `_addGenNonlinearCons`;
- `assert varpos == nvars` in `_addGenNonlinearCons`;
- `assert len(term) == 1` in `setObjective` and `chgReoptObjective`;
- `assert len(v) == 2, 'term: %s' % v` in `_addQuadCons`; (not that clear)
- `assert cons.isQuadratic()` in `getTermsQuadratic`. (yes, it's clear... which assertionError message?)

Out of curiosity: in `addConsIndicator` there are the following:
```
if cons.expr.degree() > 1:
    raise ValueError("expected linear inequality, expression has degree %d" % cons.expr.degree())

assert cons.expr.degree() <= 1
```
Is that intended? Is the check so detrimental that, in case of optimization (_e.g._ `-O` flag) the check should be included? Or is that just an involuntary redundancy?

ramabile:
Closed by https://github.com/SCIP-Interfaces/PySCIPOpt/commit/453482a1752d87fe4b21276eb2ab3ef388526a5b

mattmilten:
Cool, thanks!

PR #210, opened 2018-10-31
explain assertions + .gitignore fix + cosmetics

ramabile:
Dear SCIPers,
I am back in town.

The pull request involves:
1. self-explaining assertions (but PLEASE see #178 )
2. I shot a very tricky typo in .gitignore (now I understand why!) and other model file from testing;
3. cosmetics

Feedbacks are warmly welcome, as usual.

### DIFF
--- a/src/pyscipopt/scip.pyx
+++ b/src/pyscipopt/scip.pyx
@@ -1969,7 +1969,7 @@ cdef class Model:
         cdef int _nquadterms
         cdef int _nlinvars
 
-        assert cons.isQuadratic()
+        assert cons.isQuadratic(), "constraint is not quadratic"
 
         bilinterms = []
         quadterms  = []
--- a/src/pyscipopt/scip.pyx
+++ b/src/pyscipopt/scip.pyx
@@ -779,7 +779,7 @@ cdef class Model:
 
         # turn the constant value into an Expr instance for further processing
         if not isinstance(coeffs, Expr):
-            assert(_is_number(coeffs))
+            assert(_is_number(coeffs)), "given coefficient is neither ExprCons or number but %s" % coeffs.__class__.__name__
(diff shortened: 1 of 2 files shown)


src/pyscipopt/scip.pyx
```diff
@@ -1694,7 +1695,7 @@ cdef class Model:
         if cons.expr.degree() > 1:
             raise ValueError("expected linear inequality, expression has degree %d" % cons.expr.degree())
 
-        assert cons.expr.degree() <= 1
+        assert cons.expr.degree() <= 1 # redundant with previous if-statement?
```
```suggestion
```

ramabile:
From #178:
- `assert op2idx[valuenode[0]] == SCIP_EXPR_CONST` in `_addGenNonlinearCons`;
- `assert len(node[1]) == 1` in `_addGenNonlinearCons`;
- `assert varpos == nvars` in `_addGenNonlinearCons`;
- `assert len(term) == 1` in `setObjective` and `chgReoptObjective`;
- `assert len(v) == 2, 'term: %s' % v` in `_addQuadCons`; (not that clear) [NdR: ok now it's clear]
- `assert cons.isQuadratic()` in `getTermsQuadratic`. (yes, it's clear... which assertionError message?)
Any suggestion? How to explain them?

mattmilten:
So,I suppose for the asserts in `_addGenNonlinearCons()` we don't need user friendly messages, because all `_blablabla()` methods should be invisible to the user and are only internal methods. If one of those asserts fail, you **have** to look into the code...

Concerning the two `len(v) == 1` asserts, I would say, we don't need descriptive error messages as they are checking for internal/derived data structures and not direct user input.

For `getTermsQuadratic()` a simple: `cons is not quadratic` should be sufficient.

PR #214, opened 2018-11-01
upgrade to general expression only when necessary

### LINKED ISSUE #213: Division by float: convert ProdExpr -> Expr
Dear SCIPers,
Following this https://github.com/SCIP-Interfaces/PySCIPOpt/issues/212#issuecomment-435042087 I am here to ask for a clarification.

Consider:
```python
from pyscipopt import Model
model = Model()
x = model.addVar(vtype="c")
print((2*x).__class__.__name__)
# Expr
print((x/2.).__class__.__name__)
# ProdExpr
print((2.*x).__class__.__name__)
# Expr
print((x/2).__class__.__name__)
# ProdExpr
```
It seems like `ProdExpr` is not casted as `Expr` in any case the division is defined. This provokes:
```python
model.setObjective(.5*x) # accepted
model.setObjective(x/2.) # not accepted
# AssertionError: given coefficient is neither Expr or number but ProdExpr
```
Is this a design choice? Is SCIP behaviour?

As a Python user, I would expected a full support of the division of a continuous variable with a float number.
In other cases like integer variable divided by float or continuous variable divided by integer, I would expect a division btw. floats since integers are casted to floats in these cases.
In case of division between an integer variable and an integer number I would keep the `ProdExpr` until there is no support for division with remainders (that is why I wrote [this](https://github.com/SCIP-Interfaces/PySCIPOpt/pull/211/files#diff-3ff183b9d68ae9718e7ecc814d299d67)).

fserra:
Closes #213

### DIFF
--- a/src/pyscipopt/expr.pxi
+++ b/src/pyscipopt/expr.pxi
@@ -216,20 +216,32 @@ cdef class Expr:
 
     def __div__(self, other):
         ''' transforms Expr into GenExpr'''
+        if _is_number(other):
+            f = 1.0/float(other)
+            return f * self
         selfexpr = buildGenExprObj(self)
         return selfexpr.__div__(other)
 
     def __rdiv__(self, other):
         ''' other / self '''
+        if _is_number(self):
+            f = 1.0/float(self)
+            return f * other
         otherexpr = buildGenExprObj(other)
         return otherexpr.__div__(self)
 
     def __truediv__(self,other):
+        if _is_number(other):
+            f = 1.0/float(other)
+            return f * self
         selfexpr = buildGenExprObj(self)
         return selfexpr.__truediv__(other)
 
     def __rtruediv__(self, other):
         ''' other / self '''
+        if _is_number(self):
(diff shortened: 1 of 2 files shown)


ramabile:
Misc:
- `Exception TypeError: "'NoneType' object is not callable" in 'pyscipopt.scip.Model.__dealloc__' ignored` each time I run tests. Do you know what raises it?
- I should make `test_logical.py` more compact...

fserra:
I am not sure why the error at `__dealloc__` appears. I guess we still have some issue with when the freeing of what happens.

I would say `int/int` is float. For instance, `x/2` and `x/2.0` should both be  expressions, right?

ramabile:
Ok I see now, it makes sense.

Python3 makes `3/2==1.5`, Python2 `3/2==1`. There is no support for `3//2`.
If one wants the floor division, they realises soon `x//2` is unsuported and finds a workaround like [this](https://github.com/SCIP-Interfaces/PySCIPOpt/blob/master/examples/tutorial/even.py#L68).

PR #211, opened 2018-10-31
examples: cleaner and tutorials

ramabile:
Dear SCIPers,
this pull request follows @fserra's #176 suggestion and moves my very basic examples + `puzzle.py` to the new `examples/tutorial`  folder.

`even.py` adds hints on how to encode a constraints with integer divisions. 
`logical.py` helps to clarify how SCIP works with logical operators and its (at the moment) unconsistent logical constraints.

Feedbacks welcome!

PR #216, opened 2018-11-02
getVals + test

ramabile:
Close #177. See https://github.com/SCIP-Interfaces/PySCIPOpt/pull/177#issuecomment-434752450 .
A convenient function for newcomers. Feedbacks and criticism very welcome.

### DIFF
--- a/src/pyscipopt/scip.pyx
+++ b/src/pyscipopt/scip.pyx
@@ -2515,6 +2515,22 @@ cdef class Model:
             raise Warning("method cannot be called before problem is solved")
         return self.getSolVal(self._bestSol, var)
 
+    def getVals(self, vars=None, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


fserra:
I am a bit confused. Doesn't this PR depends on code from #177 ?

There is a SCIP function `SCIPgetSolVals` that can be used to get the values of variables.

I think that if we want to make it easier for newcomers, then more examples are better than more functions. Thoughts, @mattmilten  ?

ramabile:
I divided the old PR #177 in three PRs:
- #210, merged 
- #211, merged
- #216, this one.
This PR does not depend on #177.

Anyway, thank you for the suggestion! Yes, we should go for `SCIPgetSolVals` -> `m.getSolVals`. Since it is not implemented, I open an issue.

ramabile:
Closed by #217.

PR #224, opened 2018-11-28
conslock changed

fserra:
we should probably run the examples  before releasing new scip versions


mattmilten:
What a radical thought! ;-)

Seriously, why don't we run the examples in the integration tests?

PR #231, opened 2019-01-10
add function that computes correct dual multiplers when called within…

fserra:
…g scip


src/pyscipopt/scip.pyx
```diff
@@ -2255,6 +2255,21 @@ cdef class Model:
             valsdict[bytes(SCIPvarGetName(_vars[i])).decode('utf-8')] = _vals[i]
         return valsdict
 
+    def getDualMultiplier(self, Constraint cons):
+        """Retrieve the dual solution to a linear constraint.
```
mattmilten: maybe change the docu:

```suggestion
        """Retrieve the dual multiplier to a linear constraint.
```

PR #240, opened 2019-02-04
relaxator interface included

bzfmuelh:


mattmilten:
It looks like we need to install numpy for the new tests to pass on the CI

bzfmuelh:
numpy is not really necessary for the test, so I just removed it from the test

PR #241, opened 2019-02-11
Benders plugin bug

### LINKED ISSUE #237: Problems on benders decomposition plugin
Hello everybody, recently I am implementing a problem with both benders decomposition and column generation. To write in SCIP format, I implement the benders class and pricer class respectively, hoping to solve the problem using column generation first then call benders decomposition on several subproblems. However, when running prob.optimize(), I observe "pricerinit"->"benderssolvesub"->"bendersgetvar" were invoked sequentially. But no column generation was performed until the program ended. I suppose this is due to no variable mapping is found between master problem and subproblems (as master problem's variables are added during solving). My **questions** follows:

-  Is there a way to perform the two things at the same time? I try to adjust the pricer's priority but failed.

-  When I tested the bendersgetvar, I found the mappedvar I provided (using a dict from master problems variable to subproblem variables) failed with the following problem:

```
AttributeError: "'pyscipopt.scip.Variable' object has no attribute 'var'" in 'pyscipopt.scip.PyBendersGetvar' ignored
```
I guess maybe the interface function got something wrong? 

Hope somebody can help me with the following doubts, thanks in advance!

P.S. I implement the SCIPactivateBenders and SCIPaddBendersSubproble to make SCIP aware the subproblems.

stephenjmaher:
Closes #237.

The `bendersgetvar` function was not working correctly. This has been fixed in this pull request.

### DIFF
--- a/src/pyscipopt/__init__.py
+++ b/src/pyscipopt/__init__.py
@@ -3,6 +3,7 @@ __version__ = '2.1.1'
 # export user-relevant objects:
 from pyscipopt.Multidict import multidict
 from pyscipopt.scip      import Model
+from pyscipopt.scip      import Benders
 from pyscipopt.scip      import Branchrule
 from pyscipopt.scip      import Conshdlr
 from pyscipopt.scip      import Eventhdlr
--- a/src/pyscipopt/benders.pxi
+++ b/src/pyscipopt/benders.pxi
@@ -28,16 +28,16 @@ cdef class Benders:
         return {}
 
     def benderspresubsolve(self, solution, enfotype, checkint):
-        pass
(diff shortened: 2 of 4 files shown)


mattmilten:
I don't want to nag but could you maybe add a short test that shows how the benders plugin works?

stephenjmaher:
No, I don't really have any example to add for this. We don't have a precedent of adding tests for plugins.

mattmilten:
Well, there is `test_benders` but it isn't using the benders plugin.

mattmilten:
But yes, you are right, there was no test before either, so I suppose we can just merge this.

PR #246, opened 2019-02-21
add SCIPcopy functionality

mattmilten:
Some things are not copied, for example the objective offset - this is why the test fails.

### DIFF
--- a/src/pyscipopt/scip.pyx
+++ b/src/pyscipopt/scip.pyx
@@ -599,6 +599,7 @@ cdef void relayErrorMessage(void *messagehdlr, FILE *file, const char *msg):
 # - interface SCIPfreeProb()
 cdef class Model:
     cdef SCIP* _scip
+    cdef SCIP_Bool* _valid
     # store best solution to get the solution values easier
     cdef Solution _bestSol
     # can be used to store problem data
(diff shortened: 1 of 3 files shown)


src/pyscipopt/scip.pyx
```diff
@@ -599,27 +599,41 @@ cdef void relayErrorMessage(void *messagehdlr, FILE *file, const char *msg):
...
+        if sourceModel is None:
+            self.create()
+            self._bestSol = None
+            if defaultPlugins:
+                self.includeDefaultPlugins()
+            self.createProbBasic(problemName)
+        else:
+            self.create()
+            self._bestSol = <Solution> sourceModel._bestSol
+            n = str_conversion(problemName)
```
fserra: Maybe you wanna add a "copy" at the beginning?

fserra:
Looks good. I guess though that the usual case here will be to have a copy during the solving process. We should probably check whether this works, like adding a simple heuristic that copies and... don't know

src/pyscipopt/scip.pyx
mattmilten: No, because this could've been already overwritten by the user when calling this method.

PR #254, opened 2019-03-06
automated wget

bzfmuelh:


generate-docs.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+#get SCIP TAGFILE
+wget -O docs/scip.tag https://scip.zib.de/doc-6.0.1/html/scip.tag
```
mattmilten:
```suggestion
wget -O docs/scip.tag https://scip.zib.de/doc/scip.tag
```

mattmilten: This way we are independent of the version number. I set up the SCIP docu to generate the tagfile directly in `doc/`

PR #261, opened 2019-03-14
add addConsLocal() and addConsNode()

### LINKED ISSUE #260: Missing method addConsNode
I'm trying to implement a branch-and-price procedure to solve VRPTW using a set partitioning formulation where each variable represents one valid tour through the graph. I want to branch on the edges of the corresponding graph. Therefore, I save a graph view on each node of the branching tree which only sees the edges which have not been deleted by branching. To make this branching scheme work, variables whose tours have deleted edges in them have to be set to zero in the corresponding subtree of the branching tree. The C interface of SCIP provides the function [`SCIPaddConsNode`](https://scip.zib.de/doc-6.0.1/html/group__LocalSubproblemMethods.php#ga5fcf2b01b3850c5220113b2056b28152) for that; however, I didn't find such a function in PySCIPOpt. What can I do?

mattmilten:
closes #260

### DIFF
--- a/src/pyscipopt/scip.pxd
+++ b/src/pyscipopt/scip.pxd
@@ -576,9 +576,11 @@ cdef extern from "scip/scip.h":
     SCIP_Bool SCIPnodeIsActive(SCIP_NODE* node)
     SCIP_Bool SCIPnodeIsPropagatedAgain(SCIP_NODE* node)
     SCIP_Real SCIPcalcNodeselPriority(SCIP*	scip, SCIP_VAR* var, SCIP_BRANCHDIR	branchdir, SCIP_Real targetvalue)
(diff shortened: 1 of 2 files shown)


lukasstevens:
I think it should be `node.node` instead of `SCIP_NODE* node`.

mattmilten:
> I think it should be `node.node` instead of `SCIP_NODE* node`.

Haha, yes... I was not very concentrated. Should be good now.

lukasstevens:
Thanks for the quick fix. I'll try to finish my implementation asap to let you know if everything works.

mattmilten:
In the end, adding new methods to the interface is often just some copy-paste. So if you see that you need more you can try to add them yourself.

Issue #238, opened 2019-01-31
dropping Python 2


mattmilten:
Since Python 2 will not be maintained past 2020, we will also drop Python 2 support soon. I already enlisted PySCIPOpt here: https://python3statement.org/

mattmilten:
done with release 3.0

PR #264, opened 2019-03-15
remove future warning when running cythonize

mattmilten:


mattmilten:
This will only work after Python 2.7 support is dropped #238

Issue #268, opened 2019-03-25
data storage for all classes


mattmilten:
There should be a `cdef public object data` in every class so users are able to store data in their instances.

mattmilten:
reference: https://stackoverflow.com/questions/55342311/how-to-add-attributes-to-pyscipopt-classes/55343103#55343103

PR #269, opened 2019-03-25
add data attribute to various cython classes for users to store things

networds:
added data attribute for the following classes:
Event, Column, Row, Solution, Node, Variable
and fixed comment for class Constraint for consistency


networds:
implements #268

mattmilten:
Thank you!

PR #262, opened 2019-03-15
Adds the class for Benders' decomposition cutting methods

stephenjmaher:
In response to the [discussion](https://github.com/SCIP-Interfaces/PySCIPOpt/issues/194#issuecomment-467449033), the class for Benders' decomposition cuts has been added to PySCIPOpt. This class is a complete wrapper of the Benders' Cut plugins in SCIP.

Unfortunately, this remains untested because there are currently no test instances for this particular class. This will be updated as the interface is used more.

PR #278, opened 2019-04-17
adds methods for creating a model from an external SCIP instance

stephenjmaher:
An issue that has been found when investigating #262 is how to access data from subscip instances. For example, in the C-API of SCIP the function `SCIPbendersSubproblem` will return a `SCIP*`. This is then queried in the custom Benders' decomposition cut plugins for the primal and dual solutions.

In the Python interface, a `Model` object is required to query the primal and dual solutions.

The added functions are designed to take a `SCIP*` and create a `Model` object so that the solution information can be accessed from Python.

Suggestion for the names of the variables and functions are appreciated.


src/pyscipopt/scip.pyx
```diff
@@ -647,21 +647,33 @@ cdef class Model:
...
-    def __init__(self, problemName='model', defaultPlugins=True, Model sourceModel=None, origcopy=False, globalcopy=True, enablepricing=False):
+    def __init__(self, problemName='model', defaultPlugins=True, Model sourceModel=None, origcopy=False, globalcopy=True, enablepricing=False, noscipcreate=False):
         """
         :param problemName: name of the problem (default 'model')
         :param defaultPlugins: use default plugins? (default True)
         :param sourceModel: create a copy of the given Model instance (default None)
         :param origcopy: whether to call copy or copyOrig (default False)
         :param globalcopy: whether to create a global or a local copy (default True)
         :param enablepricing: whether to enable pricing in copy (default False)
+        :param noscipcreate: initialize the Model object without creating a SCIP instance
```
mattmilten: I don't like the double negative. Why not flip it and call it `createscip`?

tests/test_scipinstancecopy.py
```diff
@@ -0,0 +1,22 @@
...
+    x = s.addVar("x", vtype = 'C', obj = 1.0)
+    y = s.addVar("y", vtype = 'C', obj = 2.0)
+    s.setObjective(4.0 * y, clear = False)
+
+    c = s.addCons(x + 2 * y >= 1.0)
+
+    # solve problems
+    s.optimize()
+
+    s2 = s.createModelFromSCIP()
```
mattmilten: This does not look very "Pythonic". Can't we call it `s.copy()`?

src/pyscipopt/scip.pyx
```diff
@@ -679,12 +691,29 @@ cdef class Model:
...
+        """Creates a model and appropriately assigns the scip and bestsol parameters
+        """
+        if scip == NULL:
+            raise Warning("cannot create Model with SCIP* == NULL")
+        model = Model(noscipcreate=True)
+        model._scip = scip
+        model._bestSol = Solution.create(SCIPgetBestSol(scip))
+        return model
+
+    def createModelFromSCIP(self):
```
mattmilten: Why not call it `copy(self)`?

stephenjmaher:
The main reason for not adding an additional parameter to the `__init__` function was that all other classes have a `create()` function. So I was trying to replicate that.

I understand that this could be confusing. So if you feel that an additional optional parameter is best, then I can work with that.

tests/test_scipinstancecopy.py
stephenjmaher: It is not very "Pythonic", it is more in the style of the PySCIPOpt functions, such as `includeDefaultPlugins` and `createProbBasic`.

`s.copy()` would be confusing because this function doesn't actually copy SCIP, but copies the python `Model`. What do you think about `s.copyModel()` or `s.copy_model()`?

src/pyscipopt/scip.pyx
stephenjmaher: As I mentioned above, this could be confused with the copy methods of the C-API. Suggestion above.

src/pyscipopt/scip.pyx
stephenjmaher: If we have an additional parameter, such as `sourcescip`, then we could could eliminate this parameter. The logic would be if `sourcescip is not None`, then a model is created with the sourcescip as the scip instance.

Does that work for you?

mattmilten: Yes, that's what I had in mind. Similar to how `sourceModel` works. But then the other parameters `origcopy`, `globalcopy` don't make much sense, right...?

tests/test_scipinstancecopy.py
mattmilten: `crateProbBasic` and `includeDefaultPlugins` are simple oneline wrappers and are named after the SCIP functions. My point is that in the best case the user should not even have to know what a `SCIP*` is, let alone work with it. Does it make sense to not return a `SCIP*` at all in the benders code and instead create a Model around it right away?

mattmilten:
> The main reason for not adding an additional parameter to the `__init__` function was that all other classes have a `create()` function. So I was trying to replicate that.

On second thought, I suppose that all those other classes should rather have a proper `__init__` themselves instead of the `create`. What do you think?

src/pyscipopt/scip.pyx
stephenjmaher: No the `origcopy` and `globalcopy` don't make sense because this is not a copy of SCIP.

tests/test_scipinstancecopy.py
stephenjmaher: > Does it make sense to not return a SCIP* at all in the benders code and instead create a Model around it right away?
This is exactly what I am trying to do. A SCIP* is returned and then I want to access the internal data using the Model class.

stephenjmaher:
It would make more sense to remove the `create` functions from all other classes. The create is just an extra method that is not really necessary.

src/pyscipopt/scip.pyx
stephenjmaher: There is a problem with adding a SCIP* as an optional argument to the `__init__` function. SCIP* is a C variable, where the `__init__` function can only take python variables. So, it is not possible to initialise an empty Model with an available SCIP*.

We need to go with a static method that can insert the SCIP* into an empty model.

src/pyscipopt/scip.pyx
```diff
@@ -647,21 +647,33 @@ cdef class Model:
...
             warnings.warn("linked SCIP {} is not recommended for this version of PySCIPOpt - use version {}.{}.{}".format(self.version(), MAJOR, MINOR, PATCH))
-        if sourceModel is None:
+
+        self.freescip = True
+
+        if noscipcreate:
+            # if no SCIP instance should be created, then an empty Model object is created.
+            self._scip = NULL
+            self._bestSol = None
+            self.freescip = False
```
```suggestion
            self._freescip = False
```

src/pyscipopt/scip.pyx
```diff
@@ -647,21 +647,33 @@ cdef class Model:
     cdef public object data
     # make Model weak referentiable
     cdef object __weakref__
+    # flag to indicate whether the SCIP should be freed. It will not be freed if an empty Model was created to wrap a
+    # C-API SCIP instance.
+    cdef SCIP_Bool freescip
```
```suggestion
    cdef SCIP_Bool _freescip
```

mattmilten:
To summarize:

- We need a function to create a Model from an existing SCIP*
-> this would be something like `create(SCIP* scip)` and can be called without a Model instance (`Model.create(scip)`)
-  Why do we need `createModelFromSCIP()`? Can't we just use `create()`?
- The new parameter to create/__init__ a Model without constructing a new SCIP* should be `createscip=True`

stephenjmaher:
> - We need a function to create a Model from an existing SCIP*
    -> this would be something like create(SCIP* scip) and can be called without a Model instance  Model.create(scip))

This is exactly what `createModel` is doing. We can't use `create`, because this is a wrapper for `SCIPcreate`. Is there something wrong with `createModel`?

> - Why do we need createModelFromSCIP()? Can't we just use create()?

No, we can't just use `create` (or `createModel` in the current code). This function is designed to build a new model from the SCIP pointer from within a Model instance. Essentially, this method was created for testing purposes because we are unable to get the SCIP pointer from within python code, see [test_scipinstancecopy.py](../blob/sm/creates-empty-model/tests/test_scipinstancecopy.py). In reality, there is no purpose for this method as you can achieve the same thing by copying the SCIP through the `__init__` method.

> - The new parameter to create/__init__ a Model without constructing a new SCIP* should be createscip=True

Sounds good to me.

fserra:
> > * Why do we need createModelFromSCIP()? Can't we just use create()?
> 
> No, we can't just use `create` (or `createModel` in the current code). `createModelFromSCIP` is designed to build a new model from the SCIP pointer from within a Model instance. Essentially, this method was created for testing purposes because we are unable to get the SCIP pointer from within python code, see [test_scipinstancecopy.py](../blob/sm/creates-empty-model/tests/test_scipinstancecopy.py). In reality, there is no purpose for this method as you can achieve the same thing by copying the SCIP through the `__init__` method.
> 

Then we should remove such a function. It is more confusing than anything.

mattmilten:
I have another issue with the change: Why do we need `freescip`? This will end up in a Model that is destroyed leaving the SCIP* behind, given that we used the proposed way of wrapping the SCIP* in a Model. Since we don't have a `Model.free()`, I don't see how the SCIP* can be freed at all. I suppose this is not done in the Benders part, right?

stephenjmaher:
@fserra I am happy to remove `createModelFromSCIP`; however, this means that we will not be able to have a test for this feature. So we should remove the test also.

stephenjmaher:
@mattmilten We don't have a `Model.free`, but we have `__dealloc__`, which calls `SCIPfree`. So if we create an empty model and then insert the SCIP, the destructor of the model will free the SCIP. Since we never created the SCIP with the model, then this would result in undesired behaviour.

To avoid this I added the `freescip` parameter. This informs the model that the SCIP should be freed in the `__dealloc__` method.

mattmilten:
Sure, but when is the SCIP freed then?

src/pyscipopt/scip.pyx
```diff
@@ -679,12 +691,29 @@ cdef class Model:
     def __dealloc__(self):
         # call C function directly, because we can no longer call this object's methods, according to
         # http://docs.cython.org/src/reference/extension_types.html#finalization-dealloc
-        PY_SCIP_CALL( SCIPfree(&self._scip) )
+        if self._scip is not NULL and self._freescip:
```
```suggestion
        if self._freescip:
```

mattmilten: We don't check this anywhere else but always assume it's not NULL.

stephenjmaher:
Yes, the SCIP instances must be freed somewhere. Since we are working in python, it is only possible to create a SCIP instance using the Model object. Thus, the only way to free the SCIP instance is to delete the associated Model object. So in this case, there must be an associated Model object for the SCIP instance.

It is possible that a user could create the subproblem Model object and then subsequently destroy this object during this solving. However, this would cause the Benders' decomposition framework to not work at all, since we rely on the SCIP instances being available to the Benders' decomposition core. We have no way to ensure that a user doesn't destroy the Model object for the subproblem classes, except through the `weakref` member variable.

The scenario that I am looking at is related to the Benders' decomposition framework.

- A user creates a set of subproblems and registers them with the Benders' decomposition framework with a createBendersXyz function.
- The user also creates a custom Benders' decomposition cut plugin in Python.
- The Benders' decomposition cut plugin exec function is given only the index of the subproblem for which the cut is generated, not the SCIP* instances. Thus, we need to call `SCIPbendersSubproblem(benders, probnumber)` to get the SCIP* to access the appropriate data to generate the cut.
- These subproblems have been created in the main python code and should be stored by the user.  So these SCIP instances are freed whenever at the end of the execution.

stephenjmaher:
The idea to create a Model object to wrap a SCIP instance is the same as for all other objects that are available in PySCIPOpt. While the Model object is more like a **master** object, it shouldn't really be treated differently.

One possibility is to return a reference to the SCIP object so that the user can create a dictionary of Models with the key being this reference. That way the user can access the subproblem Model object from the python dictionary using a returned SCIP pointer. We do something similar with Variable and Constraint objects.

mattmilten:
I replaced the old `Model.create()` with your proposed implementation in `Model.createModel()`, because we do not need a callable `create` - this is done in the `__init__` by calling the wrapped method directly. I also removed other create-method which was only used for testing.

I guess we can have a more meaningful test with a proper Benders' implementation some time in the future.

stephenjmaher:
That looks good to me. When we get a benders example running then we can test this functionality.

PR #283, opened 2019-06-04
add functionality to access SCIP_NLROWs

mueldgog:
I have added a new class for SCIP_NLROWs. This is useful if one needs to have a global view on the problem. Iterating through the constraints is in general not possible because there might be some upgraded constraints that are not supported by the Python interface.

### DIFF
--- a/src/pyscipopt/scip.pyx
+++ b/src/pyscipopt/scip.pyx
@@ -429,15 +429,14 @@ cdef class NLRow:
         return SCIPnlrowGetConstant(self.scip_nlrow)
 
     def getLinearTerms(self):
-        """returns a list of linear terms of a nonlinear row"""
+        """returns a list of tuples (var, coef) representing the linear part of a nonlinear row"""
         cdef SCIP_VAR** linvars = SCIPnlrowGetLinearVars(self.scip_nlrow)
         cdef SCIP_Real* lincoefs = SCIPnlrowGetLinearCoefs(self.scip_nlrow)
         cdef int nlinvars = SCIPnlrowGetNLinearVars(self.scip_nlrow)
         return [(Variable.create(linvars[i]), lincoefs[i]) for i in range(nlinvars)]
 
     def getQuadraticTerms(self):
-        """returns a list of quadratic terms of a nonlinear row"""
-
+        """returns a list of tuples (var1, var2, coef) representing the quadratic part of a nonlinear row"""
         cdef int nquadvars;
         cdef SCIP_VAR** quadvars;
         cdef int nquadelems;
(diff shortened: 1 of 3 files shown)


src/pyscipopt/scip.pyx
```diff
@@ -407,6 +407,71 @@ cdef class Row:
...
+        def __get__(self):
+            cname = bytes( SCIPnlrowGetName(self.scip_nlrow) )
+            return cname.decode('utf-8')
+
+    def getConstant(self):
+        """returns the constant of a nonlinear row"""
+        return SCIPnlrowGetConstant(self.scip_nlrow)
+
+    def getLinearTerms(self):
+        """returns a list of linear terms of a nonlinear row"""
```
fserra: `returns a list of tuples (var, coef)` What about a dict between var.ptr and lincoef

src/pyscipopt/scip.pyx
```diff
@@ -407,6 +407,71 @@ cdef class Row:
...
+        cdef SCIP_VAR** linvars = SCIPnlrowGetLinearVars(self.scip_nlrow)
+        cdef SCIP_Real* lincoefs = SCIPnlrowGetLinearCoefs(self.scip_nlrow)
+        cdef int nlinvars = SCIPnlrowGetNLinearVars(self.scip_nlrow)
+        return [(Variable.create(linvars[i]), lincoefs[i]) for i in range(nlinvars)]
+
+    def getQuadraticTerms(self):
+        """returns a list of quadratic terms of a nonlinear row"""
```
fserra: list of tuples var1 var2 coef

src/pyscipopt/scip.pyx
```diff
@@ -2293,6 +2358,21 @@ cdef class Model:
...
+    def isNLPConstructed(self):
+        """returns whether SCIP's internal NLP has been constructed"""
+        return SCIPisNLPConstructed(self._scip)
+
+    def getNNlRows(self):
+        """gets current number of nonlinear rows in SCIP's internal NLP"""
+        return SCIPgetNNLPNlRows(self._scip)
+
+    def getNlRows(self):
+        """returns the nonlinear rows in SCIP's internal NLP"""
```
fserra: a list with the nonlin..

src/pyscipopt/scip.pyx
mueldgog: d1edb6f

mueldgog:
> looks good. What about expression trees?

I only care about quadratics for now. To be honest, I don't know the expression structure in PySCIPOpt good enough to implement this. I would postpone this and then implement this properly when consexpr has been released.

PR #285, opened 2019-06-07
Correct type signature mismatches in SCIPlpiCreate and SCIPgetObjlimit

mlubin:
The difference between `SCIP_OBJSEN` and `SCIP_OBJSENSE` seems harmless but triggered compiler errors for me. The incorrect return type of `SCIPgetObjLimit` looks like a bug.


mattmilten:
Thanks a lot!

PR #292, opened 2019-07-05
Add node selector plugin

pg2455:
This plugin in based on the code from branchrule.pxi.
Comments are copied from SCIP documentation.


mattmilten:
Looks good to me. Except that there is no test showing that it works. Do you mind adding a basic example?

mueldgog:
The changes look good. I think that I will merge this and add a test by myself.

mueldgog:
After merging I realized that the node selector could not be tested because we don't provide access to the open nodes. I added now `getOpenNodes()` to access all open nodes and added a node selector example to the `test` directory.

PR #300, opened 2019-08-06
use correct string transformation for get- and setParam()

mattmilten:
In the current code, we cannot pass string or char params.


fserra:
did you test it? we could add a test that just sets settings and if it doesn't crash then its good

mattmilten:
Yeah, I tested it. But if we had test, we would have noticed the bug a lot earlier...

mattmilten:
I'll just a add a quick test to `test_model`, ok?

fserra:
ok

fserra:
the build failed because of some travis stuff... do you know if one can rerun a build ?

mattmilten:
I just triggered a new build.

PR #301, opened 2019-08-16
new method to get current node's LP iterations

mattmilten:


mattmilten:
What about `getLPiterations()`?

fserra:
I think we should go with the sad name `lpiGetIterations` just for consistency

PR #302, opened 2019-08-16
Adapt setup to other SCIP installations

pfetsch:
This PR should make it possible to also use SCIP that is not installed at a central place.
The code should now work both with SCIP directories that are built using cmake as well as make.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -60,6 +60,9 @@ if "--debug" in sys.argv:
     extra_compile_args.append('-UNDEBUG')
     sys.argv.remove("--debug")
 
+# avoid errors if SCIP is build with make:
+extra_compile_args.append('-DNO_CONFIG_HEADER')
+
 extensions = [Extension('pyscipopt.scip', [os.path.join(packagedir, 'scip'+ext)],
                           include_dirs=[includedir],
                           library_dirs=[libdir],
--- a/setup.py
+++ b/setup.py
@@ -6,14 +6,6 @@ scipoptdir = os.environ.get('SCIPOPTDIR', '').strip('"')
 
 extra_compile_args = []
 extra_link_args = []
-# set runtime libraries
-if platform.system() in ['Linux', 'Darwin']:
-    extra_link_args.append('-Wl,-rpath,'+libdir)
-
-# enable debug mode if requested
-if "--debug" in sys.argv:
-    extra_compile_args.append('-UNDEBUG')
-    sys.argv.remove("--debug")
 
 # determine include directory
 if os.path.exists(os.path.join(scipoptdir, 'src')):
@@ -37,6 +29,15 @@ else:
     libname = 'scip'
 
 print('Using SCIP library <%s> at <%s>.' % (libname,libdir))
+
+# set runtime libraries
+if platform.system() in ['Linux', 'Darwin']:
+    extra_link_args.append('-Wl,-rpath,'+libdir)
+
+# enable debug mode if requested
+if "--debug" in sys.argv:
+    extra_compile_args.append('-UNDEBUG')
+    sys.argv.remove("--debug")
 
 cythonize = True
 
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,9 @@
 from setuptools import setup, Extension
 import os, platform, sys, re
 
+runtime_library_dirs = []
+extra_link_args = []
+
 # look for environment variable that specifies path to SCIP
 scipoptdir = os.environ.get('SCIPOPTDIR', '').strip('"')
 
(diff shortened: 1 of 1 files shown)


pfetsch:
Where is ``libscip.so`` installed on the Windows system? Then ``setup.py`` should be adapted accordingly to also search at that place.

setup.py
```diff
@@ -1,11 +1,31 @@
...
 
-# look for environment variable that specifies path to SCIP Opt lib and headers
+# look for environment variable that specifies path to SCIP
 scipoptdir = os.environ.get('SCIPOPTDIR', '').strip('"')
-includedir = os.path.abspath(os.path.join(scipoptdir, 'include'))
+
+# now check whether SCIP is installed on the system or whether a source code directory has been specified
+if os.path.exists(os.path.join(scipoptdir, 'include')):
+    includedir = os.path.abspath(os.path.join(scipoptdir, 'include'))
+else:
```
mattmilten: Users should always first install SCIP. Then there is no need to search for a `src` directory.

setup.py
```diff
@@ -1,11 +1,31 @@
...
+    if os.path.exists(os.path.join(scipoptdir, 'src')):
+        includedir = os.path.abspath(os.path.join(scipoptdir, 'src'))
+    else:
+        print('Neither directory \'src\' nor \'include\' exists in the path <%s>' % scipoptdir)
+        quit(1)
+
+# determine library
 libdir = os.path.abspath(os.path.join(scipoptdir, 'lib'))
 libname = 'scip'
+if (not os.path.exists(os.path.join(libdir, 'libscip.so'))) and (not os.path.exists('/usr/lib/libscip.so')):
```
mattmilten: This is Linux-only and makes the setup routine very convoluted. On macOS the libs have the suffix `dylib` and on Windows there are both `lib` and `dll`.

setup.py
pfetsch: But this is the point: I want to also use by development version that is installed somewhere else.

setup.py
pfetsch: I could turn the test around and only check for scipoptsolver. If it is there, then one can use it. Otherwise one falls back to the default.

Another option would be to use  ``find_library``  from the ``ctypes.util`` package. This should be a portable way to check for a library, e.g., ``find_library('m')`` would find ``libm.so`` on Linux.

setup.py
```diff
@@ -40,6 +60,9 @@
     extra_compile_args.append('-UNDEBUG')
     sys.argv.remove("--debug")
 
+# avoid errors if SCIP is build with make:
+extra_compile_args.append('-DNO_CONFIG_HEADER')
```
mattmilten: If we include this for all builds, we would have wrong information for cmake builds.

setup.py
mattmilten: Also development versions can be installed to some local directory. We always require a `lib` and an `include` directory.

setup.py
mattmilten: I don't want to use `find_library()` because we might then pick a system library over a user specified one.

mattmilten:
What about this: you add a new command line option - similar to `--debug` - to build with make-built SCIP, replacing the libname to `scipsolver` and adding the `-DNO_CONFIG_HEADER`.

setup.py
pfetsch: Yes, but it is easier to just use ``src`` and use SCIP in place.

setup.py
pfetsch: So how is the SCIP version chosen? If ``SCIPOPTDIR`` is set (and it exists), then this should be used, right? I would only use ``find_library`` if the SCIP directory cannot be found.

setup.py
pfetsch: O.k. I can apply this only if we use the makefile version.

setup.py
pfetsch: I have changed to code to just determine whether SCIP has been used "in place". Otherwise nothing changes.

setup.py
pfetsch: This is no only applied if a makefile version has been detected.

setup.py
pfetsch: I have changed the behavior - see the other comments.

pfetsch:
Well, everyone I know does not install SCIP, but uses SCIP "in place", i.e., they download the SCIPOptSuite and compile it (without "make install"), because they develop code with it. The fix is meant for these people (including myself). Note that the changes should also allow for a "cmade" SCIP, which does not have an "include" directory by default.

pfetsch:
Are we ready to merge? Who should do this?

PR #304, opened 2019-08-27
overload operators for solutions

mueldgog:
This change allows us to create, print, and access solutions more nicely:
```
x = m.addVar(lb=0,ub=4)
s = m.createSol()
s[x] = 2.0
print(s)
print(s[x])
```

### DIFF
--- a/src/pyscipopt/scip.pyx
+++ b/src/pyscipopt/scip.pyx
@@ -498,7 +498,7 @@ cdef class Solution:
         for i in range(SCIPgetNVars(self.scip)):
             scip_var = SCIPgetVars(self.scip)[i]
 
-            # exctract name
+            # extract name
             print(scip_var == NULL)
             cname = bytes(SCIPvarGetName(scip_var))
             name = cname.decode('utf-8')
--- a/src/pyscipopt/scip.pyx
+++ b/src/pyscipopt/scip.pyx
@@ -499,7 +499,6 @@ cdef class Solution:
             scip_var = SCIPgetVars(self.scip)[i]
 
             # extract name
-            print(scip_var == NULL)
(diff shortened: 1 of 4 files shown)


mueldgog:
The change in `__repr__` is a bit more lengthy because I didn't want to use `m.getVars()` because this would create a lot of unneeded python objects.

mueldgog:
We could also drop `getSolVal()` and `setSolVal()` if we want. They are kind of redundant now.

src/pyscipopt/scip.pyx
```diff
@@ -474,15 +474,38 @@ cdef class NLRow:
...
+        PY_SCIP_CALL(SCIPsetSolVal(self.scip, self.sol, var.scip_var, value))
+
+    def __repr__(self):
+        cdef SCIP_VAR* scip_var
+
+        vals = {}
+        for i in range(SCIPgetNVars(self.scip)):
+            scip_var = SCIPgetVars(self.scip)[i]
+
+            # exctract name
```
```suggestion
            # extract name
```

src/pyscipopt/scip.pyx
```diff
@@ -474,15 +474,38 @@ cdef class NLRow:
...
+            print(scip_var == NULL)
```
mattmilten: Do we need this `print`?

mattmilten:
Cool! That makes working with the solution object much more convenient.

src/pyscipopt/scip.pyx
mueldgog: Oh, sorry. I was just debugging a bit and accidentally added it.

mueldgog: Fixed with ee2df06.

PR #305, opened 2019-08-28
remove releaseVar() from scip.pyx

mueldgog:
Since we already release the variable during `addVar()` it is not valid to use this function to release a variable.


mattmilten:
Good. I will increase the version number to signal this change.

PR #308, opened 2019-08-29
Extended getVal method in Model class to accept Expr instances.

mcitoler:
Now you can evaluate a polynomial expression after optimizing a model. This covers part of https://github.com/SCIP-Interfaces/PySCIPOpt/issues/280. We are till missing evaluation of GenExpr.

### DIFF
--- a/src/pyscipopt/expr.pxi
+++ b/src/pyscipopt/expr.pxi
@@ -109,6 +109,13 @@ class Term:
     def __repr__(self):
         return 'Term(%s)' % ', '.join([str(v) for v in self.vartuple])
 
+    def _evaluate(self, solution):
+        '''computes the value of the term in the given solution.'''
+        prod = 1
+        for var in self.vartuple:
+            prod *= solution[var]
+        return prod
+        
 CONST = Term()
 
 # helper function
@@ -298,6 +305,10 @@ cdef class Expr:
             return 0
         else:
             return max(len(v) for v in self.terms)
+
+    def _evaluate(self, solution):
+        '''computes the value of the expression in the given solution.'''
+        return sum(term._evaluate(solution)*coeff for term, coeff in self.terms.items() if coeff != 0)
 
 
 cdef class ExprCons:
--- a/src/pyscipopt/expr.pxi
+++ b/src/pyscipopt/expr.pxi
@@ -109,11 +109,11 @@ class Term:
     def __repr__(self):
         return 'Term(%s)' % ', '.join([str(v) for v in self.vartuple])
 
-    def _evaluate(self, solution):
-        '''computes the value of the term in the given solution.'''
+    def _evaluate(self, value):
+        '''computes the value of the term in the given variable-value mapping.'''
         prod = 1
         for var in self.vartuple:
-            prod *= solution[var]
+            prod *= value[var]
         return prod
         
 CONST = Term()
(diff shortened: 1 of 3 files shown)


src/pyscipopt/expr.pxi
```diff
@@ -109,6 +109,13 @@ class Term:
     def __repr__(self):
         return 'Term(%s)' % ', '.join([str(v) for v in self.vartuple])
 
+    def _evaluate(self, solution):
```
mattmilten: Can you rename `solution` to `value`?

This also applies to later occurrences in this file. In my opinion there is no notion of a `solution` within the context of `Term` and `Expr`.

src/pyscipopt/scip.pyx
```diff
@@ -3562,28 +3562,32 @@ cdef class Model:
             raise Warning("method cannot be called before problem is solved")
         return self.getSolObjVal(self._bestSol, original)
 
-    def getSolVal(self, Solution sol, Variable var):
+    def getSolVal(self, Solution sol, Expr expr):
         """Retrieve value of given variable in the given solution or in
```
```suggestion
        """Retrieve value of given variable or expression in the given solution or in
```

src/pyscipopt/scip.pyx
```diff
@@ -3562,28 +3562,32 @@ cdef class Model:
...
-        return SCIPgetSolVal(self._scip, sol.sol, var.scip_var)
+        if isinstance(expr, Variable):
+            var = <Variable> expr
+            return SCIPgetSolVal(self._scip, sol.sol, var.scip_var)
+        else:
+            return expr._evaluate(sol)
 
-    def getVal(self, Variable var):
+    def getVal(self, Expr expr):
         """Retrieve the value of the best known solution.
```
```suggestion
        """Retrieve the value of the given variable or expression in the best known solution.
```

src/pyscipopt/expr.pxi
fserra: I was gonna suggest to change `value` to `solution` :)

What about using `point`? I guess one evaluates at a `point` to get a `value`

src/pyscipopt/scip.pyx
```diff
@@ -3562,28 +3562,32 @@ cdef class Model:
...
 
-    def getSolVal(self, Solution sol, Variable var):
-        """Retrieve value of given variable in the given solution or in
+    def getSolVal(self, Solution sol, Expr expr):
+        """Retrieve value of given variable or expression in the given solution or in
         the LP/pseudo solution if sol == None
 
         :param Solution sol: solution
-        :param Variable var: variable to query the value of
+        :param Expr expr: polynomial expression to query the value of
```
fserra: I would probably be repetitive here and add something like `Note: a variable is also an expression`

src/pyscipopt/expr.pxi
mattmilten: Yes, `point` is better.

fserra:
I have a suggestion. Benny changed the way we ask for solution values of variables: instead of `m.getSolVal(var, sol)` one can use `sol[var]`.

Why not extend this to expressions? It would be really cool to have `sol[var^2]` or stuff like that.

mattmilten:
Yeah, I also thought about this. I still would like to keep the "old-fashioned" methods as well.

mattmilten:
Shall we merge this first and then implement the "shortcut" `sol[x^3]`?

fserra:
done, thanks @mcitoler

PR #309, opened 2019-09-03
Modifications on customized benders decomposition

andynuaa:
Many thanks to @stephenjmaher 's devotion, the latest project contains necessary functions for customized benders decomposition and cuts.


src/pyscipopt/scip.pyx
```diff
@@ -2757,6 +2759,27 @@ cdef class Model:
...
+        benders -- the Benders' decomposition object for the that the subproblem belongs to (Default = None)
+        """
+        cdef SCIP_BENDERS* scip_benders
+        cdef SCIP* scip_subprob
+
+        if benders is None:
+            scip_benders = SCIPfindBenders(self._scip, "default")
+        else:
+            n = str_conversion(benders.name)
+            scip_benders = SCIPfindBenders(self._scip, n)
```
fserra: I guess this question is for @stephenjmaher 

Why does `benders` doesn't know itself and always needs to find itself? instead of storing the name one could just store the C object, right?

stephenjmaher: Is it possible to store the C object? I didn't see where we do this in the other plugins. Would it be something like
```
cdef SCIP_BENDERS* _benders
```
in benders.pxi?

fserra: I guess... and then set it in the include? Oh yeah, but you are right, we are not doing this on the other plugins... maybe we should. Probably would be good here since SCIPfindBenders is called for everything

src/pyscipopt/scip.pyx
```diff
@@ -3091,7 +3170,7 @@ cdef class Model:
         cdef SCIP_BENDERSCUT* scip_benderscut
         scip_benderscut = SCIPfindBenderscut(_benders, n)
         benderscut.model = <Model>weakref.proxy(self)
-        benderscut.benders = <Benders>weakref.proxy(benders)
+        benderscut.benders = benders
```
fserra: why this change? was `benders` getting freed before it should?

stephenjmaher: I can't remember. I would have to retest

src/pyscipopt/scip.pyx
stephenjmaher: I just tested this, it seems to work. I will make the changes now.

src/pyscipopt/scip.pyx
fserra: I think that we use weakrefs in model because otherwise we create a circular dependency and nothing gets freed (not sure though). Does this not happen here?

fserra:
@stephenjmaher @andynuaa is this ready?

src/pyscipopt/scip.pyx
stephenjmaher: I just tested this and the error that i get is
```
TypeError: Argument 'benders' has incorrect type (expected pyscipopt.scip.Benders, got weakproxy)
```
It seems that it doesn't work to use the weakref in this settiing.

fserra: that sounds weird. how can I reproduce that?

stephenjmaher:
You can reproduce this with the `testcustomizedbenders.py` test. I think the error is triggered because we need to access the Benders class from within the Benderscut class.
________________________________

fserra:
ok thanks. I thought the problem was that class Benders was not weak referentiable, but it wasn't. I don't know what the problem is, but I guess is ok.

So the last thing. Can you rename the test to test_custo.... please, it seems the test is not running in travis.

Also, some tests don't appear in travis (like test_nodesel). I guess this is because it branched from an old commit. Can you merge the current master into this branch, or must @andynuaa do it?
I guess we should improve travis so that it merges first and then run the tests. @mattmilten do you know how to do this?

stephenjmaher:
@fserra do you have any idea why the jobs are failing? I can't reproduce this on my local machine.

fserra:
I have no idea. I just fixed two other tests that were failing in debug mode on master. You might want to enable some output on the test and then modify the travis script so that it calls pytest with `-s` so that it shows the output.

fserra:
nice. What was the problem?

stephenjmaher:
It was in `checkBendersSubproblemOptimality`. I wasn't catching the case when `solution = None`. I just needed to set a variable `scip_sol` to `NULL`. It took a bit to track this down. :-)

PR #316, opened 2019-11-12
mac on travis; first attempt

fserra:


.travis.yml
```diff
@@ -26,8 +28,13 @@ notifications:
...
 before_install:
-  - wget http://scip.zib.de/download/release/SCIPOptSuite-$VERSION-Linux-Ub1604.deb
-  - sudo apt-get update && sudo apt install -y ./SCIPOptSuite-$VERSION-Linux-Ub1604.deb
+  - if [ "$TRAVIS_OS_NAME" = "linux" ]; then wget wget http://scip.zib.de/download/release/SCIPOptSuite-$VERSION-Linux-Ub1604.deb; fi
+  - if [ "$TRAVIS_OS_NAME" = "linux" ]; sudo apt-get update && sudo apt install -y ./SCIPOptSuite-$VERSION-Linux-Ub1604.deb; fi
+  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then wget http://scip.zib.de/download/release/scipoptsuite-$VERSION.tgz; fi
+  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then tar -xf scipoptsuite-$VERSION.tgz && cd scipoptsuite-$VERSION; fi
+  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then mkdir build && cd build && cmake -D CMAKE_BUILD_TYPE=Release -D ZIMPL=OFF -D GCG=OFF -D BUILD_TESTING=OFF -D CMAKE_INSTALL_PREFIX=../install ..; fi
+  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then make -j 4 && make install; fi
+  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then cd .. && export SCIPOPTDIR=$(pwd)/install && cd ..; fi
```
mattmilten: Why don't you use the pre-built package for macOS? I would suggest the one without Ipopt. In the end, this is supposed to test PySCIPOpt with SCIP and compiling SCIP takes a long time, as you know.

fserra: yes I know. This was just the first iteration, just a copy paste from the julia's interface travis to see if it just worked. Also, they have a problem when using the prebuilt package, so I didn't wanna deal with that now.

fserra:
ok so it seems that `pip3 install pyscipopt` works with scipoptsuite-darwin-no-ipopt. But for some reason py.test fails.

also, it seems that something is wrong (I guess) because one needs to do `export DYLD_LIBRARY_PATH=$SCIPOPTDIR/lib` or otherwise the library is not found. But this is a problem on the scip side (I guess)

CGraczyk:
Since Travis wont be used anymore I will close this PR.

PR #318, opened 2019-11-21
Check whether PY_SCIP_CALL is defined in __dealloc__. 

yozw:
PY_SCIP_CALL may have already been garbage collected by the time this method is called, causing an exception.


mattmilten:
Looks good to me.

For future commits: Please keep your header/first line in commit messages short - usually 50 or 70 characters are upper limits.

PR #320, opened 2019-11-26
Adds three trivial methods and a solution status case.

yozw:


mattmilten:
Thank you for your contribution!

fserra:
@mattmilten I am a bit confused. Here @yozw  added the function getNLPIterations and we had lpiGetNLPIterations. Is there any difference between them? Of course, in SCIP one uses lpiGetNLPIterations to get the iterations of an arbitrary lpi, but in our python implementation, we get SCIP's lpi. So, should we remove this function from the class Model?

mattmilten:
There actually is a diference: our old method returns the current LP's iterations while the new one returns the total number of iterations so far.

PR #321, opened 2019-11-27
Move main Python class definitions into scip.pxd.

yozw:
Doing this enables writing custom extensions (in the form of .pyx files) to PySCIPOpt.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - remove Python 2 support
 - add getOpenNodes() to access all open nodes
 - add support for node selector plugin
+- move the main Python class declarations into scip.pxd; this makes it possible to write custom `pyx` extensions.
 
 ## 2.2.1 - 2019-09-17
 ### Added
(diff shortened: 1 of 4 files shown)


mattmilten:
Looks good to me. @fserra, what do you think?

yozw:
Friendly ping to @fserra. We are currently blocked on this particular pull request being accepted: I have two .pyx libraries that depend on this and I'm postponing further development until I'm sure that writing custom extensions is supported. (That's by no means an attempt to pressure you into accepting this; just getting a decision would be great.)

fserra:
hey guys
i am sorry, i am currently on vacation without much time..i had a quick look and looks good. i guess it would make sense to add an entry in the changelog and maybe something on the readme
@mattmilten i leave the rest to you

mattmilten:
Very good point! @yozw, can you please add an entry to the CHANGELOG shortly explaining the change?

yozw:
@fserra: Sorry! I didn't realise you were on vacation!

@mattmilten: Yes, I will add an entry. Do you mind if I do that in a separate commit, or do you want this to be a single-commit pull request?

mattmilten:
Please just add a new commit - they're free :-)
We don't have rules about how many commits can be in a pull request and I don't see why we should.

Thanks!

yozw:
I've added an entry to the changelog.

PR #322, opened 2019-12-04
Expose more SCIP functionality in scip.pxd.

yozw:


yozw:
I promise this is the last one for now :) Thanks so much for all the review and the help.

fserra:
hey @yozw 

why do you want to expose `SCIPlpRecalculateObjSqrNorm`? This is an internal function of SCIP, not available in the public api. Also, the struct of SCIP is not supposed to be accessed directly.

yozw:
HI @fserra -- Thanks for pointing this out. I just now realised that I should have use only functions from "pub_lp.h". I agree with this entirely.

I want to calculate the square norms of the objective vector. Similarly, I want to calculate the dot product between the rows and the objective vector. The information is there in the structs, but I imagine you may object (rightfully) to exposing SCIP_ROW.objprod.

I'm fairly sure I can work out how to do what I want without calling internal functionionality by calculating the above myself, and this is what I'll try out first. It will be slower however (which may not be a problem). If you have any alternative ideas, that'd be great.

mueldgog:
Hi, probably the cleanest way would be to extend the API of SCIP. We could then release the missing methods with the next major release of SCIP (in January).

- You can access the square of `row->sqrnorm` by using `SCIProwGetNorm()`
- You can access the name of a branching rule with `SCIPbranchruleGetName()`.
- Would it be enough to add an API function that wraps `SCIProwGetObjParallelism`? It computes `REALABS(row->objprod) / SQRT(row->sqrnorm * lp->objsqrnorm)`.
- I discuss with the other developers whether it makes sense to have an API function for the age of a column.

Anything else that you need to access?

yozw:
Thank you! I've updated our own code to use `SCIProwGetNorm()`, `SCIPbranchruleGetName()`, and `SCIPgetObjNorm()` instead, and simplified the code in this pull request accordingly.

I believe that indeed all I need from the SCIP API would then be:

* Access to the return value of `SCIProwGetObjParallelism()`.
* A new function, `SCIPcolGetAge()`.

These seem fairly trivial to add, so I'm happy to do that myself in a pull request for SCIP. That does mean that PySCIPOpt will depend on a later version of SCIP.

mueldgog:
I am going to add these two methods to SCIP because the SCIP repository is not public.

> That does mean that PySCIPOpt will depend on a later version of SCIP.

PySCIPOpt is only compatible with the latest release version of SCIP. So I would suggest the following:

1. I add the two missing API methods to SCIP.
2. After the next release of SCIP, we can extend scip.pxd accordingly.

Btw, why didn't you add your changes to scip.pyx to this pull request? I think that you also extended scip.pyx to use the methods that you have added to scip.pxd, right?

yozw:
Ok, sounds good.

To answer your question: I moved several definitions from `scip.pyx` to `scip.pxd` (in https://github.com/SCIP-Interfaces/PySCIPOpt/pull/321), making it possible to write custom `pyx` extensions. We have our own `pyx` extensions that use the definitions I'm adding here. Those extensions aren't particularly interesting for general consumption, which is why I'm not adding them back to PySCIPOpt.

mueldgog:
> To answer your question: I moved several definitions from scip.pyx to scip.pxd (in #321), making it possible to write custom pyx extensions. We have our own pyx extensions that use the definitions I'm adding here. Those extensions aren't particularly interesting for general consumption, which is why I'm not adding them back to PySCIPOpt.

Okay, I understand. Can you revert all the changes of adding the members to the structures in scip.pxd? I think that we can then merge your pull request.

yozw:
Done

src/pyscipopt/scip.pxd
```diff
@@ -348,7 +351,8 @@ cdef extern from "scip/scip.h":
         pass
 
     ctypedef struct SCIP_BRANCHRULE:
-        pass
+        SCIP_RETCODE (*branchexeclp)(SCIP* scip, SCIP_BRANCHRULE* branchrule, SCIP_Bool allowaddcons, SCIP_RESULT* result)
+        const char* SCIPbranchruleGetName(SCIP_BRANCHRULE* branchrule)
```
mueldgog: This should appear after SCIPbranchruleGetData below.

src/pyscipopt/scip.pxd
```diff
@@ -284,6 +284,9 @@ cdef extern from "scip/scip.h":
     ctypedef struct SCIP_NLROW:
         pass
 
+    ctypedef struct SCIP_NLROW:
```
mueldgog: You redefined `struct SCIP_NLROW`. Please remove this.

src/pyscipopt/scip.pxd
```diff
@@ -348,7 +358,8 @@ cdef extern from "scip/scip.h":
         pass
 
     ctypedef struct SCIP_BRANCHRULE:
-        pass
+        char* name
+        SCIP_RETCODE (*branchexeclp)(SCIP* scip, SCIP_BRANCHRULE* branchrule, SCIP_Bool allowaddcons, SCIP_RESULT* result)
```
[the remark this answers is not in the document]
mueldgog: Please remove the callback.

src/pyscipopt/scip.pxd
```diff
@@ -1557,10 +1583,23 @@ cdef extern from "scip/pub_lp.h":
...
 
 cdef extern from "scip/scip_tree.h":
     SCIP_RETCODE SCIPgetOpenNodesData(SCIP* scip, SCIP_NODE*** leaves, SCIP_NODE*** children, SCIP_NODE*** siblings, int* nleaves, int* nchildren, int* nsiblings)
 
+cdef extern from "scip/struct_branch.h":
+    cdef struct SCIP_Branchrule:
+        pass
+
+cdef extern from "scip/def.h":
+    SCIP_Real REALABS(SCIP_Real x)
```
mueldgog: Do you need this? If not then please remove it.

src/pyscipopt/scip.pxd
```diff
@@ -1557,10 +1583,23 @@ cdef extern from "scip/pub_lp.h":
...
+    int SCIPcolGetNNonz(SCIP_COL* col)
+    SCIP_ROW** SCIPcolGetRows(SCIP_COL* col)
+    SCIP_Real* SCIPcolGetVals(SCIP_COL* col)
+    int SCIPcolGetIndex(SCIP_COL* col)
+    SCIP_Real SCIPcolGetObj(SCIP_COL *col)
 
 cdef extern from "scip/scip_tree.h":
     SCIP_RETCODE SCIPgetOpenNodesData(SCIP* scip, SCIP_NODE*** leaves, SCIP_NODE*** children, SCIP_NODE*** siblings, int* nleaves, int* nchildren, int* nsiblings)
 
+cdef extern from "scip/struct_branch.h":
```
mueldgog: I don't think that this is needed. Please remove it.

src/pyscipopt/scip.pxd
yozw: I can do without (by using Python's `abs`); I'll remove it.

src/pyscipopt/scip.pxd
yozw: Oops. Thanks.

src/pyscipopt/scip.pxd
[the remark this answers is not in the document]
yozw: I think we do actually need this. We have branch rules written in Python that internally call `branchexeclp` for existing branch rules (e.g. `relpscost`).

src/pyscipopt/scip.pxd
yozw: Yes, indeed.

src/pyscipopt/scip.pxd
yozw: This was here because without it, I get a compiler error: `incomplete definition of type 'struct SCIP_Branchrule'` (caused by the callback definition in line 354). Looking into this a bit further, it turns out that it is sufficient to just include the file `scip/struct_branch.h`, without importing anything from it.

This seems a bit odd to me. I wonder: is this actually a problem in `type_branch.h` in SCIP?

src/pyscipopt/scip.pxd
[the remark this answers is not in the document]
mueldgog: Hmm, I understand but I feel very mixed about adding this to PySCIPOpt. At least for `relpscost` you don't need this because you can use `SCIPexecRelpscostBranching` from `branch_relpscost.h`. Would this be enough for your application? If yes, then please remove the callback from the struct. Thanks!

@fserra and @mattmilten, what do you generally think about adding the callback?

src/pyscipopt/scip.pxd
mueldgog: No, I don't think so. Clearly, `type_branch.h` is needed if you include `pub_branch.h`. This allows you to access the API methods related to, e.g., branching rules, but you want to ensure that the user does not have access to the struct of a branching rule.

This is also my problem with this change because PySCIPOpt should not give the user access to some of the structs.

fserra:
as a matter of principle we don't want to allow anything that cannot be
done through SCIP C api, because this is completely untested, new types of
bugs can appear, etc. So, if it is not intended that users have access to
the callback, then we shouldn't allow it here.

mueldgog:
I talked to @gerald-gamrath  and one alternative would be to extend the API of SCIP in a way that it is possible to directly call the EXECLP callback of a branching rule. This would allow users to write a custom branching rule in SCIP/PySCIPOpt that can dynamically switch between different branching rules.

As mentioned above, if you only need access to `relpscost` and `fullstrong` then please have a look at their API methods.

yozw:
> I talked to @gerald-gamrath and one alternative would be to extend the API of SCIP in a way that it is possible to directly call the EXECLP callback of a branching rule. This would allow users to write a custom branching rule in SCIP/PySCIPOpt that can dynamically switch between different branching rules.

I've removed the callback declaration. I agree that it shouldn't be exposed and I'm happy with these three lines as local changes in our internal code base. This is not too painful to maintain. I'll also have a look at using the branching rules directly, as you suggest.

Thanks for your patience, and for all the insightful comments.

yozw:
> PySCIPOpt is only compatible with the latest release version of SCIP. So I would suggest the following:
> 
> 1. I add the two missing API methods to SCIP.
> 2. After the next release of SCIP, we can extend scip.pxd accordingly.

Just to clarify: will this happen, or do you want me to file feature requests for this?

mueldgog:
@yozw, I have added
- `SCIPgetRowObjParallelism()` and
- `SCIPcolGetAge()`
to SCIP. You can access both methods after the next release of SCIP. Here is a patch if you want to access the methods right now:

```
diff --git a/CHANGELOG b/CHANGELOG
index 3459e6bd37..568939b304 100644
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -259,6 +259,8 @@ Interface changes
 - new API functions SCIPcleanupConss{Linear,Varbound,Setppc,Logicor,Knapsack}() to clean up inactive variables from those types of linear constraints
 - new API function SCIPsetBendersSubproblemComp() used to add a custom comparison method for ordering the Benders'
   decomposition subproblem solves. The comparison method is used to help with load balancing.
+- new API function SCIPgetRowObjParallelism to get the objective parallelism of a row
+- new API function SCIPcolGetAge to get the age of a column
 
 ### Command line interface
 
diff --git a/src/scip/lp.c b/src/scip/lp.c
index 2a9e70ba65..ced5696e7d 100644
--- a/src/scip/lp.c
+++ b/src/scip/lp.c
@@ -16733,6 +16733,7 @@ SCIP_RETCODE SCIPlpWriteMip(
 #undef SCIPcolGetVals
 #undef SCIPcolGetStrongbranchNode
 #undef SCIPcolGetNStrongbranchs
+#undef SCIPcolGetAge
 #undef SCIPboundtypeOpposite
 #undef SCIProwGetNNonz
 #undef SCIProwGetNLPNonz
@@ -17018,6 +17019,16 @@ int SCIPcolGetNStrongbranchs(
    return col->nsbcalls;
 }
 
+/** gets the age of a column, i.e., the total number of successive times a column was in the LP and was 0.0 in the solution */
+int SCIPcolGetAge(
+   SCIP_COL*             col                 /**< LP column */
+   )
+{
+   assert(col != NULL);
+
+   return col->age;
+}
+
 /** gets opposite bound type of given bound type */
 SCIP_BOUNDTYPE SCIPboundtypeOpposite(
    SCIP_BOUNDTYPE        boundtype           /**< type of bound (lower or upper) */
diff --git a/src/scip/pub_lp.h b/src/scip/pub_lp.h
index 1bcbc49100..bc5b9c0227 100644
--- a/src/scip/pub_lp.h
+++ b/src/scip/pub_lp.h
@@ -188,6 +188,12 @@ int SCIPcolGetNStrongbranchs(
    SCIP_COL*             col                 /**< LP column */
    );
 
+/** gets the age of a column, i.e., the total number of successive times a column was in the LP and was 0.0 in the solution */
+SCIP_EXPORT
+int SCIPcolGetAge(
+   SCIP_COL*             col                 /**< LP column */
+   );
+
 /** gets opposite bound type of given bound type */
 SCIP_EXPORT
 SCIP_BOUNDTYPE SCIPboundtypeOpposite(
@@ -221,6 +227,7 @@ SCIP_BOUNDTYPE SCIPboundtypeOpposite(
 #define SCIPcolGetVals(col)             (col)->vals
 #define SCIPcolGetStrongbranchNode(col) (col)->sbnode
 #define SCIPcolGetNStrongbranchs(col)   (col)->nsbcalls
+#define SCIPcolGetAge(col)              (col)->age
 #define SCIPboundtypeOpposite(boundtype) \
    ((boundtype) == SCIP_BOUNDTYPE_LOWER ? SCIP_BOUNDTYPE_UPPER : SCIP_BOUNDTYPE_LOWER)
 
diff --git a/src/scip/scip_lp.c b/src/scip/scip_lp.c
index 4f7fbf787d..fa468644b7 100644
--- a/src/scip/scip_lp.c
+++ b/src/scip/scip_lp.c
@@ -2120,6 +2120,25 @@ SCIP_Real SCIPgetRowSolFeasibility(
       return SCIProwGetPseudoFeasibility(row, scip->set, scip->stat);
 }
 
+/** returns the parallelism of row with objective function
+ *
+ *  @return 1 is returned if the row is parallel to the objective function and 0 if it is orthogonal
+ *
+ *  @pre this method can be called in one of the following stages of the SCIP solving process:
+ *       - \ref SCIP_STAGE_SOLVING
+ */
+SCIP_Real SCIPgetRowObjParallelism(
+   SCIP*                 scip,               /**< SCIP data structure */
+   SCIP_ROW*             row                 /**< LP row */
+   )
+{
+   assert(row != NULL);
+
+   SCIP_CALL_ABORT( SCIPcheckStage(scip, "SCIPgetRowObjParallelism", FALSE, FALSE, FALSE, FALSE, FALSE, FALSE, FALSE, FALSE, FALSE, TRUE, FALSE, FALSE, FALSE, FALSE) );
+
+   return SCIProwGetObjParallelism(row, scip->set, scip->lp);
+}
+
 /** output row to file stream via the message handler system
  *
  *  @return \ref SCIP_OKAY is returned if everything worked. Otherwise a suitable error code is passed. See \ref
diff --git a/src/scip/scip_lp.h b/src/scip/scip_lp.h
index a515b2496e..e55ff42800 100644
--- a/src/scip/scip_lp.h
+++ b/src/scip/scip_lp.h
@@ -1452,6 +1452,18 @@ SCIP_Real SCIPgetRowSolFeasibility(
    SCIP_SOL*             sol                 /**< primal CIP solution */
    );
 
+/** returns the parallelism of row with objective function
+ *
+ *  @return 1 is returned if the row is parallel to the objective function and 0 if it is orthogonal
+ *
+ *  @pre this method can be called in one of the following stages of the SCIP solving process:
+ *       - \ref SCIP_STAGE_SOLVING
+ */
+SCIP_Real SCIPgetRowObjParallelism(
+   SCIP*                 scip,               /**< SCIP data structure */
+   SCIP_ROW*             row                 /**< LP row */
+   );
+
 /** output row to file stream via the message handler system
  *
  *  @return \ref SCIP_OKAY is returned if everything worked. Otherwise a suitable error code is passed. See \ref
```

yozw:
Wow, impressive efficiency! Thanks so much for already taking care of this.

PR #324, opened 2019-12-17
Add all SCIP status codes

juchengquan:
- All SCIP status codes are added.


mueldgog:
Thanks for the pull request!

PR #331, opened 2020-01-16
Add scip.Model export to / import from Python pointer

### LINKED ISSUE #330: Model from/to pointer
Hello,

We're developing a [library](github.com/ds4dm/ecole) with Python bindings using Scip in the background and we would like our user to be able to leverage PyScipOpt for their own customization.

The easiest for us would be to exchange a `SCIP*` through Python, using [`ctypes.c_void_p`](https://docs.python.org/3.8/library/ctypes.html#ctypes.c_void_p).
For PyScipOpt, I would see this as adding two methods:

```python
class Model:
    ...

    @classmethod
    def from_ptr(cls, ptr: ctypes.c_void_p, take_ownership: bool) -> cls: ...

    def to_ptr(self, give_ownership:bool) -> ctypes.c_void_p: ...
```
Where ownership is about de-allocating or not.
I'd happily send a PR if you're interested 🙂

AntoinePrv:
Here's a working solution to resolve #330 
Turns out `uintptr_t` was the key to proper casting, while `ctypes.c_void_p` is just a wrapper around an `int`.

- [x] Add export/import functions
- [x] Document functions
- [x] Add test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - remove Python 2 support
 - add Model.getParams that returns a dict mapping all parameter names to their values
 - add Model.setParams to set multiple parameters at once using a dict
+- Add Model.from_ptr and Model.to_ptr to interface with SCIP* managed outside of PySCIPOpt
 
 ## 2.2.3 - 2019-12-10
 ### Added
--- a/src/pyscipopt/scip.pyx
+++ b/src/pyscipopt/scip.pyx
@@ -5,10 +5,12 @@ from os.path import abspath
 from os.path import splitext
 import sys
 import warnings
+import ctypes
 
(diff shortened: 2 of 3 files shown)


AntoinePrv:
Do not know why Cython doesn't allow keyword here. I tried removing all typing from the declarations, it does not change anything.
Any idea?

mattmilten:
Hmm, strange. No idea how to fix it.

AntoinePrv:
So this is seems to be a Cython optimization for 1 argument functions.
I added a decorator to allow keywords, see `always_allow_keywords` from
[Cython doc](https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html).

fserra:
can you add an entry to the changelog, please? Otherwise, I guess we are good to merge.

PR #333, opened 2020-01-23
Node.getParent() should return None if SCIPnodeGetParent() returns NULL.

yozw:


src/pyscipopt/scip.pyx
```diff
@@ -502,7 +502,8 @@ cdef class Node:
 
     def getParent(self):
         """Retrieve parent node."""
```
fserra: can you modify the documentation to say that it returns none if the node doesn't have a parent and add a changelog entry please! thanks!

yozw:
Sure, done.

fserra:
thanks!

PR #335, opened 2020-01-24
Adds NULL pointer checks to several Python wrapper classes.

yozw:


yozw:
@fserra: which of the test failures are legitimate usages of NULL pointers, and which ones are bugs in PySCIPopt (if any)?

yozw:
I also think, by the way, that a "Warning" exception is not the appropriate exception to raise. Warnings usually mean that the code can continue running, but something seems not quite right. "Warning" is also probably too general an exception? I am tempted to change it to "ValueError", but I appreciate that that changes the PySCIPOpt interface.

fserra:
sol can be NULL because when sol is NULL we just mean the LP solution. This should fix tests: benders, 

Node can also be NULL. It just means that there is no Node. In test_branch_probing_lp it says:
```
            down, eq, up = self.model.branchVarVal(self.cont, 1.3)
```
The function `branchVarVal` can return up to 3 children but sometimes returns two. In that case, one is going to be NULL. I don't know, though, if instead of creating a Node with a NULL pointer inside it would be better to just return a None.

test_conshdlr is an infeasible problem, but `optimize` always calls:
```
        PY_SCIP_CALL(SCIPsolve(self._scip))
        self._bestSol = Solution.create(self._scip, SCIPgetBestSol(self._scip))
```
So we could argue that this is a bug because we are not really checking that the solve produced a solution. On the other hand, when there is no solution `SCIPgetBestSol` just returns NULL, so maybe again `Solution.create(scip, NULL)` should return None.

yozw:
> sol can be NULL because when sol is NULL we just mean the LP solution. This should fix tests: benders,

Can a user do anything with a Solution(scip, NULL) object? Wouldn't __getitem__, __setitem__, and __repr__ just segfault? In that case I agree with you (below in your message) that PySCIPOpt should return None instead of Solution(scip, NULL).

> Node can also be NULL. It just means that there is no Node. In test_branch_probing_lp it says:
> 
> ```
>             down, eq, up = self.model.branchVarVal(self.cont, 1.3)
> ```
> 
> The function `branchVarVal` can return up to 3 children but sometimes returns two. In that case, one is going to be NULL. I don't know, though, if instead of creating a Node with a NULL pointer inside it would be better to just return a None.

The reason I started doing this in the first place is that Node.getParent() returned Node(NULL), and I realised that that object is really invalid: calling any method on Node(NULL) caused a Segfault. So, again, I would argue that returning None is better for the user.

> 
> test_conshdlr is an infeasible problem, but `optimize` always calls:
> 
> ```
>         PY_SCIP_CALL(SCIPsolve(self._scip))
>         self._bestSol = Solution.create(self._scip, SCIPgetBestSol(self._scip))
> ```
> 
> So we could argue that this is a bug because we are not really checking that the solve produced a solution. On the other hand, when there is no solution `SCIPgetBestSol` just returns NULL, so maybe again `Solution.create(scip, NULL)` should return None.

fserra:
yes, I also like None

yozw:
@fserra: I've actually reverted to check for SCIP_SOL* == NULL, because there is a legitimate reason that it can be NULL. (Although the distinction may not be clear to the users.)

The branching methods now return middle child = None when there is no middle child.

fserra:
can't the create method of the Node return None? I find it too cumbersome to start checking in the functions if SCIP returned NULL or not.

yozw:
I preferred this solution because it is explicit about the few cases where the user can expect None. Moreover, bugs due to mistakenly passing in a NULL pointer are caught early on. But I'll go with whatever you prefer.

fserra:
The thing is that we want to keep the interface a thin a layer as possible. I guess there are more functions that can just return NULL and so we would have to be doing the checks all over the place. For example, `getParent` shouldn't really check if parent is NULL.

I guess we can tell the user when to expect NULL through the documentation and not through the code, just like the documentation of `getParent` is doing.

To catch bugs about passing NULL around I guess the best is just to link pyscipopt to a SCIP compiled in debug mode as an assert would complain. This, however, would require the user to compile pyscipopt... I don't know if we provide a debug mode pyscipopt in pip, @mattmilten ?

mattmilten:
Hmm, I guess we could build a debug PySCIPOpt and also put it on PyPI. But I don't really see the added benefit as there are no debug libraries of SCIP itself, so the user would need to compile the SCIPOptSuite in debug first. IMO, linking then PySCIPOpt to this debug library is the less complicated part.

yozw:
I've implemented @fserra's suggestion: Node.create(NULL) now returns None, and I've simplified Node.getParent().

fserra:
cool, thanks!

PR #337, opened 2020-01-24
fix some memory leaks

fserra:
the mallocs don't make sense. all the functions return SCIP own data


yozw:
Nice :) I just noticed this too.

PR #338, opened 2020-01-25
Expose domain changes, branching decisions, added constraints for nodes.

yozw:


src/pyscipopt/scip.pyx
```diff
@@ -535,6 +624,19 @@ cdef class Node:
         """Retrieve the estimated value of the best feasible solution in subtree of the node"""
         return SCIPnodeGetEstimate(self.scip_node)
 
+    def getAddedConss(self):
```
yozw: Do you know why getNAddedConss() always returns zero when calling self.getModel().getNAddedConss() inside a branch rule? Consequently, I empirically always see getAddedConss() return an empty list. This happens even when I see the total number of constraints of the model change.

src/pyscipopt/scip.pxd
```diff
@@ -1637,6 +1670,30 @@ cdef class Solution:
...
+cdef class Branching:
+    """Branching decision."""
+    cdef SCIP_VAR* scip_var
+    cdef SCIP_Real scip_bound
+    cdef SCIP_BOUNDTYPE scip_boundtype
+
+    @staticmethod
+    cdef create(SCIP_VAR* scip_var, SCIP_Real scip_bound,
+                SCIP_BOUNDTYPE scip_boundtype)
+
```
fserra: So here I am not sure if we really want to have this in the interface. This is a convenience class to store the triples (var, bound, boundtype). I would say no, we probably don't want to maintain this nor write documentation for it, but @mattmilten, @mueldgog  what do you think?

src/pyscipopt/scip.pyx
```diff
@@ -547,6 +649,55 @@ cdef class Node:
...
+        SCIPnodeGetParentBranchings(self.scip_node, branchvars, branchbounds,
+                                    boundtypes, &nbranchvars, nbranchvars)
+
+        branchings = [Branching.create(branchvars[i], branchbounds[i], boundtypes[i])
+                      for i in range(nbranchvars)]
+
+        free(boundtypes)
+        free(branchbounds)
+        free(branchvars)
+        return branchings
```
fserra: I guess the discussion about the branching class is going to be relevant here. So, in principle, this function should just return a tuple of variables, bounds and boundtype.

yozw: This is indeed how I implemented this originally, but it was a bit clunky to use because (a) the return values are arranged in an unnatural way (by variables, bounds, boundtype rather than by branching), and (b) printing the return values was not helpful. (In the current form, printing out the return value gives a list of inequalities of the form x_i <= u_i, etc.)

src/pyscipopt/scip.pxd
```diff
@@ -620,6 +631,28 @@ cdef extern from "scip/scip.h":
...
+                                     int branchvarssize)
+    void SCIPnodeGetAddedConss(SCIP_NODE* node, SCIP_CONS** addedconss,
+                               int* naddedconss, int addedconsssize)
+    void SCIPnodeGetNDomchg(SCIP_NODE* node, int* nbranchings, int* nconsprop,
+                            int* nprop)
+    SCIP_DOMCHG* SCIPnodeGetDomchg(SCIP_NODE* node)
+
+    # Domain change methods
+    int SCIPdomchgGetNBoundchgs(SCIP_DOMCHG* domchg)
+    SCIP_BOUNDCHG *SCIPdomchgGetBoundchg(SCIP_DOMCHG* domchg, int pos)
```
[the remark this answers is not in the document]
yozw: Done

src/pyscipopt/scip.pyx
```diff
@@ -510,6 +595,10 @@ cdef class Node:
         node.scip_node = scipnode
         return node
 
+    def isRoot(self):
+        """Returns whether this is the root node."""
+        return self.getParent() is None
+
```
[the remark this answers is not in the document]
yozw: Done

src/pyscipopt/scip.pxd
yozw: I've removed this class.

src/pyscipopt/scip.pyx
yozw: I've changed the method so that it returns a tuple of variables, bounds, boundtypes.

fserra:
you probably need to check if the depth is 0 before calling getparentbranching

yozw:
Egh. It's the end of the day :)  Do you prefer getParentBranchings to return [] or None if called on the parent node?

fserra:
None sounds better to me

fserra:
thanks!

PR #340, opened 2020-01-31
Add Row.isRemovable and Row.getOrigintype

yozw:


mattmilten:
Thanks for your contribution!

PR #354, opened 2020-03-24
Changes to PySCIPOpt for Benders' API in SCIP 7.0

stephenjmaher:
Here are some changes implemented to update the PySCIPOpt to represent the changes to the Benders' API in SCIP 7.0.


stephenjmaher:
I suspect that before this merge should happen, then we should merger `master` into `scip-7`. The latter branch appeared to be quite a bit behind the former.

stephenjmaher:
Both `scip-7` and this branch are up to date. This pull request can be reviewed now.

fserra:
mmm I guess all check are going to fail because the ci tries to compile against SCIP 6

stephenjmaher:
Yea, I just noticed that. How do we proceed? Should we just wait for the release?

src/pyscipopt/scip.pyx
```diff
@@ -3075,11 +3075,13 @@ cdef class Model:
         """
         SCIPbendersSetSubproblemIsConvex(benders._benders, probnumber, isconvex)
 
-    def setupBendersSubproblem(self, probnumber, Benders benders = None, Solution solution = None):
+    def setupBendersSubproblem(self, probnumber, checktype, Benders benders = None, Solution solution = None):
```
fserra: wanna give a default value to checktype?

fserra:
nope, we just merge and check against our Jenkins job if everything is alright.

Why do you add the change bound probing functions here?

src/pyscipopt/scip.pyx
stephenjmaher: I thought about it, but didn't know what the default should be because the most common are `LP` and `CHECK`. I think by making a default, it doesn't highlight to the user that a different option will be needed in some circumstances.

stephenjmaher:
I added the probing bound changes because it was included on my branch I was using. We can take it out if you want.

fserra:
yes let's take it out, that can go directly to master. Though you would need to add a comment about the None being interpret as -infinity. Then I just merge and we check against the Jenkins job

stephenjmaher:
The probing bound changes have now been removed.

fserra:
you missed the changelog entry!

stephenjmaher:
how about now

PR #364, opened 2020-04-11
add SCIP version compatibility table

mattmilten:


ambros-gleixner:
I moved the table to INSTALL.md, because I think this is where it should also be, and two locations are more work to update.

PR #378, opened 2020-05-20
Change Model.from_ptr/to_ptr to use PyCapsule

AntoinePrv:
Follow up on on #330 to use PyCapsule

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # CHANGELOG
 
+## Unreleased
+### Changed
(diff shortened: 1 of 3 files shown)


fserra:
looks good, except for the failing test. i guess it is failing because it now no longer compares the pointers but the capsule objects, and every time a new (different) capsule object is created. Is there a way to extract the pointer from the capsule to fix the test?

AntoinePrv:
I don't know how I let that slide (I guess that's why there is CI 😅). I replace it with model comparison.
Travis seems to have problem fetching SCIP here though.

mattmilten:
I just fixed the wget certificate issue. Can you merge master again? It should work then.

PR #379, opened 2020-05-22
Add branch priority

jmTUD:
This pull request adds a pure wrapper of the SCIP function "SCIPchgVarBranchPriority" from "scip/scip_var.h".
The python version can be accessed from `Model`.chgVarBranchPriority(`Variable` var, priority).

No tests/examples are provided, since it is a direct wrapper of a SCIP method.

### DIFF
--- a/src/pyscipopt/scip.pyx
+++ b/src/pyscipopt/scip.pyx
@@ -4537,6 +4537,15 @@ cdef class Model:
 
         free(_coeffs)
 
+    def chgVarBranchPriority(self, Variable var, priority):
+        """Set the branch priority of a variable.
+
+        :param Variable var: variable to change priority of
+        :param priority: the new priority of the variable (each variable is initialized with 0 as priority)
+        """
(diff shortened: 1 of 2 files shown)


fserra:
Hi thanks for the contribution, could you merge master into your branch so that the test can pass.

src/pyscipopt/scip.pyx
```diff
@@ -4537,6 +4537,15 @@ cdef class Model:
 
         free(_coeffs)
 
+    def chgVarBranchPriority(self, Variable var, priority):
+        """Set the branch priority of a variable.
```
fserra, line 4541: can you copy verbose the documentation from scip, please:

> sets the branch priority of the variable; variables with higher branch priority are always preferred to variables with lower priority in selection of branching variable

jmTUD: Thank you for your comment. I changed it.

fserra:
thanks for your contribution!

Issue #385, opened 2020-06-03
correct function to get dual variables


glliu:
Hi,

I was implementing a column generation algorithm with PySCIPOpt, now I'm a little confused about getting the dual variables.

In test_pricer.py, the pricer get the dual variables with Model.getDualMultiplier(c). However, in Issue #136 Model.getDualsolLinear(c) is used to access the dual variables. In addition, getDualMultiplier() and getDualsolLinear() often give different values. So what's the difference between these two functions? And which one is the proper function to get the dual variables?

Thank you very much.

fserra:
Hi, yeah this is bad. You should use getDualMultiplier.

@mattmilten we should remove getDualsolLinear and rename getDualMultiplier to getDualsolLinear :)

The problem is that SCIP does not support solving an LP problem and then accessing the dual variables. Only in some special cases, this works, and we tried to improve this with getDualsolLinear.

However, during the solving process of SCIP, SCIP solves, internally, LPs. For those, you can definitely access the dual information. This is what should happen in the pricer and for this, the function getDualMultiplier should be used.

glliu:
Thanks for your reply 👍

PR #386, opened 2020-06-04
update getDualMultiplier and getDualsolLinear

mattmilten:
according to #385 

This marks getDualMultiplier as deprecated and getDualsolLinear actually behaves as intended (wrapping the corresponding SCIP function).


fserra:
oh this is beautiful, thanks!

mattmilten:
... and the third attempt is even correct! ;-)

Issue #395, opened 2020-06-25
Solving BPP using B&P in pyscipopt


abb-omidi:
Dear community team,

First of all, thanks for your interesting SCIP python wrapper. I would like to use SCIP as a branch and price framework and am trying to do on the win10 using CMake. It works fine but, as I don't have a good background in the C/C++ programming, it is a bit hard to understand its functions. I read other issues about that and understand it's not possible for some examples like test_coloring. ([this](https://github.com/SCIP-Interfaces/PySCIPOpt/issues/281) and [this](https://github.com/SCIP-Interfaces/PySCIPOpt/issues/394) issues.). 
I'm wondering if, is there BPP or VRP example in which implementing BP/BCP in python?

Regards
Abbas

fserra:
Hi Abbas

Sorry, but currently there is no such example in python, only the ones available in SCIP (https://scipopt.org/doc/html/BINPACKING_MAIN.php and https://scipopt.org/doc/html/VRP_MAIN.php)
The coloring was translated from https://scipopt.org/doc/html/COLORING_MAIN.php but back then we were trying to test some extensions to the interface, thus, sadly,  it doesn't work with the current pyscipopt.

I might spend some time trying to make the coloring work, but I don't promise anything.

Steve8881:
Hi there

I am trying to implement a branch-and-price algorithm in Python to solve the bin packing problem, mainly based on the coloring example, the pricer example and the branching example in PySCIPOpt, and the bin packing example in SCIP. When `Model.branchVar()` is applied, everything works well (though the performance is not satisfying). However, when I try to implement the Ryan/Foster branching rule with a SameDiff constraint handler, a segmentation fault occurs (`Segmentation fault: 11`). After some debugging, the problem seems to be caused by the `consactive()` method and I have no idea how to do the further debugging. Could you guys give me some guidance to solve the problem? 

Thank you.

Steve

PS: I have implemented the `Model.repropagateNode()` function by myself and used the `Model.getObjVal()` function to replace the `SCIPgetLocalTransEstimate()` function. Here is the code:

```python

from pyscipopt import *
import numpy as np
import inspect

entering = lambda: print("IN ", inspect.stack()[1][3]) # debugging magic
leaving = lambda: print("OUT ", inspect.stack()[1][3]) # debugging magic

class SameDiff(Conshdlr):

    def consdataCreate(self, name, item1, item2, constype, node):
        entering()
[...]

PR #410, opened 2020-08-06
extends interface

fserra:
some functions I needed to try to reproduce the behavior from #395


src/pyscipopt/scip.pyx
```diff
@@ -877,6 +877,10 @@ cdef class Constraint:
         """Retrieve True if constraint is only locally valid or not added to any (sub)problem"""
         return SCIPconsIsStickingAtNode(self.scip_cons)
 
+    def isActive(self):
+        """returns TRUE iff constraint is active in the current node"""
```
mattmilten, line 881:
```suggestion
        """returns True iff constraint is active in the current node"""
```

Issue #416, opened 2020-08-21
addConsNode currently unusable?


CharJon:
Hi everyone,

so I might be missing something, but it seems to me as if the interface is lacking functionality to make use of `addConsNode`.
To implement some custom branching rules I want to use `makeChild` and `addConsNode`. The later expects a constraint as a `Constraint` object as input (next to a node to add the constraint to). But I can not wrap my head around how to transform my constraint of type `ExprCons` into a constraint. I tried adding it to the model via `addCons`, but this lead to problems, as I don't want the constraint to be part of the model in general, but just stick to one child (and it's children).
I might be misunderstanding something as my SCIP knowledge is fairly limited, but the constraint handlers make use of the `createCons` method, to generate special constraints. Do the branching rules need something similar, or am I doing something wrong?
I appreciate any help and/or comment.

fserra:
does #391 answer your question?

CharJon:
Hi, thanks for your answer. #391 is indeed closely related to my problem.
After reading the information there, I still feel like extending the Interface is the only solution for my case, as `addCons` does not work for me.
What I did for now to solve my problem is also wrapping [SCIPcreateConsLinear](https://www.scipopt.org/doc-7.0.1/html/group__CONSHDLRS.php#gaea3b4db21fe214be5db047e08b46b50e) and using Constraints created this way for `addConsNode`. This seems to work as expected, I'm just not familiar with wrapping and get a `Process finished with exit code 139 (interrupted by signal 11: SIGSEGV)` after everything finishes.

If this is the preferred solution and if someone could help me with the SIGSEGV I could create a pull-request to extend the functionality.

fserra:
I think it would make sense to modify `addConsNode` to also accept `ExprCons`, maybe only when the `ExprCons` is linear.

CharJon:
This works as well.

Right now I'm basically doing the same as Model._addLinCons does, just without `PY_SCIP_CALL(SCIPaddCons(self._scip, scip_cons))` before and `PY_SCIP_CALL(SCIPreleaseCons(self._scip, &scip_cons))` after `PyCons = Constraint.create(scip_cons)`.
If I add the constraint creation via `PY_SCIP_CALL(SCIPcreateConsLinear(...))` into `addConsNode`, how do I prevent a SIGSEGV?

fserra:
I don't understand what you are saying. You can open a PR so we can see the code, but my guess is that it should look something like:

```
[...]

PR #418, opened 2020-08-24
Change addConsNode to (also) accept ExprCons

CharJon:
Based on discussion in #416 
To keep the prototype simple a new addExprConsNode method is added.
Do you prefer having addConsNode accept ExprCons only or both (Constraint and ExprCons)?

### DIFF
--- a/src/pyscipopt/scip.pyx
+++ b/src/pyscipopt/scip.pyx
@@ -2237,7 +2237,6 @@ cdef class Model:
                                    propagate, removable, separate,stickingatnode)
             py_cons = self._createCons(cons, **kwargs)
             scip_cons = <SCIP_CONS*> py_cons.scip_cons
-            PY_SCIP_CALL(SCIPreleaseCons(self._scip, &scip_cons))
         else:
             raise Warning(f"Argument cons is of incorrect type ({type(cons)}).")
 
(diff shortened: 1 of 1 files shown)


fserra:
thanks. I think it would make sense to factor out the code that creates the linear constraint since it also appears in `_addLinCons`

Regarding the name and if it should also be part of `addConsNode` I am unsure. I am pinging @mattmilten here.
So we have `addCons` and `addPyCons` for when the argument is an `ExprCons` or a `Constraint`, resp.
So for consistency, we could have `addConsNode` and `addPyConsNode`. 
But then we also have `addConsLocal` and we would need to do the same there.

An alternative would be to remove `addPyCons` and let `addCons` handle both cases.

CharJon:
Personally I think the addPyCons naming is a bit confusing, as the `Constraint` objects are the ones actually having a pointer to a SCIPConstraint, while the ExprCons `addCons` expects as input don't. If you prefer having a single method for both cases I will restructure the code.

mattmilten:
I prefer having a nice and easy-to-use method `addCons` that just accepts whatever you throw at it - be it `ExprCons` or `Constraint`.

CharJon:
I made changes to addCons and addConsNode to avoid duplicate code. The current version of `addConsNode` works for `Constraint` and `ExprCons`. There a two TODOs in the code as I'm still unsure about how to handle the reference counting of SCIP. If you have any sources I could study to understand why and when `PY_SCIP_CALL(SCIPreleaseCons(self._scip, &scip_cons))` and `Py_INCREF(cons)` are necessary please tell me.

mattmilten (commented):
Looks good to me. Just the CHANGELOG entry is missing.

SanielDous:
Did this find its way into master branch already? Would be really nice

zzy-opt:
I am implementing the branch and price where I need to add a linear constraint to the newly created node in branching rule. I am facing the exact same problem stated here.
The suggest solution definitely can solve my problem.  Howerever I don't see this updated version of addConsNode function exists. What goes wrong here? Isn't it successfully merged?

zzy-opt:
@SanielDous Thank you. It seems this cannot be fixed in short term. Maybe I had better figure out how to use cython and change source code myself.

PR #442, opened 2020-10-09
fix getTransformedVar

### LINKED ISSUE #439: Problem to release variables for branch-and-price Graph Coloring
Hi,

I have worked on a Python Graph coloring code initially proposed by the SCIP team, eventually getting a working version.
However I couldn’t manage to release variables as no ‘releaseVar()’ instruction exists on Pyscipopt, getting pestered by such warnings:

_WARNING: Transformed variable <t_Initial_Set_1> not released when freeing SCIP.
WARNING: Transformed variable <t_Initial_Set_0> not released when freeing SCIP.
WARNING: Original variable <Initial_Set_38> not released when freeing SCIP.
WARNING: Original variable <Initial_Set_37> not released when freeing SCIP, etc._

What would be the way to release variables in Python, or is there any specific recommendation to avoid such warnings? 

The code is available at https://github.com/pchervi/Graph-Coloring and these warnings can be reproduced with any graph proposed, even a simple 3-wheel graph.

Thank you for your help!

fserra:
`getTransformedVar` was using `SCIPtransformVar` which on top of transforming the variable, it captures it. Thus, the variables were unreleased at the end.
Fixes #439

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Fixed
 - strip quotes from SCIPOPTDIR path variable that might confuse Windows systems
+* `getTransformedVar` now uses `SCIPgetTransformedVar` instead of `SCIPtransformVar` which captures the variable
 
 ## 3.0.2 - 2020-08-09
 ### Added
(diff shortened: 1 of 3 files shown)


mattmilten (approved):
Cool!

PR #451, opened 2020-11-15
Feature/solution evaluates expr

mcitoler:
Following up on https://github.com/SCIP-Interfaces/PySCIPOpt/issues/280 and https://github.com/SCIP-Interfaces/PySCIPOpt/pull/308#issuecomment-527368035,  added support to evaluate polynomial expressions in solutions via `__getitem__`,  e.g. `solution[x*y]` will evaluate the expression `x*y` according to the variable values in `solution`.

### DIFF
--- a/src/pyscipopt/conshdlr.pxi
+++ b/src/pyscipopt/conshdlr.pxi
@@ -259,8 +259,7 @@ cdef SCIP_RETCODE PyConsSepasol (SCIP* scip, SCIP_CONSHDLR* conshdlr, SCIP_CONS*
     cdef constraints = []
     for i in range(nconss):
         constraints.append(getPyCons(conss[i]))
-    solution = Solution()
-    solution.sol = sol
+    solution = Solution.create(scip, sol)
     result_dict = PyConshdlr.conssepasol(constraints, nusefulconss, solution)
     result[0] = result_dict.get("result", <SCIP_RESULT>result[0])
     return SCIP_OKAY
@@ -280,8 +279,7 @@ cdef SCIP_RETCODE PyConsEnforelax (SCIP* scip, SCIP_SOL* sol, SCIP_CONSHDLR* con
     cdef constraints = []
     for i in range(nconss):
         constraints.append(getPyCons(conss[i]))
-    solution = Solution()
-    solution.sol = sol
+    solution = Solution.create(scip, sol)
     result_dict = PyConshdlr.consenforelax(solution, constraints, nusefulconss, solinfeasible)
     result[0] = result_dict.get("result", <SCIP_RESULT>result[0])
     return SCIP_OKAY
@@ -302,8 +300,7 @@ cdef SCIP_RETCODE PyConsCheck (SCIP* scip, SCIP_CONSHDLR* conshdlr, SCIP_CONS** 
     cdef constraints = []
     for i in range(nconss):
         constraints.append(getPyCons(conss[i]))
-    solution = Solution()
-    solution.sol = sol
+    solution = Solution.create(scip, sol)
     result_dict = PyConshdlr.conscheck(constraints, solution, checkintegrality, checklprows, printreason, completely)
     result[0] = result_dict.get("result", <SCIP_RESULT>result[0])
     return SCIP_OKAY
--- a/src/pyscipopt/expr.pxi
+++ b/src/pyscipopt/expr.pxi
@@ -109,13 +109,7 @@ class Term:
     def __repr__(self):
         return 'Term(%s)' % ', '.join([str(v) for v in self.vartuple])
 
-    def _evaluate(self, point):
-        '''computes the value of the term in the given variable-value mapping.'''
-        prod = 1
-        for var in self.vartuple:
-            prod *= point[var]
-        return prod
-        
+
 CONST = Term()
 
 # helper function
@@ -288,10 +282,6 @@ cdef class Expr:
             return 0
         else:
             return max(len(v) for v in self.terms)
-
-    def _evaluate(self, point):
-        '''computes the value of the expression in the given variable-value mapping.'''
-        return sum(term._evaluate(point)*coeff for term, coeff in self.terms.items() if coeff != 0)
 
 
 cdef class ExprCons:
--- a/src/pyscipopt/scip.pyx
+++ b/src/pyscipopt/scip.pyx
@@ -538,8 +538,12 @@ cdef class Solution:
         sol.scip = scip
         return sol
 
-    def __getitem__(self, Variable var):
-        return SCIPgetSolVal(self.scip, self.sol, var.scip_var)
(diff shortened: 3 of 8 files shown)


src/pyscipopt/scip.pyx
```diff
@@ -4140,11 +4144,9 @@ cdef class Model:
         """
         if sol == None:
             sol = Solution.create(self._scip, NULL)
-        if isinstance(expr, Variable):
-            var = <Variable> expr
-            return SCIPgetSolVal(self._scip, sol.sol, var.scip_var)
-        else:
-            return expr._evaluate(sol)
+        if self._scip != (<Solution>sol).scip:
```
mcitoler, line 4147: The idea behind the change in `getSolVal` is to make `Solution.__getitem__` the only one responsible for retrieving values for any `Expr`. However,  it appears that sometimes `sol.scip` is out of sync with `self._scip` and causes a segmentation fault error. I've observed this in https://github.com/SCIP-Interfaces/PySCIPOpt/blob/3fea23e9c993cdfea3e4f13f52463e29d5fbf735/tests/test_alldiff.py#L184 and https://github.com/SCIP-Interfaces/PySCIPOpt/blob/3fea23e9c993cdfea3e4f13f52463e29d5fbf735/tests/test_tsp.py#L23.
I suspect that the reason has something to do with how references to model are kept in the different plugins https://github.com/SCIP-Interfaces/PySCIPOpt/blob/3fea23e9c993cdfea3e4f13f52463e29d5fbf735/src/pyscipopt/scip.pyx#L3384 but I don't know how to fix it. @mattmilten @fserra, do you have any suggestions?

This makes me think that maybe the `Model` class should be the one in charge of evaluating `Expr` instead of `Solution`. In this case, `Solution` would need some back reference to the model. Another alternative would be to repeat the logic both in `Solution.__getitem__` and `Model.getSolVal` but I would prefer to avoid that if possible.

mattmilten: Hmm, I don't see the issue here. All tests are running fine, right?

Also, please excuse the delay in replying!

mattmilten (approved):

src/pyscipopt/scip.pyx
fserra: mmm I think this shouldn't be happening. It is a bit inefficient to create a python object to wrap the solution every time one accesses a value. e.g. from test_tsp
```
        for (i, j) in x:
            if self.model.getSolVal(solution, x[i, j]) > EPS:
```
For that particular code, I think the problem is that in `conshdlr.pxi` we do

```
cdef SCIP_RETCODE PyConsCheck (SCIP* scip, SCIP_CONSHDLR* conshdlr, SCIP_CONS** conss, int nconss, SCIP_SOL* sol, SCIP_Bool checkintegrality,
                               SCIP_Bool checklprows, SCIP_Bool printreason, SCIP_Bool completely, SCIP_RESULT* result):
    PyConshdlr = getPyConshdlr(conshdlr)
    cdef constraints = []
    for i in range(nconss):
        constraints.append(getPyCons(conss[i]))
    solution = Solution()
    solution.sol = sol
    result_dict = PyConshdlr.conscheck(constraints, solution, checkintegrality, checklprows, printreason, completely)
    result[0] = result_dict.get("result", <SCIP_RESULT>result[0])
    return SCIP_OKAY
```
So when the solution is created, we do not set its `scip` field. Maybe that would be enough to remove the second if.

Regarding the first if, i.e.,
```
        if sol == None:
            sol = Solution.create(self._scip, NULL)
```
We could just call directly `SCIPvarGetSol` instead of creating a solution object. (anyway SCIP should end up calling `SCIPvarGetSol` internally, unless I am missing something)

mcitoler: @fserra, you are absolutely right, the problem was that the `scip` field was not being set. I've seen a few other places with the same issue and I've updated them to use `Solution.create(scip, sol)`. Regarding `SCIPvarGetSol` I'm not sure what you mean, as far as I can tell the function returns a real number https://scipopt.org/doc-7.0.2/html/group__PublicVariableMethods.php#ga110d0d9fad1f59b009126a84add117a8.

src/pyscipopt/conshdlr.pxi
```diff
@@ -259,8 +259,7 @@ cdef SCIP_RETCODE PyConsSepasol (SCIP* scip, SCIP_CONSHDLR* conshdlr, SCIP_CONS*
     cdef constraints = []
     for i in range(nconss):
         constraints.append(getPyCons(conss[i]))
-    solution = Solution()
-    solution.sol = sol
+    solution = Solution.create(scip, sol)
```
mcitoler, line 262: My only concern with the switch to `Solution.create` is that it raises and error if `scip` is NULL, which it is not the same as the current behavior. @fserra, @mattmilten is that ok? See `src/pyscipopt/sepa.pxi` as well.

fserra: I would argue that SCIP shouldn't be NULL... if it is NULL I guess there are bigger problems

src/pyscipopt/scip.pyx
fserra: I can't find the code anymore ;) but let me comment in another change what I mean

src/pyscipopt/scip.pyx
```diff
@@ -4167,11 +4171,7 @@ cdef class Model:
         """
         if sol == None:
             sol = Solution.create(self._scip, NULL)
-        if isinstance(expr, Variable):
-            var = <Variable> expr
-            return SCIPgetSolVal(self._scip, sol.sol, var.scip_var)
-        else:
-            return expr._evaluate(sol)
+        return sol[expr]
```
fserra, line 4174: So here. before there was this code checking whether expr is a variable. What I would argue is that if expr is a variable and sol is None, then we call directly `SCIPgetSolVal(self._scip, NULL, var.scip_var)` or `SCIPvarGetSol`. My point is that we do not need to create a wrapper for the NULL solution in such a case

fserra: actually, we could even have a nullsol stored in model and we always use that one. @mattmilten @CGraczyk ?

src/pyscipopt/expr.pxi
```diff
@@ -109,13 +109,7 @@ class Term:
...
         return 'Term(%s)' % ', '.join([str(v) for v in self.vartuple])
 
-    def _evaluate(self, point):
-        '''computes the value of the term in the given variable-value mapping.'''
-        prod = 1
-        for var in self.vartuple:
-            prod *= point[var]
-        return prod
-        
+
```
mcitoler, line 112: I've have taken the liberty to simplify the implementation of how `Expr` are evaluated. Before there was some recursion that was a bit confusing and I think that now it is more clear with everything being handled under `Solution`.

PR #477, opened 2021-02-11
Add Conda installation instructions

AntoinePrv:
Hello,

SCIP and pyscipopt are now available on Conda-Forge! 🚀 
Here is a small proposal for the installation, but feel free to edit as you prefer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,19 @@ See [CHANGELOG.md](CHANGELOG.md) for added, removed or fixed functionality.
 Installation
 ------------
 
+**Using Conda**
+
+[![Conda (channel only)](https://img.shields.io/conda/vn/conda-forge/pyscipopt?logo=conda-forge)](https://anaconda.org/conda-forge/pyscipopt)
+![Conda](https://img.shields.io/conda/pn/conda-forge/pyscipopt?logo=conda-forge)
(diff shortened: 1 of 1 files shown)


CGraczyk:
Hey Antoine,
Thanks for the update, looks fine to me, do you agree @mattmilten ?

mattmilten:
Neat! I would remove the last badge, though. The download counter might look impressive, but it's vastly overestimating the true number of downloads or users and the platform badge should link to the anaconda page just like the first badge. There is no Windows version, then? This would be really helpful for many users...

AntoinePrv:
I don't know why the download count is so inflated. 2.5k with scip 250, that's not really possible.

Anyways, windows is not available yet, but I don't think there is anything preventing it from being there.
It's low priority for me right now, perhaps in the next few months.

mattmilten:
Sure, it's just that most people who struggle with the installation are
using Windows. And people will ask for Windows support.

AntoinePrv:
If you know someone knowledgeable on Windows (CMake and `bat` scripts), I could help on the Conda side.

mattmilten:
Did you check the [Appveyor configuration file](https://github.com/scipopt/PySCIPOpt/blob/master/appveyor.yml)? It should contain all required commands to install the SCIPOptSuite and PySCIPOpt. Let me know if you need some help.

PR #484, opened 2021-03-15
enum that wraps SCIP_STAGE

LovisAnderson:
`getStage()` used to return an int which is hard to interpret. I changed it, such that it returns an enum `PY_SCIP_STAGE`. 

What I did:
* define `enum PY_SCIP_STAGE` on both C and python side 
* `model.getStage()` returns that enum
* changed stage comparisons to python enum accordingly

Before:
```
print(m.getStage())
1
```
Now:
```
print(m.getStage())
PY_SCIP_STAGE.PROBLEM
```
If you agree with my methodolgy, this could be done for other enums as well, as it would make them human readable.

### DIFF
--- a/src/pyscipopt/scip.pyx
+++ b/src/pyscipopt/scip.pyx
@@ -101,7 +101,7 @@ cdef class PY_SCIP_STATUS:
     UNBOUNDED      = SCIP_STATUS_UNBOUNDED
     INFORUNBD      = SCIP_STATUS_INFORUNBD
 
-cdef class PY_SCIP_STAGE:
+cpdef enum PY_SCIP_STAGE:
     INIT         = SCIP_STAGE_INIT
     PROBLEM      = SCIP_STAGE_PROBLEM
     TRANSFORMING = SCIP_STAGE_TRANSFORMING
@@ -2729,7 +2729,7 @@ cdef class Model:
         cdef SCIP_Real activity
         cdef SCIP_SOL* scip_sol
 
-        if not self.getStage() >= SCIP_STAGE_SOLVING:
+        if not self.getStage() >= SOLVING:
             raise Warning("method cannot be called before problem is solved")
 
         if isinstance(sol, Solution):
@@ -2762,7 +2762,7 @@ cdef class Model:
         cdef SCIP_SOL* scip_sol
 
 
-        if not self.getStage() >= SCIP_STAGE_SOLVING:
+        if not self.getStage() >= SOLVING:
             raise Warning("method cannot be called before problem is solved")
 
         if isinstance(sol, Solution):
@@ -4162,7 +4162,7 @@ cdef class Model:
         :param original: objective value in original space (Default value = True)
 
         """
-        if not self.getStage() >= SCIP_STAGE_SOLVING:
+        if not self.getStage() >= SOLVING:
             raise Warning("method cannot be called before problem is solved")
         return self.getSolObjVal(self._bestSol, original)
 
@@ -4191,7 +4191,7 @@ cdef class Model:
 
         Note: a variable is also an expression
         """
-        if not self.getStage() >= SCIP_STAGE_SOLVING:
+        if not self.getStage() >= SOLVING:
             raise Warning("method cannot be called before problem is solved")
         return self.getSolVal(self._bestSol, expr)
 
@@ -4217,7 +4217,7 @@ cdef class Model:
 
     def getStage(self):
         """Retrieve current SCIP stage"""
-        return SCIPgetStage(self._scip)
+        return PY_SCIP_STAGE(SCIPgetStage(self._scip))
 
     def getStatus(self):
         """Retrieve solution status."""


mattmilten:
I don't see why this is necessary. The typical use-case is to check the return code with the respective enum:

    def getStatus(self):
        """Retrieve solution status."""
        cdef SCIP_STATUS stat = SCIPgetStatus(self._scip)
        if stat == SCIP_STATUS_OPTIMAL:
            return "optimal"

In your case, it would look like this:

    stage = m.getStage()
    if stage == SCIP_STAGE.PROBLEM:
        print("problem stage")

The `PY_` enums are not supposed to be used or visible. That's why there are wrappers to map them to the proper SCIP terminology (in `__init__.py`):

    from pyscipopt.scip      import PY_SCIP_STAGE        as SCIP_STAGE

LovisAnderson:
I think it would be a good change for usability. I think a common case a user experiences is, that some method raises a warning, since you are in the wrong stage (e.g. assume calling `m.getObjVal()` before the solving stage). In order to find out in which stage I  actually am, I found myself going to the pyscipopt documentation, then to the scip documentation, only to match the number. 

Even if you want to check against a stage, it would make it way easier to find them, as a user will most likely not know how to import `SCIP_STAGE`. Overall,  I see no argument why `getStage` would return an `int` and not an `SCIP_STAGE`  of some form. I agree that it would be preferable to have exactly the same terminology as in SCIP.

Another way would be like implemented in `m.getStatus()`, where the enum is matched and a string is returned. I do not think that this is a good solution though. I think this invites all kind of problems. E.g. you have to remember changing the matching function everytime you change the enum. Or doing a `+1` on a stage as string would append the one instead of increasing the stage, `>/<` obv. would also work different than you expect.

mattmilten:
Hmm, I am not convinced. I still think there might be more complications down the road.
I agree, though, that the current implementation is inconsistent - `getStatus()` for example should also just return an int instead of a string.

LovisAnderson:
> Hmm, I am not convinced. I still think there might be more complications down the road.
> I agree, though, that the current implementation is inconsistent - `getStatus()` for example should also just return an int instead of a string.

Ok, too bad that I could not convince you while I am still fully convinced :/ I personally think enums would be far better than int returns, or maybe I missed something completely? But I also see that it would be a bigger change, if you want to be consistent. If it has not high priority and you do not want to be inconsistent I understand that.

I just think it would be a good step towards usability of pyscipopt  for non-experts. With  jupyter-notebooks and the way people develop and test their models there, I think people (at least I am) call methods like `getStage` or `getStatus` often to check what happened. It's hard for them to search in the documentation to find out if the model was unbounded or infeasible or what stage it is in (I even had to go to the C source code of SCIP to find out what the return of `getLPSolstat` means).

Have a nice sunday evening :)

fserra:
sorry @mattmilten but what is the problem with the enum of @LovisAnderson?
I mean, it is infinitely better
```
print(m.getStage())
PY_SCIP_STAGE.PROBLEM
```
than returning 1.

And if we can get that with something like
```
return PY_SCIP_STAGE(SCIPgetStage(self._scip))
```
instead of the similar
```
    def getStatus(self):
        """Retrieve solution status."""
        cdef SCIP_STATUS stat = SCIPgetStatus(self._scip)
        if stat == SCIP_STATUS_OPTIMAL:
            return "optimal"
        elif stat == SCIP_STATUS_TIMELIMIT:
            return "timelimit"
        elif stat == SCIP_STATUS_INFEASIBLE:
            return "infeasible"
        elif stat == SCIP_STATUS_UNBOUNDED:
            return "unbounded"
        elif stat == SCIP_STATUS_USERINTERRUPT:
            return "userinterrupt"
        elif stat == SCIP_STATUS_INFORUNBD:
            return "inforunbd"
        elif stat == SCIP_STATUS_NODELIMIT:
            return "nodelimit"
        elif stat == SCIP_STATUS_TOTALNODELIMIT:
            return "totalnodelimit"
        elif stat == SCIP_STATUS_STALLNODELIMIT:
            return "stallnodelimit"
        elif stat == SCIP_STATUS_GAPLIMIT:
            return "gaplimit"
        elif stat == SCIP_STATUS_MEMLIMIT:
            return "memlimit"
        elif stat == SCIP_STATUS_SOLLIMIT:
            return "sollimit"
        elif stat == SCIP_STATUS_BESTSOLLIMIT:
            return "bestsollimit"
        elif stat == SCIP_STATUS_RESTARTLIMIT:
            return  "restartlimit"
        else:
            return "unknown"
```
then I think it is cool.

What are the technical difficulties with having it as python enums?

I kind of see that the normal use case would just be to do something like
`if model.getStatus() == SCIP_STATUS.BLA then ...`. But I have to agree with Lovis, if somebody wants to know that status and model.getStatus() returns 1, then it is **completely** useless. (I always remember when debugging something in SCIP and I asked the debugger for the var status and it returned 4... 4? what the hell is 4? so annoying)

mattmilten:
OK, then. Let's go for it. I am not going to block progress or anything. I was thinking of maintaining the "SCIP" way of doing this, I guess. Aren't people going to get confused when there are `PY_SCIP_xyz` enums and `SCIP.xyz` classes that actually refer to the same thing?

fserra:
But do we need the `SCIP.xyz` classes? Couldn't we remove them and rename the enums to `SCIP_xyz` (instead of `PY_SCIP_xyz`?)

LovisAnderson:
Glad that we could win you over @mattmilten and thanks @fserra for convincing.

> But do we need the `SCIP.xyz` classes? Couldn't we remove them and rename the enums to `SCIP_xyz` (instead of `PY_SCIP_xyz`?)

I think the `SCIP.xyz` classes are needed, as they are needed to define the wrapper class `PY_SCIP_xyz`. But I think one could  import the C objects with a different name, as in `from *** cimport SCIP.xyz as C_SCIP,xyz` and then define `SCIP_xyz` as you proposed. This does not work if imported from a pxd file with the same name as the file you import to apparently. But it should be possible if imported from a different file. The approach is described here: https://cython.readthedocs.io/en/latest/src/userguide/sharing_declarations.html#using-cimport-to-resolve-naming-conflicts

I tried, but I have problems with setup/cythonize finding my new .pxd file.

fserra:
Mmm. I am still not fully understand this. So if it would require some maintenance, then the decision should be from @CGraczyk 

Given that the `PY_SCIP_STAGE` `PY_SCIP_STATUS` etc are `cdef class`, can't we implement the `__repr__` that just prints the name? Would that solve the issue?

LovisAnderson:
> Given that the `PY_SCIP_STAGE` `PY_SCIP_STATUS` etc are `cdef class`, can't we implement the `__repr__` that just prints the name? Would that solve the issue?

At least for `PY_SCIP_STAGE` I am sceptical that this alone does the trick. Using an enum here allows `<` like comparison, using `cdef class` does not. 

> Aren't people going to get confused when there are PY_SCIP_xyz enums and SCIP.xyz classes that actually refer to the same thing?

While I agree that this is not super elegant, is this really a problem? pyscipopt users will never see the `SCIP.xyz` class, or do they?

fserra:
> At least for PY_SCIP_STAGE I am sceptical that this alone does the trick. Using an enum here allows < like comparison, using cdef class does not. 

I think it does. You would just need to implement the `__richcmp__` or something... Can you give it a try @LovisAnderson ?

LovisAnderson:
> > At least for PY_SCIP_STAGE I am sceptical that this alone does the trick. Using an enum here allows < like comparison, using cdef class does not.
> 
> I think it does. You would just need to implement the `__richcmp__` or something... Can you give it a try @LovisAnderson ?

Hmm, but I still prefer the current approach. From my perspective the only downside is a PY_ at the beginning of the name as an artifact of wrapping and some ambiguouty due to  __init\__.py where the class is exposed as  `import PY_SCIP_STAGE as SCIP_STAGE`. Without my changes we already have the 
```
print(pyscipopt.SCIP_STAGE)
<class 'pyscipopt.scip.PY_SCIP_STAGE'>
```

It is not even clear to me that the PY at the beginning is a downside, as it indicates the wrapping. Expect for  the __init\__.py  there should be no confusion, since the user never sees the same object without `PY_`. And again, that confusion is not due to my code, just a bit more visible now.

With the `__repr__` approach you make the class look different to the outside than the actual object. Could this not lead to more confusion? Like, people look at the source code of `SCIP_STAGE` when they actually meant `PY_SCIP_STAGE`. It also seems like a lot more implementation effort, especially if used everywhere.

The cleanest solution would to have just SCIP_STAGE on both sides. https://stackoverflow.com/questions/54047125/wrapping-a-typedefed-enum-in-cython describes how it could work. Not for me though, when I tried to make it work for 2 hours. Note that I have quite limited knowledge of both C and Cython, so that might be the reason.

In my opinion the current approach is a clear improvement to status quo. And I can not spend much more time on this for now.

CGraczyk:
> Mmm. I am still not fully understand this. So if it would require some maintenance, then the decision should be from @CGraczyk
> 
> Given that the `PY_SCIP_STAGE` `PY_SCIP_STATUS` etc are `cdef class`, can't we implement the `__repr__` that just prints the name? Would that solve the issue?

What kind of maintenance would this encompass? Im generally in favour of reducing the difficulty for users to understand their code, but I dont know how these components are interconnected in the PySCIP framework overall.

CGraczyk:
Is there anything left to do here actually @fserra or should we merge this as is for a 4.2 release?

fserra:
I am sorry @CGraczyk, but I have no idea anymore :(. I guess if it works, then it should be ok. Sadly I don't have the time now to look into this... Maybe we can just ignore this since I can imagine @LovisAnderson doesn't care anymore and @mattmilten was against it in the first place

PR #485, opened 2021-03-18
addConsLocal updated and SCIPdelConsNode added

CGraczyk:
Based on pull request in #418 
Changelog updated.
Should the parameter "stickingatnode" in the addConsLocal function be True or False?

There are now two separate createCons functons, old:

    def createCons(self, Conshdlr conshdlr, name, initial=True, separate=True, enforce=True, check=True, propagate=True,
                   local=False, modifiable=False, dynamic=False, removable=False, stickingatnode=False):
        """Create a constraint of a custom constraint handler

        :param Conshdlr conshdlr: constraint handler
        :param name: name of constraint
        :param initial:  (Default value = True)
        :param separate:  (Default value = True)
        :param enforce:  (Default value = True)
        :param check:  (Default value = True)
        :param propagate:  (Default value = True)
        :param local:  (Default value = False)
        :param modifiable:  (Default value = False)
        :param dynamic:  (Default value = False)
        :param removable:  (Default value = False)
        :param stickingatnode:  (Default value = False)

        """

        n = str_conversion(name)
        cdef SCIP_CONSHDLR* scip_conshdlr
        scip_conshdlr = SCIPfindConshdlr(self._scip, str_conversion(conshdlr.name))
        constraint = Constraint()
        PY_SCIP_CALL(SCIPcreateCons(self._scip, &(constraint.scip_cons), n, scip_conshdlr, <SCIP_CONSDATA*>constraint,
                                initial, separate, enforce, check, propagate, local, modifiable, dynamic, removable, stickingatnode))
        return constraint
        
and the new one added in #418 :

        def _createCons(self, cons, **kwargs):
                deg = cons.expr.degree()
                cdef Constraint py_cons

        if deg <= 1:
            py_cons = self._createLinCons(cons, **kwargs)
        elif deg <= 2:
            py_cons = self._createQuadCons(cons, **kwargs)
        elif deg == float('inf'): # general nonlinear
            return self._createGenNonlinearCons(cons, **kwargs)
        else:
            return self._createNonlinearCons(cons, **kwargs)
        return py_cons
        
Is there a reason to have this distinction?


CGraczyk:
i restored the branch, but the issue here is that in the new SCIP version, the way nonlinear constraints are handled has changed dramatically, which is why that distinction is no longer valid as is - but maybe we can make it work for the new version and merge it.

mmghannam:
This has currently a lot of conflicts with the master branch, I will close it for now until it's ready to be merged.

PR #488, opened 2021-03-23
fix: use_cython bool different from imported function

LovisAnderson:
Small fix for setup.py. Here, a boolean `cythonize` was introduced that was later overwritten by `from Cython.Build import cythonize`. While `if cythonize` still evaluated to `True` at the right places, I don't think it was intended.

I also changed the formatting a bit to make it look the same (and PEP conform) and removed commented code. If you object, I can reverse those changes.


mattmilten:
Interesting! This looks like a "How did this ever work!?"
The format changes are also good. Thanks!

PR #494, opened 2021-04-23
fix for #493

mattmilten:
don't create a Solution object for a NULL SCIP_SOL*

Issue #493, opened 2021-04-23
Segfault when using Null solution


AntoinePrv:
## Problem
The following code crashes with `[1]    4536 segmentation fault (core dumped)  python3 crash.py`.
```python
import pyscipopt

model = pyscipopt.Model()

x = model.addVar("x")
y = model.addVar("y", vtype="INTEGER")
model.setObjective(x + y)
model.addCons(2*x - y*y >= 0)

sol = model.getBestSol()

# Any of this lines would segfault
print(sol)
model.getSolObjVal(sol)
```
If `print(sol)` is remove, the code runs fine.

## Diagnostic
For print, Gdb suggest segfaults is coming from `SCIPgetSolVal` in [Solution.__repr__](https://github.com/scipopt/PySCIPOpt/blob/4ccafc20bfc510bd07ca307fb780914eb3ebc4c8/src/pyscipopt/scip.pyx#L568).
Indeed, one can create a Solution with `SCIP_SOL == NULL`, but it is not checked in `__repr__` before passing it to `SCIPgetSolVal`.

For `getSolObjVal`, `NULL` is also [not checked](https://github.com/scipopt/PySCIPOpt/blob/4ccafc20bfc510bd07ca307fb780914eb3ebc4c8/src/pyscipopt/scip.pyx#L4143)

## Suggested fix
Either (easy fix): 

```cython
def __repr__(self):
    if self.sol == NULL:
        return "No valid solution"  # Or anything else
   ...
```
```cython
def getSolObjVal(self, Solution sol, original=True):
    if sol == None:
        sol = Solution.create(self._scip, NULL)
    if sol.sol == NULL:
        return None  # Or anything else
   ...
```

Or forbid creating `NULL` `Solution` altogether (always replace with `None`).

## Context
 - Linux
 - Python 3.9.4
 - PySCIPOpt 3.1.2 from conda-forge

mattmilten:
Good catch! I implemented a fix that just returns `None` instead of a Solution when there is no valid `SCIP_SOL*`.

AntoinePrv:
Thanks!

mattmilten:
This is not a valid fix.
A Solution with NULL pointer is used to get the current LP solution from a node. So it's a perfectly fine value for the Solution object.

CGraczyk:
What is the status here?

mattmilten:
This can be closed. The code will raise a warning now when used incorrectly, aka in the wrong state.

PR #497, opened 2021-04-24
proper fix for #493

mattmilten:
We need to allow Solutions with NULL sol pointer to be able to query LP solution values during the SOLVING stage.
This extends #494.

### DIFF
--- a/src/pyscipopt/scip.pyx
+++ b/src/pyscipopt/scip.pyx
@@ -541,13 +541,13 @@ cdef class Solution:
     def __getitem__(self, Expr expr):
         # fast track for Variable
         if isinstance(expr, Variable):
-            self._checkStage()
+            self._checkStage("SCIPgetSolVal")
             var = <Variable> expr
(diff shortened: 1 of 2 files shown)


src/pyscipopt/scip.pyx
```diff
@@ -569,6 +570,11 @@ cdef class Solution:
 
             vals[name] = SCIPgetSolVal(self.scip, self.sol, scip_var)
         return str(vals)
+    
+    def _checkStage(self):
+        if self.sol == NULL and not SCIPgetStage(self.scip) == SCIP_STAGE_SOLVING:
+            raise Warning(f"method cannot only be called in stage SOLVING without a valid solution")
```
matbesancon, line 576: For more informative error messages, we could pass the current method name to `_checkStage`,
`method X cannot be called` would be more helpful

mattmilten: agreed. Please see my updated code.

matbesancon (approved):

matbesancon:
looks good to me!

PR #498, opened 2021-05-07
Add method to model for setting the logfile

steffanschlein:
In addition to redirecting the solver's output to Python, it would be useful to save it to a file directly.


mattmilten:
Thanks for your contribution!

PR #499, opened 2021-05-07
Fix writing LP to file

steffanschlein:
Before, the conversion to a byte string was performed twice. This caused a Python error.


jurgen-lentz (approved):

mattmilten:
Looks good. Thanks!

Issue #430, opened 2020-09-15
Release the GIL on CPU intensive methods


AntoinePrv:
The Python [Global Interpreter Lock](https://wiki.python.org/moin/GlobalInterpreterLock) (GIL) only lets one thread execute Python code at the time.
So when using `threading` with pure Python, the code is actually concurrent, but not parallel.

When running non Python code, such as the C code of libscip, the GIL can be realsed, allowing `threading` to have some level of parallelism.

In Cython, this can be done by wrapping C function calls with:
```Cython
with nogil:
    SCIPsolve(self._scip)
```
However, one needs to make sure that no Python code is executed when releasing the GIL. This can be hard for two reasons:
  - Cython generates some Python code which can be hard to distinguish from C code;
  - The GIL needs to be reacquired in all Python callbacks such as BranchRule

fserra:
But scip itself is sequential. Maybe this would be helpful for concurrent scip, or for the parallelism that @stephenjmaher implemented.

AntoinePrv:
This is not about SCIP parallelism because Python would not care about C threads that do not run Python code.
This is about parallelism of user code, such as solving multiple Scip models in parallel (perhaps need `PARA_SCIP=ON`), or more generally doing any other non-SCIP computation while SCIP is solving

fserra:
ah ok, I see. Well, I don't know anything about this. I cannot help, sorry

AntoinePrv:
I can send a PR in the future. Just wanted to open an issue to keep track of it in the meantime.

AntoinePrv:
@CGraczyk, in case you are still interested in this feature, #505 is not a (full) fix for this.

CGraczyk:
I see, then i will reopen the issue, if you think this is nice to have feel free to open a new pull request.

AntoinePrv:
@axelvonkamp, as you noticed, releasing the GIL in PySCIPOpt is not extremely hard, simply adding a number of `nogil` around (the GIL is already being re-aquired during Python callbacks).
For it to get into PySCIPOpt, it's mostly a matter of doing and testing it properly, but I personally don't have much time for it.

As far as SCIP parallelism is involved, I'm not sure when it safe, I believe compiling SCIP with `-D PARASCIP=ON` (it's already the case if you get SCIP/PySCIPOpt from `conda`) should be enough if you use separate Models in each threads (as opposed to multiple threads modifying a single Model).

Joao-Dionisio:
[...]

PR #505, opened 2021-06-03
Ensure GIL during C callbacks

AntoinePrv:
During all C callbacks ensure that the GIL is held before executing Python code.

In Ecole we don't hold the GIL when calling `SCIPsolve`, so if users have added a callback through Python, user would get
```
Fatal Python error: PyThreadState_Get: no current thread
```

This is a prerequisite to  #430 , but I think it we should do a bit more testing before actually removing the GIL in intensive methods.

### DIFF
--- a/src/pyscipopt/benders.pxi
+++ b/src/pyscipopt/benders.pxi
@@ -70,10 +70,10 @@ cdef Variable getPyVar(SCIP_VAR* var):
     return <Variable>vardata
 
 
-cdef SCIP_RETCODE PyBendersCopy (SCIP* scip, SCIP_BENDERS* benders, SCIP_Bool threadsafe):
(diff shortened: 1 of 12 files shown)


gasse:
I've had segfaults when using Ecole + an event handler registered through PySCIPOpt, which this PR seems to fix. Any reason it hasn't been merged yet ?

Related discussion about this issue here:
https://github.com/ds4dm/ecole/discussions/193

CGraczyk:
@mattmilten do you have some insight here?

mattmilten:
I guess we can merge this. I don't see an issue and the tests pass ¯\_(ツ)_/¯

gasse:
Awesome. It seems safe indeed. And it would greatly improve the Ecole + PySCIPOpt compatibility :)

PR #506, opened 2021-06-03
Automatically download Cython

AntoinePrv:
Cython can be automatically downloaded with `setup_requires` for source builds.

Right now, I have some weird errors on `pip install .` but I think the changes are valid and the bug is with `setuptools`.
```
FileExistsError: [Errno 17] File exists: '.../PySCIPOpt/.eggs/Cython-0.29.23-py3.9-macosx-10.9-x86_64.egg'
```

To follow up in the future...

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,27 +41,18 @@ if "--debug" in sys.argv:
     extra_compile_args.append('-UNDEBUG')
     sys.argv.remove("--debug")
 
-use_cython = True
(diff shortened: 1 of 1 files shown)


mkoeppe:
`setup_requires` is an outdated mechanism. In modern Python packaging, one declares build-time dependencies using `pyproject.toml`. See #628

PR #508, opened 2021-06-07
Add convenience function to add multiple constraints at once

### LINKED ISSUE #507: Convenience functions to add multiple constraints
It would be useful to have a function in the `Model` class that allows adding multiple constraints at once using a generator. A possible function could be called `addConss()`. This would allow replacing code like
```python
for i in range(1, n_locations + 1):
    model.addCons(quicksum(x[i, j] for j in range(1, n_locations + 1)) == 1)
```
by
```python
model.addConss([quicksum(x[i, j] for j in range(n_locations)) == 1 for i in range(n_locations)])
```

The new function would use the default parameters of `addCons()` and return a list of added constraints. It would be easy to maintain because it is only a thin wrapper around the existing `addCons()`.

For example, Gurobi features a similar function [addConstrs()](https://www.gurobi.com/documentation/9.1/refman/py_model_addconstrs.html).

If this would fit the design principles of PySCIPOpt, I could create a pull request with my propased extension.

steffanschlein:
The added function by default behaves like `addCons()` but takes an iterable (i.e. a list) of constraints. In addition, it supports specifying the same parameters as for `addCons()` - either the same value for all constraints or a list with one value for each constraint.

Also adds a few somewhat elaborate tests for the added `addConss()`. 

A separate commit also fixes some inconsistancies with the existing `addCons()` documentation.

Resolves #507

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 ### Added
+- add convenience function `Model.addConss()` to add multiple constraints at once
 ### Fixed
 ### Changed
 ### Removed
--- a/src/pyscipopt/scip.pyx
+++ b/src/pyscipopt/scip.pyx
@@ -1979,18 +1979,19 @@ cdef class Model:
                 stickingatnode=False):
         """Add a linear or quadratic constraint.
 
-        :param cons: list of coefficients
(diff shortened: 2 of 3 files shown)


mattmilten:
Nice! Thank you for your contribution!

PR #511, opened 2021-06-12
Documentation of pyscipopt.Model.writeProblem()

dpaleka:
See https://stackoverflow.com/questions/67921357/filename-parsing-logic-in-pyscipopt.


mattmilten:
Thanks!

PR #513, opened 2021-06-16
add github actions

fschloesser:
We are transitioning from travis ci to github actions here. 

This is a heavy work in progress.


fschloesser:
We merge this now, because we need to have one of the workflow files on the default-branch (master) to be able to depend on it. The followup pull request is https://github.com/scipopt/PySCIPOpt/pull/514

PR #519, opened 2021-07-12
Add githubactions mac windows

fschloesser:
This is a work in progess, before the merge 
- for mac we need to solve the problem that pyscipopt finds the library. 
- for both mac and windows we might add a matrix for more python versions.

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -98,7 +98,7 @@ jobs:
           cd scipoptsuite-${{ env.version }}
           mkdir build
           cd build
-          cmake .. -DCMAKE_BUILD_TYPE=Debug -DCMAKE_INSTALL_PREFIX=${GITHUB_WORKSPACE}/scipoptsuite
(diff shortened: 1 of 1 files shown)


fschloesser:
So apparently there is something weird on mac with the linebreaks. the run commands are treated as one long command resulting in the error

fschloesser:
having fixed the prevous problem now there is a problem with the `py.test` command on mac.

fschloesser:
Turns out the problem is pytest itself on mac...

fschloesser:
The bottom line is that everything works out fine if one brew install tbb@2020, compiles the scipoptsuite and uses that.
In the same configuration the installer does not work. 
The next step is to compile an installer in debug mode and dig into what the problem is

fschloesser:
For now we will compile and install the scipoptsuite on the mac and use [caching](https://github.com/actions/cache) to speed up the process.

fschloesser:
The installation of scipoptsuite and pyscipopt now works. The fail is due to a failing pyscipopt test. 
Therefore i will sqash-merge this.

PR #520, opened 2021-07-23
Extend the API to access solution counters

gasse:
Just added those three getters:
 - getNSolsFound()
 - getNLimSolsFound()
 - getNBestSolsFound()

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,12 @@
 
 ## Unreleased
 ### Added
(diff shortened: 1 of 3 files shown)


CGraczyk:
Hi, have you added the SCIP functions to the scip.pxd file? This should be necessary for the functions to work.

Also please document the added functions in the CHANGELOG.md before the merge.

gasse:
Thanks @CGraczyk for the feedback.
Just did the fix (C declarations) and updated the changelog. Seems to compile and work after a quick test.
Let me know if anything else must be done.

Best

Issue #229, opened 2018-12-14
How to use Scip Parallel solve using Pyscipopt Model


LordSycho:
So, I have a CMIP that i am trying to solve using Pyscipopt( + I have installed SCIP Opt suite version 6). And as it is taking long and only using one core. I wanted to know is there a way that i can use Multiple cores using pyscipopt. As of now when i run it eats up 16gb ram but uses only one core. 

I tried to search for it and came to an understanding that it needs to be enabled while compiling the make file using TPI=tny and then a concurrent solve can be started by using the concurrentopt command instead of the optimize command in the SCIP shell

but when i try to replace optimize i get this error : 
AttributeError: 'pyscipopt.scip.Model' object has no attribute 'concurrentopt'

If it is possible please help me here. Or provide me some clarity over this. (Gurobi and IBM Cplex have these parallel mode functionality.)

LordSycho:
Hi,
Thanks for the above changes it makes pyscipopt try to run solveConcurrent but it at the end fails stating: 
"SCIP was compiled without task processing interface. Parallel solve not possible"

I am using windows installer any idea how to enable parallel solve on windows. 

As windows installer while installing doesn't provide any options at the time of installing to enable parallel mode.

mattmilten:
Interesting, we'll see what we can do about this and try to include the parallel mode in the Windows installer.

LordSycho:
Hi @mattmilten I am trying to optimize a MICP problem using scip. I observed 
1 . That scip 6 takes too long like more then an hour and still couldnt reach to optimal solution where 
     scip 5 was able to come with an optimal sol in 250 secs. 
2.  Scip 6 has some issues while installing on ubuntu with TPI=tny it raises issues with pthread. where 5 
     just gets installed with out any prob. 
3.  The issue that i am concerned about as of now : as i have started using scip 5: their is some problem 
     with DLL when running with pyscipopt.  I added the path and all but its not working. Do i need to 
     mention scip version in pyscipopt ?
****Thanks Please look in to 3rd And help asap.****

mattmilten:
Probably you need to clean up your system to avoid that PySCIPOpt links to the wrong SCIP library. I don't even understand whether you're using Windows or Linux right now.

Please understand that we are not some kind of support staff.

ravir252:
It had tried with fresh ubuntu 16.04 but it is not working

mattmilten:
Please follow the instructions **precisely**.

[...]

PR #535, opened 2021-10-29
SCIPsolverConcurrent implementation from issue #229

TNonet:
Most of this code was thanks to @fserra from issue #229.

Hopefully, this will help support Concurrent pySCIPOpt

### DIFF
--- a/src/pyscipopt/scip.pxd
+++ b/src/pyscipopt/scip.pxd
@@ -631,6 +631,7 @@ cdef extern from "scip/scip.h":
 
     # Solve Methods
     SCIP_RETCODE SCIPsolve(SCIP* scip)
+    SCIP_RETCODE SCIPsolveConcurrent(SCIP* scip)
     SCIP_RETCODE SCIPfreeTransform(SCIP* scip)
     SCIP_RETCODE SCIPpresolve(SCIP* scip)
 
--- a/src/pyscipopt/scip.pyx
+++ b/src/pyscipopt/scip.pyx
@@ -3093,6 +3093,12 @@ cdef class Model:
         PY_SCIP_CALL(SCIPsolve(self._scip))
         self._bestSol = Solution.create(self._scip, SCIPgetBestSol(self._scip))
 
+    def solveConcurrent(self):
+        """Transforms, presolves, and solves problem using additional solvers which emphasize on
+        finding solutions."""
+        PY_SCIP_CALL(SCIPsolveConcurrent(self._scip))
+        self._bestSol = Solution.create(self._scip, SCIPgetBestSol(self._scip))
+
     def presolve(self):
         """Presolve the problem."""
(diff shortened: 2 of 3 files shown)


mattmilten:
Another fortunate side effect of your contribution was that we realized that our CI actions did not run on PRs. So, this is also fixed now. You may have seen from the other PR #537 that concurrent optimization is only possible when compiling SCIP with a TPI (task processing interface) - this is not included in the default installer package.

TNonet:
Thanks for your help fixing that PR. 

Is there any guide on how to compile SCIP with a TPI? I have attempted to compile SCIP (and then install pySCIPOpt from local source with the compile flag). However, I haven't seen any documentation on the TPI aspect.

A second, potentially, more relevant question is: Is it feasible to create an extra with pySCIPOpt that includes a TPI? For example (using PIP instead of conda to more clearly show the [extra] aspect).
```python
pip install pySCIPOpt[TPI]
```

I would be happy to help develop this, but I will need some guidance.

mattmilten:
@CGraczyk can you assist here, please?

CGraczyk:
Regarding the installation of SCIP with TPI, there is a compile option (see bottom of the table [here](https://www.scipopt.org/doc/html/md_INSTALL.php) whereby tny stands for TinyCThread and omp for OpenMP.
(See also [this issue](http://listserv.zib.de/pipermail/scip/2018-April/003342.html))

If you have more trouble with SCIP with TPI feel free to ask in the SCIP mailing list (have a look at [this](https://scipopt.org/index.php#contact)).

Regarding the developement of an extra i am not sure, what would be the goal here when using this?

TNonet:
@CGraczyk Thank you for sharing those new resources. I will make sure I look into them.

The purpose of developing an extra, in my opinion, would be to take advantage of the concurrent and parallel solving capability of SCIP (https://www.scipopt.org/doc-5.0.1/html/CONCSCIP.php). However, I don't know if an extra is the appropriate method for this.

Specifically, I am looking into seeing if using the parallel/{max/min}nthreads parameter will help my MIP problem solve faster. I believe having the parallel and concurrent options available through an extra could be a way to make sure the current version of PySCIPOpt remains stable while still allowing for more "experimental" options for those interested.

Happy to discuss this further and figure out what (if any) way is appropriate.

jirkacigler:
Hello, I am encountering problem when calling model.solveConcurrent() leading to segmentation fault.  I was trying to analyse where the problem is, but no success. 
I use clean ubuntu 20.04 LTS, scip sources 7.0.3, building SCIP with arguments
cmake .. -DSYM=bliss -DWORHP=off -DZIMPL=off -DTPI=tny -DPARASCIP=on -DIPOPT=off -DPAPILO=off 
Do you have any advise, how to build SCIP so that I can use solveConcurrent()? thank you

PR #539, opened 2021-11-09
Add missing GIL statement

AntoinePrv:
Missed one in  #505.


AntoinePrv:
Some test failing due to TBB, I don't believe it's related to this change.

PR #545, opened 2021-11-18
import pre-installed scip

JPelda:
Changed the add_dll_directory to old os.environ['PATH']. add_dll_directory does not make pyscipopt use pre-installed scip versions.


mattmilten:
As you can see from the failing Windows tests, we cannot just remove/replace the DLL command.

PR #547, opened 2021-11-23
[Fixes #546] Delete stored pointer

### LINKED ISSUE #546: addVar / delVar / addVar
The following code:

```
model = pyscipopt.Model()
x = model.addVar('x')
model.delVar(x)
y = model.addVar('y')
```
fails with failed assertion in line 1464 in scip.pyx. The line is

```
         assert not pyVar.ptr() in self._modelvars
```

I don't see delVar removing old pointer from self._modelvars, may this be the issue?

CGraczyk:


mattmilten (approved):
Looks good to me 👍

PR #551, opened 2021-12-04
SCIP 8.0.0

jurgen-lentz:
I have merged all changes of the ```main``` branch into it and have adapted the integration tests to cope with SCIP 8.0.0.
I have increased the version of pyscipopt to ```3.5.0``` .

### DIFF
--- a/src/pyscipopt/__init__.py
+++ b/src/pyscipopt/__init__.py
@@ -1,4 +1,4 @@
-__version__ = '3.5.0'
+__version__ = '4.0.0'
 
 # required for Python 3.8 on Windows
 import os
--- a/src/pyscipopt/__init__.py
+++ b/src/pyscipopt/__init__.py
@@ -1,4 +1,4 @@
-__version__ = '3.3.0'
+__version__ = '3.5.0'
 
 # required for Python 3.8 on Windows
 import os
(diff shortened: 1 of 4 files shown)


src/pyscipopt/__init__.py
```diff
@@ -1,4 +1,4 @@
-__version__ = '3.3.0'
+__version__ = '3.5.0'
```
CGraczyk, line 1: This should be 4.0.0 since we are doing a major change to the core functionality of the API, will be changed before merging scip8 to master

jurgen-lentz: Now, I have changed it to 4.0.0 👍🏼

jurgen-lentz: Now, pyscipopt successfully builds on every os. But the solveConcurrent test fails on macOS with a seg fault. This test only fails on macOS because the linux and windows packages don't install the task processing interface.

CGraczyk: good to know, for the moment i have disabled the mac test and we can maybe fix it in the near future to reenable it.

jurgen-lentz: could you only disable py.test on MacOS (name "Run pyscipopt tests")? the rest works fine on MacOS

CGraczyk: sure good point

CGraczyk:
can you add the last commit of the partial mac integration test, then i can merge this pull request

jurgen-lentz:
done

PR #553, opened 2021-12-10
Scip 8

fschloesser:
Prepare the new release

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -83,58 +83,58 @@ jobs:
         run: |
           py.test
 
-#  MacOS-test:
-#    runs-on: macos-latest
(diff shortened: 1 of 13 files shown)


pfetsch:
We should also allow compilation with SCIP build with make, once ``config.h`` can be written with make.

In any case, we should do the following changes in ``setup.py``, because ``libscipsolver`` now is called ``libscip`` also for make:
```
# determine library
if os.path.exists(os.path.join(scipoptdir, 'lib/shared/libscip.so')):
    # SCIP seems to be created with make
    libdir = os.path.abspath(os.path.join(scipoptdir, 'lib/shared'))
    libname = 'scip'
    extra_compile_args.append('-DNO_CONFIG_HEADER')
else:
```

PR #558, opened 2022-01-28
add activeone param and getSlackVarIndicator()

pfetsch:
This pull request implements two features for indicator constraints:

- The `activeone` parameter of `addConsIndicator()` allows to activate the constraint if the binary (indicator) variable is 1 or 0.
- The function `getSlackVarIndicator()` returns the slack variable of the indicator constraint.

I have picked `scip-8` as a base, since this is what I have tested the code for. It should, however, also work for other versions.

### DIFF
--- a/src/pyscipopt/scip.pxd
+++ b/src/pyscipopt/scip.pxd
@@ -1657,6 +1657,8 @@ cdef extern from "scip/cons_indicator.h":
                                      SCIP_VAR* var,
                                      SCIP_Real val)
 
+    SCIP_VAR* SCIPgetSlackVarIndicator(SCIP_CONS* cons)
+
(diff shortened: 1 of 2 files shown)


CGraczyk:
so the master branch is already in v4 and uses scip8, but i guess the scip8 branch is behind master, can you merge into master instead? i will look into why the scip8 branch still exists in that form

fschloesser:
scip-8 was squash-merged already and should have been deleted. can you please rebase your branch on current master on your end? The interface does not seem to allow for this.

pfetsch:
O.k. rebasing resulted in a mess, so I see up a new pull request. Can we delete the branch `add_two_features_for_indicator_constraints`?

PR #561, opened 2022-02-04
Try to fix scip in place installation

pfetsch:
This pull request tries to allow a build for SCIP in place (both cmake and make).
The installation of SCIP installed in the system or somewhere else should be unchanged.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,6 +3,9 @@ import os, platform, sys, re
 
 # look for environment variable that specifies path to SCIP
 scipoptdir = os.environ.get('SCIPOPTDIR', '').strip('"')
+if not scipoptdir:
+    print('Environment variable SCIPOPTDIR is empty. Please specify path to SCIP.')
+    quit(1)
 
 extra_compile_args = []
 extra_link_args = []
--- a/setup.py
+++ b/setup.py
@@ -3,9 +3,6 @@ import os, platform, sys, re
 
 # look for environment variable that specifies path to SCIP
 scipoptdir = os.environ.get('SCIPOPTDIR', '').strip('"')
-if not scipoptdir:
-    print('Environment variable SCIPOPTDIR is empty. Please specify path to SCIP.')
-    quit(1)
 
 extra_compile_args = []
 extra_link_args = []
--- a/setup.py
+++ b/setup.py
@@ -7,42 +7,55 @@ scipoptdir = os.environ.get('SCIPOPTDIR', '').strip('"')
 extra_compile_args = []
 extra_link_args = []
 
-# determine include directory
-if os.path.exists(os.path.join(scipoptdir, 'src')):
-    # SCIP seems to be installed in place; check whether it was build using make or cmake
-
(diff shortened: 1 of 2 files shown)


pfetsch:
Hmm, it seems that my test that `SCIPOPTDIR` is empty is bad. I wonder how `setup.py` could work with this being empty. Maybe someone can explain?

setup.py
```diff
@@ -3,14 +3,27 @@
...
-    # SCIP seems to be installed in place
-    includedir = os.path.abspath(os.path.join(scipoptdir, 'src'))
+    # SCIP seems to be installed in place; check whether it was build using make or cmake
+
+    if os.path.exists(os.path.join(scipoptdir, 'src/scip')):
+        # SCIPOPTDIR pointed to the main source directory
+        includedir = os.path.abspath(os.path.join(scipoptdir, 'src'))
+    else:
+        # SCIPOPTDIR probably pointed to a cmake build directory; try one level up
+        if os.path.exists(os.path.join(scipoptdir, '../src')):
```
fschloesser, line 22: can we change this to 
```
if os.path.exists(os.path.join(scipoptdir, '../src/scip')):
```
just to be sure we are not messing with anything else?

setup.py
```diff
@@ -3,14 +3,27 @@
...
+            includedir = os.path.abspath(os.path.join(scipoptdir, '../src'))
+        else:
+            print('SCIPOPT contains a src directory, but we could not find the include files.')
```
fschloesser, line 25: Maybe better:
```
             print('Location SCIPOPTDIR does not contain a src/scip directory, exiting because we cannot not find the include files.')
```

setup.py
```diff
@@ -23,6 +36,8 @@
     libdir = os.path.abspath(os.path.join(scipoptdir, 'lib/shared'))
     libname = 'scip'
     extra_compile_args.append('-DNO_CONFIG_HEADER')
+    # the following is a temporary hack to make it compile with SCIP/make:
+    extra_compile_args.append('-DTPI_NONE') # this is just a guess, if other TPIs are used, please modify
```
fschloesser, line 40: If this is just a guess: can you please try if this works and paste the output? Thanks!

setup.py
pfetsch: Yes, thanks. Fixed.

setup.py
pfetsch: Yes. Fixed.

setup.py
pfetsch: This works. What kind of output do you want? I do not think that you see anything in the output ...

pfetsch:
Any idea why the tests fails?

setup.py
fschloesser: I just quoted your comment ;) 
If you checked that this worked it's fine

fschloesser:
Yes, the test fails due to your check if the SCIPOPTDIR variable is set. The test installs scipoptsuite globally.

pfetsch:
I have restructured the tests for the include directory and library. It should work now.

PR #562, opened 2022-02-05
clean up setup.py as best as possible

mattmilten:
using black formatting and some more pythonic string and path manipulations

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -93,7 +93,7 @@ ext = ".pyx" if use_cython else ".c"
 extensions = [
     Extension(
         "pyscipopt.scip",
-        [os.path.join(packagedir, "scip", "ext")],
+        [os.path.join(packagedir, "scip", ext)],
         include_dirs=[includedir],
         library_dirs=[libdir],
         libraries=[libname],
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -2,128 +2,128 @@
...
+        r'^__version__\s*=\s*[\'"]([^\'"]*)[\'"]', initfile.read(), re.MULTILINE
+    ).group(1)
 
 try:
     from Cython.Build import cythonize
 except ImportError as err:
     # if cython is not found _and_ src/pyscipopt/scip.c does not exist then we cannot do anything.
-    if not os.path.exists(os.path.join(packagedir, 'scip.c')):
-        print('Cython is required')
-        quit(1)
```
fschloesser, line 83: did you take this line out on purpose?

fschloesser (commented):
apart from the small question this looks good to me

setup.py
mattmilten: yeah, `sys.exit` is preferred over `quit` according to "the internet" and it has the added benefit of printing a message. So we save one line and have slightly cleaner code.

fschloesser: okay, i cannot find the sys.exit though

mattmilten: image

I just replaced single quotes with double quotes and added sys.exit.

fschloesser: ah yes, i was too close to the blackboard, all is good!

PR #566, opened 2022-03-01
Reader plugin

jurgen-lentz:
Hey all,

I have wrapped the reader plugin in `PySCIPOpt` and added a small test using the Soduko example.

Kind regards
Jurgen


mattmilten:
Looks good to me. @CGraczyk, can you review and merge, please?

Issue #571, opened 2022-03-08
cannot open shared object file


keithbriggs:
Two issues: first, the docs say that SCIPOPTDIR must point to a directory with this structure:

```
SCIPOPTDIR
  > lib
    > libscip.so ...
  > include
    > scip
    > lpi
    > nlpi
    > ...
```

However, scip-8.0.0 has nothing like this.   The only folders called `include` are:
```

scipoptsuite-8.0.0/scip/src/amplmp/include
scipoptsuite-8.0.0/scip/tests/include
```

So how is this supposed to work?  Second, despite this, when I set 

`export SCIPOPTDIR=/home/kbriggs/Downloads/scipoptsuite-8.0.0/scip/`

everything builds and installs correctly, but I still cannot import:

```
kbriggs@Z230-work:~/Downloads/PySCIPOpt-master> python3 setup.py build
...

kbriggs@Z230-work:~/Downloads/PySCIPOpt-master> python3 -m pip install  .
Processing /home/kbriggs/Downloads/PySCIPOpt-master
Collecting wheel (from PySCIPOpt==4.1.0)
  Using cached https://files.pythonhosted.org/packages/27/d6/003e593296a85fd6ed616ed962795b2f87709c3eee2bca4f6d0fe55c6d00/wheel-0.37.1-py2.py3-none-any.whl
Building wheels for collected packages: PySCIPOpt
  Running setup.py bdist_wheel for PySCIPOpt ... done
  Stored in directory: /home/kbriggs/.cache/pip/wheels/72/bc/7f/dc5033c1d6634cf37e8df4fc89f95919a978a3040b0e48fcac
Successfully built PySCIPOpt
Installing collected packages: wheel, PySCIPOpt
Successfully installed PySCIPOpt-4.1.0 wheel-0.37.1

kbriggs@Z230-work:~/Downloads/PySCIPOpt-master> python3
Python 3.6.9 (default, Dec  8 2021, 21:08:43) 
[GCC 8.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from pyscipopt import Model,scip
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/kbriggs/.local/lib/python3.6/site-packages/pyscipopt/__init__.py", line 11, in <module>
    from pyscipopt.scip      import Model
ImportError: libscip.so.8.0: cannot open shared object file: No such file or directory
```

What's going on here?

[...]

PR #573, opened 2022-03-10
enforce explicit specification of non-default SCIP installation

mattmilten:
This enforces the user to explicitly specify when a local CMake build or a make-style build of SCIP is supposed to be used for PySCIPOpt.

Use `python setup.py install --cmake` or `python setup.py install --make`, respectively.

This change is supposed to avoid incorrect path detections/assumptions as in #571.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,37 +15,19 @@ if not scipoptdir:
     print("Assuming that SCIP is installed globally.\n")
 
 else:
-    # determine include directory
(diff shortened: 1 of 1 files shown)


pfetsch:
I would rather refine the detection step and check for `src/scip`. Indeed this check is missing from `setup.py`.

pfetsch:
#575 contains my solution proposal.

PR #575, opened 2022-03-11
reorder search for installation place

pfetsch:
This pull request tries to improve the detection of include files in `setup.py`. First check for installation that contains an include directory. Then search for in-place installation. I also tried to make the error/output more precise.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,29 +12,29 @@ if not scipoptdir:
     includedir = "."
     libdir = "."
     libname = "libscip" if platform.system() in ["Windows"] else "scip"
-    print("Assuming that SCIP is installed globally.\n")
+    print("Assuming that SCIP is installed globally, because SCIPOPTDIR is undefined.\n")
 
 else:
-    # determine include directory
-    if os.path.exists(os.path.join(scipoptdir, "src")):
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -12,29 +12,29 @@
...
+    # check whether SCIP is installed in the given directory
+    if os.path.exists(os.path.join(scipoptdir, "include")):
+        includedir = os.path.abspath(os.path.join(scipoptdir, "include"))
     else:
-        # assume that SCIP is installed on the system
-        if os.path.exists(os.path.join(scipoptdir, "include")):
-            includedir = os.path.abspath(os.path.join(scipoptdir, "include"))
-        else:
-            sys.exit(f"SCIPOPTDIR={scipoptdir} does not contain an include directory.")
+        print("SCIPOPTDIR={scipoptdir} does not contain an include directory; searching for include files in src or ../src directory.\n")
```
mattmilten, line 23:
```suggestion
        print(f"SCIPOPTDIR={scipoptdir} does not contain an include directory; searching for include files in src or ../src directory.")
```

mattmilten (approved):
Looks good! And we don't need additional command line flags 👍

mattmilten:
fixes #571

PR #581, opened 2022-03-26
fix confusing error message in checkStage

### LINKED ISSUE #576: Cryptic error message
scip.pyx has:

` raise Warning(f"{method} cannot only be called in stage SOLVING without a valid solution (current stage: {SCIPgetStage(self.scip)})")`

I don't understand the grammar here - it makes no sense.  Is one of these the intended meaning:

` raise Warning(f"{method} cannot be called in stage SOLVING except with a valid solution (current stage: {SCIPgetStage(self.scip)})")`

or:

` raise Warning(f"{method} can only be called in stage SOLVING without a valid solution (current stage: {SCIPgetStage(self.scip)})")`

?   Or maybe something else?

mattmilten:
fix #576

### DIFF
--- a/src/pyscipopt/scip.pyx
+++ b/src/pyscipopt/scip.pyx
@@ -573,7 +573,7 @@ cdef class Solution:
     def _checkStage(self, method):
         if method in ["SCIPgetSolVal", "getSolObjVal"]:
             if self.sol == NULL and not SCIPgetStage(self.scip) == SCIP_STAGE_SOLVING:
-                raise Warning(f"{method} cannot only be called in stage SOLVING without a valid solution (current stage: {SCIPgetStage(self.scip)})")
+                raise Warning(f"{method} can only be called in stage SOLVING with a valid solution (current stage: {SCIPgetStage(self.scip)})")
 
 
 cdef class BoundChange:


Joao-Dionisio:
Hey, sorry for digging up an old merge. This is what master currently has:

```
if self.sol == NULL and not SCIPgetStage(self.scip) == SCIP_STAGE_SOLVING:
    raise Warning(f"{method} can only be called in stage SOLVING with a valid solution (current stage: {SCIPgetStage(self.scip)})")
```

But isn't this logically incorrect? The message is ambiguous, but I interpret it as saying that the method can only be called if we are in stage solving but with a valid solution. But the if condition allows for calling the method in the solving stage, even if we do not have a valid solution, right? Shouldn't the error message be:

```
if self.sol == NULL and not SCIPgetStage(self.scip) == SCIP_STAGE_SOLVING:
    raise Warning(f"{method} can only be called with a valid solution or in stage SOLVING (current stage: {SCIPgetStage(self.scip)})")
```

mattmilten:
I suppose the if condition just needs to be changed to an OR instead of the AND. Then, the warning is raised whenever there is no solution available or we are not in the right stage.

mattmilten:
Actually, this is correct because it should only guard against calling the two listed methods in that specific case.

mattmilten:
Hmm, I am somewhat confused now. You should also check https://github.com/scipopt/PySCIPOpt/issues/493 for som added context. Maybe the check is not correct for all cases. I you think the error message needs a different wording, feel free to improve it.

Joao-Dionisio:
Yeah, my issue was just with the error message, it wasn't the negation of the if condition when it must be. I think it should be

```
if self.sol == NULL and not SCIPgetStage(self.scip) == SCIP_STAGE_SOLVING:
    raise Warning(f"{method} can only be called with a valid solution or in stage SOLVING (current stage: {SCIPgetStage(self.scip)})")
```

instead. I'll get around to fixing it, eventually. I hadn't seen #493 , thanks! It seems to be related to #625 as well.

Issue #591, opened 2022-05-01
Unknown return code rowGetNNonz


stefanvanberkum:
Hi all,

For a cut selector plugin, I am trying to compute the integral support for cuts. For this purpose, given a cut (a row) I computed: 
```self.model.getRowNumIntCols(cut) / self.model.rowGetNNonz(cut)```

However, this yields the following error:
```Traceback (most recent call last):
  File "/home/stefan/.local/share/JetBrains/Toolbox/apps/PyCharm-C/ch-0/221.5080.212/plugins/python-ce/helpers/pydev/_pydevd_bundle/pydevd_exec2.py", line 3, in Exec
    exec(exp, global_vars, local_vars)
  File "<input>", line 1, in <module>
  File "src/pyscipopt/scip.pyx", line 1934, in pyscipopt.scip.Model.rowGetNNonz
  File "src/pyscipopt/scip.pyx", line 271, in pyscipopt.scip.PY_SCIP_CALL
Exception: SCIP: unknown return code!
```

So, the issue seems to be with the PY_SCIP_CALL of ```self.model.rowGetNNonz(cut)```, as its counterpart ```cut.getNNonz()``` does work as expected. While it does not create any problems for me, I figured I'd let you know. I am running PySCIPOpt 4.2.0 (installed via Conda as recommended) with Python 3.9.

Opt-Mucca:
Cheers for finding this error and using the cut selector! The error should be fixed by #592

stefanvanberkum:
Thank _you_ for including it, the plugin will be very useful for my research :)! I'll close the issue then.

PR #592, opened 2022-05-02
Removes incorrect PY_SCIP_CALL wrapper

Opt-Mucca:
This fixes the issue raised in #591

### DIFF
--- a/src/pyscipopt/scip.pyx
+++ b/src/pyscipopt/scip.pyx
@@ -1930,8 +1930,8 @@ cdef class Model:
         return SCIPgetRowNumIntCols(self._scip, row.scip_row)
 
     def rowGetNNonz(self, Row row):
-        """Gets number of non-zero etnries in the row"""
-        return PY_SCIP_CALL(SCIProwGetNNonz(row.scip_row))
+        """Gets number of non-zero entries in the row"""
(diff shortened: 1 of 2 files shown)


CGraczyk:
Please update the changelog before merge, thx!

src/pyscipopt/scip.pyx
```diff
@@ -1930,8 +1930,8 @@ cdef class Model:
         return SCIPgetRowNumIntCols(self._scip, row.scip_row)
 
     def rowGetNNonz(self, Row row):
-        """Gets number of non-zero etnries in the row"""
-        return PY_SCIP_CALL(SCIProwGetNNonz(row.scip_row))
+        """Gets number of non-zero entries in the row"""
+        return SCIProwGetNNonz(row.scip_row)
```
fserra, line 1934: I would say this function shouldn't exist. First, it doesn't use the scip pointer so it shouldn't be in class Model. Second, this function already is implemented in the class Row.

So maybe the proper fix is to just delete the function :)

Opt-Mucca: Fair point, was wondering if it was needed because of the duplicate. Nice to know about the SCIP pointer rule.

Opt-Mucca:
Added a line to the CHANGELOG and deleted the function following Felipe's advice

PR #598, opened 2022-06-07
Mt/row origin call

Opt-Mucca:
Allows users to determine the constraint type of a Row.


Opt-Mucca:
Also removes all the fclose calls that produce errors. (@mattmilten already seems to have removed them)

Issue #281, opened 2019-05-21
Column Generation example in PySCIPOpt


belzebuu:
Hi,

I would like to implement a column generation approach in PySCIPOpt for teaching purposes.

I have seen from tests/test_pricer.py that it is possible to use the callback function includePricer(). But I would like the students to work on a lower level to understand better
the technique.

I would like to follow the approach of: examples/unfinished/cutstock.py

However, I understood that relax() and column() are not available/implemented yet.

I could resort to reimplementing from scratch the restricted master problem every time a new variable is added. Thus no warm restart.

Is there another way I could mimic at this stage the implementation of cutstock.py. For example using addConstrCoeff()??

Thank you!
Best regards,
Marco

fserra:
sorry we never replied. is this still relevant?

belzebuu:
Yes 😊


--
Marco Chiarandini, Asc. Prof.
Department of Mathematics and Computer Science,
University of Southern Denmark, Campusvej 55,
DK-5230 Odense M, Denmark. Phone: +45 6550 4031
marco@imada.sdu.dk<mailto:marco@imada.sdu.dk>,  http://www.imada.sdu.dk/~marco

Joao-Dionisio:
Hey @belzebuu :) I don't know if the code is still useful for you, but I think I did what you wanted.

Here is the code: 
[cutstock.zip](https://github.com/scipopt/PySCIPOpt/files/9691010/cutstock.zip)

Let me know what you think!

belzebuu:
Dear @Joao-Dionisio,

yes! this is what I wanted. It will be useful for my teaching purposes. 

I have read the discussion around the workaround on line 48. Maybe if you publish this example under the finished ones, you could avoid redefining y below on line 73, as it maybe confusing.

I will then reconsider pyscipopt for the new term! 
Thanks a lot for keeping this in mind!

Joao-Dionisio:
I'm very glad I could help @belzebuu , I hope your students end up liking pyscipopt!

As for publishing the code, it is not in SCIP's philosophy to use it as a black box solver. The proper way would be to use a pricer as you first mentioned. If it interests you, you can read the discussion on [Pull Request #617 ](https://github.com/scipopt/PySCIPOpt/pull/617). Thank you for the recommendation on the y variables, though, you're absolutely right.

Good luck for the new term!

belzebuu:
[...]

PR #617, opened 2022-10-01
Fixed cutting stock example

Joao-Dionisio:
Fixed cutting_stock.py example that had the relax() and column() functions missing. Simplified the testing as well.


Joao-Dionisio:
Sorry for removing the pull-request and adding it again. I'm new to this Git business, and I wanted to do it properly, with local branches and whatnot. I think it's better now.

fserra:
Hey @Joao-Dionisio 

I have an issue with the example as it is. Although SCIP can, of course, be used as a black box solver, I believe that that's not SCIP's philosophy. The best example of that is the amount of work one needs to do in order to get the dual multipliers (although I guess one could just set up an LP problem and solve that instead of going through SCIP, see maybe `test_lp`

Anyway, it is my belief that the "SCIP way" of implementing the example, would be to write and add a pricer plugin (see `test_pricer`) and let all the column generation occur within the scip solving process. With the pricer, it should be trivial to obtain the dual multiplier, no need for strange workarounds or turning off presolve, propagation, etc.

The example would definitely be different from how it would be implemented in other solvers, but then again SCIP is a different type of solver and I guess it would be nice if the example would showcase this.

All that said, I am not involved in this repo that much anymore, so I guess the final word is @CGraczyk's.

Joao-Dionisio:
Thanks for the reply, @fserra ! I started looking at this example due to issue [#281](https://github.com/scipopt/PySCIPOpt/issues/281), where the commenter asks for a way for his students to better understand column generation, I'm guessing because using the pricer is too opaque. But you're right, of course, I'll just pass the code to him, that's probably the way to go. Thanks for the input!

fserra:
maybe the best is to use an LP for the pricing problem, because what #281 is asking for is very reasonable, and to be honest, I can't see how asking the students to disable presolve, add fixed variables to the bound, etc is any less obscure that just writing a pricing plugin.

Joao-Dionisio:
Sorry for the delay. When I get the time I will incorporate your suggestions @fserra , thanks for the input!

PR #623, opened 2022-11-06
Add ability to set priced variable score

mmghannam:


mattmilten (changes requested):
Looks good to me! Could you please update the CHANGELOG accordingly?
Thanks a lot!

mattmilten (approved):
Thanks!

PR #628, opened 2022-11-13
pyproject.toml: New

### LINKED ISSUE #626: Declare build dependencies in pyproject.toml
Building a source distribution using `python3 -m build --sdist .` fails because PySCIPOpt does not declare its PEP-518 build dependencies using `pyproject.toml`.
```
* Creating venv isolated environment...
* Installing packages in isolated environment... (setuptools >= 40.8.0, wheel)
* Getting build dependencies for sdist...
Assuming that SCIP is installed globally, because SCIPOPTDIR is undefined.
Cython is required.
```
https://github.com/mkoeppe/PySCIPOpt/actions/runs/3452468222/jobs/5762391556

mkoeppe:
Fixes #626

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,2 +1,2 @@
 [build-system]
-requires = ['setuptools', 'cython <3']
+requires = ['setuptools', 'cython <3, >=0.21']
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,2 @@
+[build-system]
+requires = ['setuptools', 'cython <3']
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -55,14 +55,6 @@ Building everything from source
 Recommended is to install in a virtual environment (e.g. `python3 -m venv <DIR_PATH>`).
 Please note that a globally installed version of PySCIPOpt on your machine might lead to problems.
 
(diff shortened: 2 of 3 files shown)


mkoeppe:
@mattmilten @wkschwartz ping

pyproject.toml
```diff
@@ -0,0 +1,2 @@
+[build-system]
+requires = ['setuptools', 'cython <3']
```
wkschwartz, line 2: Instructions you (correctly) deleted above said `cython >= .21`. If the `< 3` constraint is also correct (I don't know what pyscipopt needs so I have no opinion other than that the old instructions don't mention it) then you probably want
```suggestion
requires = ['setuptools', 'cython <3, >=0.21']
```

wkschwartz (commented):
I don't have enough familiarity with pyscipopt to opine on the list of requirements itself beyond what I see in this patch set

pyproject.toml
mkoeppe: Yes, adding this is also fine and correct; but setting this lower bound is not really needed – these build system requirements are resolved in their own venv, and there won't be a conflict with other packages that would force a lower cython version into this venv.

Cython < 3 protects against the untested major Cython release that has been in the making for many years.

mattmilten:
Thanks a lot!

PR #630, opened 2022-11-23
CI: Portability tests for building on various Linux distributions and macOS

mkoeppe:
Using the reusable workflows of the Sage CI. It builds papilo, soplex, scip, pyscipopt from source.

The same is already [in use in SCIP-SDP](https://github.com/scipopt/SCIP-SDP/blob/main/.github/workflows/ci-sage.yml)

Sample run: https://github.com/mkoeppe/PySCIPOpt/actions/runs/7969460775

### DIFF
--- a/.github/workflows/ci-sage.yml
+++ b/.github/workflows/ci-sage.yml
@@ -1,0 +1,105 @@
+name: Run Sage CI for Linux
+
+## This GitHub Actions workflow provides:
+##
+##  - portability testing, by building and testing this project on many platforms
+##    (Linux variants)
+##
+##  - continuous integration, by building and testing other software
+##    that depends on this project.
+##
+## It runs on every push to the GitHub repository.
+##
+## The testing can be monitored in the "Actions" tab of the GitHub repository.
+##
+## After all jobs have finished (or are canceled) and a short delay,
+## tar files of all logs are made available as "build artifacts".
+##
+## This GitHub Actions workflow uses the portability testing framework
+## of SageMath (https://www.sagemath.org/).  For more information, see
+## https://doc.sagemath.org/html/en/developer/portability_testing.html
+
+## The workflow consists of two jobs:
+##
+##  - First, it builds a source distribution of the project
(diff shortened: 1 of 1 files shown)


mkoeppe:
@mattmilten Any interest in this?

mattmilten:
This looks very impressive, but I don't think it's necessary to run such an extensive test suite on a regular basis - certainly not for every single push.

@CGraczyk, do you have an opinion on this?

mkoeppe:
> certainly not for every single push

I've changed it so that it only runs when a tag is pushed

matbesancon:
normally we should, as in, SCIP should compile on all distributions since there is not much changing from the SCIP point of view if the dependencies are available there

CGraczyk:
It would be good to have the logs for the action again, to see why certain distributions are failing.

mkoeppe:
> It would be good to have the logs for the action again, to see why certain distributions are failing.

Running at https://github.com/mkoeppe/PySCIPOpt/actions/runs/5195176263

mkoeppe:
Meanwhile Sage has switched to Cython 3, so we see build errors on all platforms: https://github.com/mkoeppe/PySCIPOpt/actions/runs/7867400154/job/21463042569#step:11:3964

Opt-Mucca:
@mkoeppe I'm not sure if this should be added to the repository. From my understanding the new CI ensures that pyscipopt is compatible with sage, and will continue working on those pipelines.
Shouldn't this be a responsibility on the sage repository? We want to the pipeline to work of course, and we are against changes that would break it without good reason, but we already have support for tests over multiple distributions. Our build process is also now finally up and running (in part thanks to your original suggestion). So I would be in favour of closing this merge request unless I've missed something.

Joao-Dionisio:
I'm out of my depth with this, but it's very old and a lot of things have happened since it was opened. Unless there are any objections and counter-arguments to Mark's points, I'm closing it and redirecting the PR to [scipoptsuite-deploy](https://github.com/scipopt/scipoptsuite-deploy/issues/24).

PR #649, opened 2023-02-18
.github/workflows/integration-test.yml: Update tested Python versions

mkoeppe:
... all the way to 3.12-dev

@mattmilten


mattmilten:
It seems like Python 3.12 may be too ambitious for now. Looks like a problem coming from cython.

Thanks for updating the tests! This was long overdue!

mattmilten:
I don't like turning off fail-fast, though. This could save a bunch of computing resources 🍃

mattmilten:
Thanks!

PR #652, opened 2023-02-20
Add GH Actions workflow to build sdists and wheels, deploy to PyPI

mkoeppe:
wheel building via cibuildwheel, using Sage for installing SCIP from source.

Preview: https://github.com/mkoeppe/PySCIPOpt/actions/runs/4220475298

To deploy to PyPI, it will need an Actions secret `PYPI_API_TOKEN`

### DIFF
--- a/.github/workflows/cibuildwheel.yml
+++ b/.github/workflows/cibuildwheel.yml
@@ -9,10 +9,10 @@ on:
     types:
       - published
 
-concurrency:
-  # Cancel previous runs of this workflow for the same branch
-  group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+# concurrency:
+#   # Cancel previous runs of this workflow for the same branch
+#   group: ${{ github.workflow }}-${{ github.ref }}
+#   cancel-in-progress: true
 
 jobs:
   build_wheels:
--- a/.github/workflows/cibuildwheel.yml
+++ b/.github/workflows/cibuildwheel.yml
@@ -21,7 +21,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest]
+        os: [ubuntu-latest, macos-latest]
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


mattmilten:
@CGraczyk, could you please check how we can integrate this with the existing release workflow?

mkoeppe:
@mattmilten more infrastructure

matbesancon:
ping @CGraczyk

matbesancon:
thanks a lot for the PR! This should really be added

mkoeppe:
Wheels.

CGraczyk:
Okay, so we would replace the old release action with this without having changes for the users on new versions - since the PYPI secret is already setup, this should be easy to do. 

It would be best to test this workflow from a branch first - this should be possible if i remember correctly from back when we first setup the github actions @fschloesser ?

jni:
Hi folks! I'm *actually* looking for information about macos-ARM/Apple Silicon builds — are there fundamental reasons why those don't exist (I notice this PR specifies x86 for the macos build)? Or it just hasn't been tried yet? 😬 

re

> To deploy to PyPI, it will need an Actions secret PYPI_API_TOKEN

I just wanted to point out the [trusted-publishers mechanism](https://docs.pypi.org/trusted-publishers/) on PyPI, which is super super nice and avoids the use of the API token. It just needs someone with maintainer access to the PyPI package to specify the build script here.

Thanks! 🙏

mkoeppe:
Until recently, nobody had macOS arm64 runners, so any builds for that platform would have to be cross. Since early this year, GH Actions provides such runners though, so it has become easier

jni:
> any builds for that platform would have to be cross.

Sure, but cibuildwheel got those working which has been very valuable for many projects for some time now. At any rate, as you point out, now is a good time to add them, but I guess first getting this PR merged and then adding ARM wheels would be a good plan of attack. Hopefully this can get merged soon. 🙏 Thanks!

mkoeppe:
> cibuildwheel got those working which has been very valuable for many projects for some time now

Of course, but the CIBW_BEFORE_ALL script (where you build any library deps) needs to support cross compilation in that case.

mkoeppe:
No longer needed.

PR #667, opened 2023-03-30
Add numerics and benchmark paramemphasis options

erdos4d:
@Joao-Dionisio @mattmilten Here are the changes, I hope this is correct. Please let me know if there is anything else needed. Thanks!


Joao-Dionisio:
Hey, @mattmilten do you think this looks good? I can't (and rightfully so) accept pull requests yet.

PR #693, opened 2023-06-18
Raise error when pricer fundamental callbacks are not implemented

mmghannam:


mmghannam:
I wanted initially to use the `@abstractmethod` decorator but apparently it's not supported by _Cython_. If this is okay I can also do the same for all fundamental callbacks in plugins.

ambros-gleixner (approved):

mattmilten:
Thanks for adding this!

PR #694, opened 2023-06-18
Add code coverage to CI

mmghannam:


mmghannam:
I moved the contents of `scip.pyx` to `scip.pxi` file and included it since the coverage tool was complaining about parsing `.pyx` files, but it had no trouble with `.pxi` files. I hope this is okay.

mattmilten:
Hmm, I am not sure about renaming scip.pyx to scip.pxi.
@CGraczyk what do you say?

Having code coverage info is super nice, though, so I suppose it's fine to rename...

PR #716, opened 2023-08-07
Update test_branch_probing_lp to avoid failure due to root node solving

sbolusani:
The test `test_branch_probing_lp.py` fails with SCIP's master branch because it gets solved to optimality in the root node itself. As a workaround, I switched off the heuristics to allow the solution process to enter the branching phase.

It passes fine now on the z1 login node. Do we need to increase the time limit just to be sure of other architectures?

Feel free to suggest alternative changes to fix this failure.

### DIFF
--- a/tests/test_branch_probing_lp.py
+++ b/tests/test_branch_probing_lp.py
@@ -45,7 +45,7 @@ def test_branching():
     m.setHeuristics(SCIP_PARAMSETTING.OFF)
     m.setIntParam("presolving/maxrounds", 0)
     #m.setLongintParam("lp/rootiterlim", 3)
-    m.setRealParam("limits/time", 60)
+    m.setLongintParam("limits/nodes", 1)
 
     x0 = m.addVar(lb=-2, ub=4)
     r1 = m.addVar()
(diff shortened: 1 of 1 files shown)


tests/test_branch_probing_lp.py
```diff
@@ -42,6 +42,7 @@ def branchexeclp(self, allowaddcons):
 
 def test_branching():
     m = Model()
+    m.setHeuristics(SCIP_PARAMSETTING.OFF)
     m.setIntParam("presolving/maxrounds", 0)
     #m.setLongintParam("lp/rootiterlim", 3)
     m.setRealParam("limits/time", 60)
```
mmghannam, line 48: Since the code only works when it's called the first time, how about we make this node limit instead of 2? that should be system agnostic and would probably be faster. 
```suggestion
    m.setLongintParam("limits/nodes", 2)
```

sbolusani: Yes, it sounds good. I made a similar change but with "limits/node" as 1 instead of 2. This parameter corresponds to the number of processed nodes. So the node limit of 1 works as intended.

PR #723, opened 2023-09-25
Remove warning about SCIP licensing

### LINKED ISSUE #699: scary warning about licensing can be removed as of SCIP 8.0.3
SCIP is now licensed Apache 2.0, so telling readers to check licensing conditions (while technically correct) is probably unneeded.

Joao-Dionisio:
I think this is what @bremner was talking about with #699. Since SCIP is now under the Apache 2.0 license, a warning may not be necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,7 @@
 PySCIPOpt
 =========
 
-This project provides an interface from Python to the [SCIP Optimization
-Suite](https://www.scipopt.org/). Please review [SCIP's license restrictions](https://scipopt.org/index.php#license) before installing PySCIPOpt.
+This project provides an interface from Python to the [SCIP Optimization Suite](https://www.scipopt.org/). 
 
 [![Gitter](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/PySCIPOpt/Lobby)
 [![PySCIPOpt on PyPI](https://img.shields.io/pypi/v/pyscipopt.svg)](https://pypi.python.org/pypi/pyscipopt)
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 PySCIPOpt
 =========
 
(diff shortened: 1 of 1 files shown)


bremner:
João Dionísio ***@***.***> writes:

> I think this is what @bremner was talking about. 
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/scipopt/PySCIPOpt/pull/723
>
> -- Commit Summary --

Looks reasonable to me

d

mmghannam:
Since the user is allowed to link against any SCIP installation, I think the warning is still useful, in case they use an earlier version of SCIP. But maybe a mention of the new Apache 2.0 license would be helpful. What do you think?

bremner:
Mohammed Ghannam ***@***.***> writes:

> Since the user is allowed to link against any SCIP installation, I think the warning is still useful, in case they use an earlier version of SCIP. But maybe a mention of the new Apache 2.0 license would be helpful. What do you think? 
>

If you want to do that, I'd suggest being specific about problematic
SCIP versions. Personally, I think the warning is no longer very useful,
since anyone who finds the old license conditions problematic can just
upgrade.

Joao-Dionisio:
> Starting from v8.0.3, SCIP uses the [Apache2.0](https://www.apache.org/licenses/LICENSE-2.0) license. If you plan to use an earlier version of SCIP, please review [SCIP's license restrictions](https://scipopt.org/index.php#license).

What about this?

bremner:
João Dionísio ***@***.***> writes:

>> Starting from v8.0.3, SCIP uses the [Apache2.0](https://www.apache.org/licenses/LICENSE-2.0) license. If you plan to use an earlier version of SCIP, please review [SCIP's license restrictions](https://scipopt.org/index.php#license).
>
> What about this?
>

Sure, it's definitely an improvement.

d

PR #733, opened 2023-10-14
Fix solution print test, don't depend on sols order

mmghannam:


Joao-Dionisio (commented):
Seems good, Mo! But what do you think in regards to tests, should we test the methods with more complex problems, or with these trivial problems it's fine?

mmghannam:
I'm not sure if a more complex model would test the behavior differently, I will merge it as is for now. If you think there's some behavior uncovered let me know and I'd be happy to modify the test.

PR #743, opened 2023-10-28
Add hasPrimalRay, getPrimalRay, getPrimalRayVal

Joao-Dionisio:


src/pyscipopt/scip.pxi
```diff
@@ -4621,6 +4621,31 @@ cdef class Model:
...
+        """
+        Gets value of given variable in primal ray causing unboundedness of the LP relaxation
+        """
+        return SCIPgetPrimalRayVal(self._scip, var.scip_var)
+
+    def getPrimalRay(self):
+        """
+        Gets primal ray causing unboundedness of the LP relaxation
+        """
+        cdef int _nvars = SCIPgetNVars(self._scip)
```
mmghannam, line 4641: We should probably check here if a primal ray exists, to avoid a segmentation fault. In case it doesn't exist we should throw an error that explains it.

tests/test_solution.py
```diff
@@ -86,5 +127,4 @@ def test_getSols():
     assert any(sol[x] == 0.0 for sol in m.getSols())
 
 if __name__ == "__main__":
```
mmghannam, line 129: Let's remove this unneeded main too 😄

src/pyscipopt/scip.pxi
```diff
@@ -4621,6 +4621,31 @@ cdef class Model:
...
+        """
+        Returns whether a primal ray is stored that proves unboundedness of the LP relaxation
+        """
+        return SCIPhasPrimalRay(self._scip)
+        
+    def getPrimalRayVal(self, Variable var):
+        """
+        Gets value of given variable in primal ray causing unboundedness of the LP relaxation
+        """
+        return SCIPgetPrimalRayVal(self._scip, var.scip_var)
```
mmghannam, line 4635: Same comment as for the getPrimalRay method.

mmghannam (changes requested):
Thanks João! I added some minor comments.

src/pyscipopt/scip.pxi
Joao-Dionisio: Ah, nice one, Mo! Added it now.

tests/test_solution.py
Joao-Dionisio: I removed all of them on the other PR, that should be okay, right?

mmghannam: Yup :)

PR #744, opened 2023-11-03
Add some tests. Remove unnecessary if __name__ == __main__

Joao-Dionisio:


tests/test_pricer.py
```diff
@@ -10,6 +10,7 @@ def pricerredcost(self):
         dualSolutions = []
         for i, c in enumerate(self.data['cons']):
             dualSolutions.append(self.model.getDualsolLinear(c))
+            #assert self.model.getDualsolLinear(c) == self.model.getDualSolVal(c)
```
mmghannam, line 13: Why is this commented out?

tests/test_reader.py
```diff
@@ -59,7 +59,7 @@ def createFile(filename):
 def deleteFile(filename):
     os.remove(filename)
 
-def test():
+def test_main():
```
mmghannam, line 62: Shouldn't this be something more descriptive, like `test_sudoko_reader` for example.

tests/test_solution.py
```diff
@@ -62,6 +62,22 @@ def test_solution_evaluation():
...
+        x[i] = m.addVar(ub=i)
+
+    for i in range(1,6):
+        m.addCons(quicksum(x[j] for j in range(20) if j%i==0) >= i)
+        m.addCons(quickprod(x[j] for j in range(20) if j%i==0) <= i**3)
+    
+    m.setObjective(quicksum(x[i] for i in range(20)))
+    m.optimize()
+    for s in m.getSols():
+        m.getSolTime(s)
```
mmghannam, line 80: There should be an assert here, to at least check that it's a positive value, and potentially check that each one has a time after the other, I'm assuming they are returned in the order they were found but I could be wrong here.

mmghannam (changes requested):
Thanks a lot @Joao-Dionisio!

tests/test_pricer.py
Joao-Dionisio: I forgot to delete it, but the assertion fails, ```getDualSolVal``` returns 0, even with ```boundconstraint=True```. 

As a side not, it's annoying that sol is not capitalized in ```getDualsolLinear````

tests/test_reader.py
Joao-Dionisio: You're right, Mo! I also noticed that there were many different spellings of sudoku in this file, I fixed those typos :)

tests/test_model.py
```diff
@@ -170,6 +172,38 @@ def assert_conss_neq(a, b):
 
     assert_conss_neq(conss[0], conss[1])
 
+def test_addCoefLinear():
+    m = Model()
+    x = m.addVar(obj=1)
+    y = m.addVar()
```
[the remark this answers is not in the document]
Joao-Dionisio: Done!

tests/test_solution.py
Joao-Dionisio: Yeah, agreed. But unfortunately, it seems that ```getSols``` doesn't return solutions in the order they were found.

tests/test_solution.py
```diff
@@ -79,6 +79,47 @@ def test_getSolTime():
...
+    x = m.addVar()
+    m.setObjective(x, "maximize")
+    m.setPresolve(SCIP_PARAMSETTING.OFF)
+    
+    m.optimize()
+    
+    assert m.hasPrimalRay()
+
+    m = Model()
+    x = m.addVar(lb = 0) # for readability
```
mmghannam, line 93: no need to add a comment here 😄 
```suggestion
    x = m.addVar(lb = 0)
```

tests/test_pricer.py
mmghannam:
```suggestion
```
Let's remove the line then for now. The naming comes from the corresponding SCIP C function, better to be consistent even if not the best naming.

mmghannam: I opened a new issue #746  to track this.

mmghannam (approved):

PR #751, opened 2023-11-16
Expand testing

Joao-Dionisio:
Some notes and questions.

There is no method to check whether a solution has been freed. If we try to ```setSolVal``` in a freed solution, we get a segfault. 

```addConsLocal``` appears to still be causing issues.

```addConsIndicator``` appears to still be causing issues when the binary variable is $0$. Should we add tests that we know will fail because the methods are wrong?

I think logical constraints were already tested in ```test_logical.py```, but the syntax is completely different from the rest of the tests. 

Should we treat the tests as a potential tutorial? I know I used ```test_pricer.py``` to know how a pricer works, but I've now added some methods there (```getNSolsFound```, ```getNBestSolsFound```, ```getNSols```) just to test them, and I worry that new users would find it weird. In my opinion, we should greatly expand the tutorial folder, and assume that the tests will not be accessed by regular users.

### DIFF
--- a/tests/test_cons.py
+++ b/tests/test_cons.py
@@ -111,3 +111,11 @@ def test_cons_indicator():
     assert m.getVal(slack) == 0
     assert m.getVal(binvar) == 1
     assert m.getVal(x) == 1
+
+def test_printCons():
+    m = Model()
+    x = m.addVar()
+    y = m.addVar()
+    c = m.addCons(x*y <= 5)
+    
+    m.printCons(c)
--- a/tests/test_cons.py
+++ b/tests/test_cons.py
@@ -114,7 +114,7 @@ def test_cons_indicator():
     assert m.isEQ(m.getVal(x), 1)
 
 
-@pytest.mark.xfail
+@pytest.mark.xfail(reason="addConsIndicator doesn't behave as expected when binary variable is False. See Issue #717."
 def test_cons_indicator_fail():
     m = Model()
     binvar = m.addVar(vtype="B")
--- a/tests/test_cons.py
+++ b/tests/test_cons.py
@@ -155,6 +155,6 @@ def test_getValsLinear():
     assert True
 
 
-@pytest.mark.skip(reaons="TODO: test getRowLinear()")
+@pytest.mark.skip(reason="TODO: test getRowLinear()")
 def test_getRowLinear():
     assert True
--- a/tests/test_cons.py
+++ b/tests/test_cons.py
@@ -114,7 +114,7 @@ def test_cons_indicator():
     assert m.isEQ(m.getVal(x), 1)
 
 
-@pytest.mark.xfail(reason="addConsIndicator doesn't behave as expected when binary variable is False. See Issue #717."
+@pytest.mark.xfail(reason="addConsIndicator doesn't behave as expected when binary variable is False. See Issue #717.")
 def test_cons_indicator_fail():
     m = Model()
     binvar = m.addVar(vtype="B")
@@ -150,11 +150,11 @@ def test_printCons():
     m.printCons(c)
 
 
-@pytest.mark.skip
+@pytest.mark.skip(reason="TODO: test getValsLinear()")
 def test_getValsLinear():
     assert True
 
 
-@pytest.mark.skip
+@pytest.mark.skip(reaons="TODO: test getRowLinear()")
 def test_getRowLinear():
     assert True
--- a/tests/test_cons.py
+++ b/tests/test_cons.py
@@ -40,3 +40,54 @@ def test_constraint_option_setting():
         assert c.isEnforced() == option
         assert c.isRemovable() == option
         assert c.isInitial() == option
+
+def test_cons_logical():
+    m = Model()
+    x1 = m.addVar(vtype="B")
+    x2 = m.addVar(vtype="B")
+    x3 = m.addVar(vtype="B")
+    x4 = m.addVar(vtype="B")
+    result1 = m.addVar(vtype="B")
+    result2 = m.addVar(vtype="B")
+
+    m.addCons(x3 == 1-x1)
+    m.addCons(x4 == 1-x2)
+    
+    # result1 true
+    m.addConsAnd([x1,x2], result1)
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: tests/test_cons.py [hunk 1]
```diff
@@ -39,4 +39,94 @@ def test_constraint_option_setting():
...
+    m.addConsXor([x1,x3], True)
+
+    # result2 false
+    m.addConsOr([x3,x4], result2)
+    m.addConsAnd([x1,x3], result2)
+    m.addConsXor([x1,x2], False)
+
+    m.optimize() 
+
+    assert m.getVal(result1) == 1
```
mmghannam, line 68: Should be compared with `isEQ`

tests/test_cons.py
```diff
@@ -39,4 +39,94 @@ def test_constraint_option_setting():
...
+
+    m.addVarSOS1(c1, x[2], 1)
+    m.addVarSOS2(c2, x[3], 1)
+
+    m.appendVarSOS1(c1, x[4])
+    m.appendVarSOS2(c2, x[5])
+
+    m.optimize()
+
+    assert m.getVal(x[0]) == 0
```
mmghannam, line 88: same here

tests/test_cons.py
```diff
@@ -39,4 +39,94 @@ def test_constraint_option_setting():
...
+    assert m.getVal(x[1]) == 0
+    assert m.getVal(x[2]) == 0
+    assert m.getVal(x[3]) == 1
+    assert m.getVal(x[4]) == 1
+    assert m.getVal(x[5]) == 1
+
+
+# WIP. addConsIndicator doesn't behave as expected when binary variable is False. See Issue #717.
```
mmghannam, line 96: Maybe change it to an xfail test.

tests/test_solution.py
```diff
@@ -32,9 +32,19 @@ def test_solution_create():
 
     s = m.createSol()
     s[x] = 2.0
-    s[y] = 4.0
+    s[y] = 5.0
+    assert not m.checkSol(s)
     assert m.addSol(s, free=True)
 
+    s1 = m.createSol()
+    m.setSolVal(s1,x,1.0)
```
mmghannam, line 40: Missing spaces, use an auto-formatting tool 😄

tests/test_solution.py
```diff
@@ -32,9 +32,19 @@ def test_solution_create():
...
+    m.setSolVal(s1,y,2.0)
+    assert m.checkSol(s1)
+
+    m.optimize()
+
+    assert m.getSolObjVal(s1) == -1
```
mmghannam, line 46: With `isEQ` please

tests/test_solution.py
```diff
@@ -59,9 +69,13 @@ def test_solution_evaluation():
 
     # Check consistency with Models's getVal method
     assert sol[x] == m.getVal(x)
+    assert m.getSolVal(sol,x) == m.getVal(x)
```
mmghannam, line 72: here too please and also formatting.

tests/test_cons.py
Joao-Dionisio: Oh yeah, thanks Mo!

tests/test_cons.py
Joao-Dionisio: Didn't know about this, thanks for letting me know!

tests/test_cons.py
```diff
@@ -39,4 +42,120 @@ def test_constraint_option_setting():
...
+
+    m.optimize()
+
+    assert m.isEQ(m.getVal(slack), 0)
+    assert m.isEQ(m.getVal(binvar), 1)
+    assert m.isEQ(m.getVal(x), 1)
+
+
+# addConsIndicator doesn't behave as expected when binary variable is False. See Issue #717.
+@pytest.mark.xfail
```
mmghannam, line 118:
```suggestion
@pytest.mark.xfail(reason="addConsIndicator doesn't behave as expected when binary variable is False. See Issue #717."
```

tests/test_cons.py
```diff
@@ -39,4 +42,120 @@ def test_constraint_option_setting():
...
+    slack = m.getSlackVarIndicator(c)
+
+    m.optimize()
+
+    assert m.isEQ(m.getVal(slack), 0)
+    assert m.isEQ(m.getVal(binvar), 1)
+    assert m.isEQ(m.getVal(x), 1)
+
+
+# addConsIndicator doesn't behave as expected when binary variable is False. See Issue #717.
```
mmghannam, line 117:
```suggestion
```

tests/test_cons.py
```diff
@@ -39,4 +42,119 @@ def test_constraint_option_setting():
...
+    slack = m.getSlackVarIndicator(c)
+
+    m.optimize()
+
+    assert m.isEQ(m.getVal(slack), 0)
+    assert m.isEQ(m.getVal(binvar), 1)
+    assert m.isEQ(m.getVal(x), 1)
+
+
+@pytest.mark.xfail(reason="addConsIndicator doesn't behave as expected when binary variable is False. See Issue #717."
```
mmghannam, line 117:
```suggestion
@pytest.mark.xfail(reason="addConsIndicator doesn't behave as expected when binary variable is False. See Issue #717.")
```

tests/test_cons.py
```diff
@@ -39,4 +42,119 @@ def test_constraint_option_setting():
...
+def test_printCons():
+    m = Model()
+    x = m.addVar()
+    y = m.addVar()
+    c = m.addCons(x * y <= 5)
+
+    m.printCons(c)
+
+
+@pytest.mark.skip
```
mmghannam, line 153:
```suggestion
@pytest.mark.skip(reason="TODO: test getValsLinear()")
```

tests/test_cons.py
```diff
@@ -39,4 +42,119 @@ def test_constraint_option_setting():
...
+def test_getValsLinear():
+    assert True
+
+
+@pytest.mark.skip
```
mmghannam, line 158:
```suggestion
@pytest.mark.skip(reaons="TODO: test getRowLinear()")
```

mmghannam (approved):

Issue #755, opened 2023-11-24
Cannot addVar after solving and freeTransform with pricedVars


igorbanfi:
Hello

**Describe the bug**

After solving the model with pricer and running freeTransform, pricedVars created during solution process are deleted from scip instance, but _modelvars is not updated. When trying to add a new variable to the model with addVar, assertion error is thrown

```
  File "***/example.py", line 162, in <module>
    s.addVar("x", vtype = "C", obj = 1.0)
  File "src/pyscipopt/scip.pyx", line 1468, in pyscipopt.scip.Model.addVar
AssertionError
```
In provided example 4 variables are added without an issue, but when adding the fifth the exception is raised.

**To Reproduce**
[Example](https://gist.github.com/igorbanfi/b143ffde19f6fa06048d40a483abd034)
Derived from _tests/test_pricer.py_ 

**System**
 - OS: Ubuntu 22.04
 - Version 4.3.0
 - SCIP version 8.0.4

igorbanfi:
A solution to the issue is to recreate _modelvars after freeTransform
323212ed5519da3480e7f642091ed84854d28cfb

Joao-Dionisio:
Hey @igorbanfi! Thanks for the detailed issue and the PR. I've been trying to go over what's going wrong, and indeed I think it's trying to access ```self_modelvars``` after free transforming. Someone else may accept your PR (I can't). Maybe this is something that could be accomplished without deleting and getting the variables again.

But, when it comes to fixing the issue it works great, and it may clear other open issues, so that's awesome!

igorbanfi:
Hi. I agree it would be better if _modelvars is modified before running freeTransform. Opened a PR.

Joao-Dionisio:
PR #757 solves this issue. Thanks for the help, Igor!

Issue #739, opened 2023-10-22
Increase test-coverage


Joao-Dionisio:
At the time of writing, PySCIPOpt's test coverage is around $58.4\\%$. Increasing the test coverage is not too difficult and would improve the project.

### How to write a new test

1. Go to the code coverage [website](https://app.codecov.io/gh/scipopt/pyscipopt/tree/master/src%2Fpyscipopt) and find a function that is not covered by tests.
2. Locate the python file in the ```tests``` folder that is better suited for this method. If you feel one doesn't exist, you may create a new file. **Make sure it starts with ```test_```**
3. Create a new function (**its name must start with ```test_```**) and write a script that calls the method, finishing the script with an assertion that the result of the method must equal the expected result.
4. Run the new test with ```pytest test_XYZ.py```. Ensure you're in the right directory and have pytest on your machine.

#### Progress 
 
| Date | Code Coverage ($\\%$)|
| ------ | ------------------------------- |
| 23.10 | $58.4$ |
| 06.11| $59.6$|
| 17.11| $62$|

PR #757, opened 2023-11-27
Update Model._modelvars before freeTransfrom

igorbanfi:
Before running freeTransform nonOriginal variables are removed from _modelvars.

fix for https://github.com/scipopt/PySCIPOpt/issues/755


mmghannam:
Hi @igorbanfi, thank you for your contribution! Would be great if you could also add a test with the code that was failing in #755.

Joao-Dionisio:
For help with writing a test, see Issue #739, @igorbanfi. If it's too much of a hassle, we can also do it ourselves, of course, but it's nicer if it's just one author :)

mmghannam (approved):
Looks good, thanks a lot @igorbanfi!

PR #758, opened 2023-11-27
Type loop variables

Joao-Dionisio:
Added ```cdef int``` to looping variables. Should speed PySCIPOpt a little bit. I don't think integer overflow will be a problem, but I'm a newbie at C.

Still need to add other types (sometimes we iterate through variables, for example).


mmghannam (approved):
Thanks João!

PR #759, opened 2023-11-28
Fix wrong numbers for Event Types

Joao-Dionisio:
An important issue came to our attention. PySCIPOPt has been using very outdated identifiers for event types. This draft PR is an attempt to correct this.

I think we should do the same for all other types, not relying on the use of magic numbers. 

The way I'm doing this is probably not the best, as defining a new SCIP_EVENTTYPE inside PySCIPOPt is quite convenient. 

Also, almost none of these events are tested. This code passes the tests we have, but I can't guarantee that it's correct.

### DIFF
--- a/src/pyscipopt/scip.pxd
+++ b/src/pyscipopt/scip.pxd
@@ -456,9 +456,6 @@ cdef extern from "scip/scip.h":
     ctypedef struct SCIP_HASHMAP:
         pass
 
-    #ctypedef struct SCIP_BOUNDTYPE:
-    #   pass
-
     ctypedef struct SCIP_BDCHGIDX:
         pass
 


mattmilten:
Oh nice! Yes, this is much better and way less error-prone.

Joao-Dionisio:
Thank you, Mathias!

I think this should be working, @mmghannam, but I'm a bit wary of pushing this without doing proper tests on all these things. On the other hand, I suppose that it's difficult that this breaks even more stuff...

There's also the matter of having a double declaration of ```SCIP_BOUNDTYPE``` in line 44 and 441, but that was there in the previous versions as well, but hidden.

mmghannam:
Thanks a lot João! Let's add some tests and merge this straight away.

Joao-Dionisio:
Okay @mmghannam, I think I added every event type now, a bunch were missing. 

As for the tests, I'm only testing whether the events that PySCIPOPt is catching and executing are the same. I could add a lot more sophisticated tests, but it will take a reaaaally long time to create a proper test for every event.

PS: I really need Github Copilot...

mmghannam (approved):

Issue #713, opened 2023-08-01
Accessing information of constraints


MoRic3:
I am using PySCIPopt version 4.3.0 which corresponds to SCIP version 8.0.3.

I am wondering if PySCIPopt/SCIP allows to do something like the following:

Given a model object, I want to access information of a constraint. In fact, I want to know which variables appear in the specific constraint. And I want to get the constraint rule/expression, so I can use it further, e.g., for embedding it into a new constraint.

I attached code showing what I want to do using the Pyomo environment (see [https://anaconda.org/conda-forge/pyomo](https://anaconda.org/conda-forge/pyomo)). 

Do you see any possibility of doing something similar in PySCIPopt?
Thanks a lot.

```
from pyomo.environ import *
from pyomo.core.expr.current import identify_variables

# define the model
def build_model():
    model = ConcreteModel()

    # define variables
    model.x1 = Var(within=Reals, bounds=(0,1))
    model.x2 = Var(within=Reals, bounds=(0,1))
    model.x3 = Var(within=Integers, bounds=(-4,1))

    # define constraint
    model.cons_nonlin0 = Constraint(expr = (1 - model.x1**2 - model.x2**2) <= 0)

    # define objective
    model.objective0_nonlin = Objective(expr = model.x2 - exp(model.x3))

    return model

model = build_model()

for c in model.component_objects(Constraint):
    var_list = list(identify_variables(c.body))
    print('appearing variables (pyomo objects):', var_list)
    
    print('constraint expression:', c.body)
    
    # add modified constraint
    model.add_component('modified_'+c.name,
                        Constraint(expr = c.body + 5 <= 0))
    
    # deactivate old one
    c.deactivate()
```

Joao-Dionisio:
I did not manage to find anything, sorry... @mattmilten @fserra, do any of you know how one would do this? Ambros asked something about it a few days ago as well and I didn't know how to answer.

ambros-gleixner:
I guess for this we would just need to interface SCIPconsGetVars(), see https://scipopt.org/doc/html/cons_8c.php#aeacdfba7623596329c1cfb1500cfdb36, right? That should not be too hard to do.

Joao-Dionisio:
I think I've already done everything, Ambros, I'm only having some issues with the compilation (it's my first modification to the source code). I'm waiting for some help with it, but with the weekend and people on vacation, it may take a bit.

ambros-gleixner:
Great, thanks!

[...]

PR #760, opened 2023-12-02
Add addExprNonlinear

Joao-Dionisio:
This is not the best way to solve the issue, but I think it gets the work done.

- SCIPaddExprNonlinear requires a nonlinear constraint, but it should be possible to convert a linear constraint into a nonlinear one in ```scip.pxi``` (```x = self.addVar(lb=0,ub=0), self.addCons(linear expr + x**2 <= RHS)```). This would work with other methods.
- I'm using a Python union, since ```GenExpr``` doesn't include ```Expr```. Most likely it should.
- I couldn't find a proper way to get a scip expression from the python expression, so I add a temporary constraint, and then delete it.


Joao-Dionisio:
No worries @Opt-Mucca, I think it was me who spammed PRs of dubious interest :)

This came up with Issue #713, and it seems like it could be helpful, maybe for reoptimizations and such. I suppose it would serve the same purpose that it serves in SCIP.

Opt-Mucca:
@Joao-Dionisio I'm convinced for this one after seeing the issue you commented. I still think you could simply keep track of the expression inside of the constraint, but that is actually duplicate work, and maybe not that intuitive.

Suggestion: Do you need to add `coef * expr` instead of just the expression? The user could always just put `coef * expr` as the input then.

Joao-Dionisio:
@Opt-Mucca yeah, you just need the expression, but I tried to keep the function as close to the [SCIP function](https://www.scipopt.org/doc-8.1.0/html/group__CONSHDLRS.php#ga7d9ee22be522a00878f626aac5d96866) as I could, which does take ```coef``` as an input.

Opt-Mucca:
Then I'm happy with this one!

Joao-Dionisio:
Oh man, what have I done... I just wanted to add a TODO :(

mmghannam (approved):

PR #764, opened 2023-12-07
Mt/add getconshdlrname

Opt-Mucca:
Adds a single function to the class Constraint.

Simply allows you to get access to a string that tell you the origin of the constraint.


Joao-Dionisio (commented):
Thanks, Mark! And congrats on your defense :)

mmghannam (approved):
Looks good, thanks @Opt-Mucca!

Issue #766, opened 2023-12-14
Error printing event in event handler


ryanjoneil:
**Describe the bug**
I get an error when I attempt to print or otherwise inspect the `event` argument of the `Eventhdlr.eventexec` method.

**To Reproduce**

Here is a simple example:
```python
from pyscipopt import quicksum as sum
import pyscipopt as scip
import random

class handler(scip.Eventhdlr):
    def eventexec(self, event):
        print(event)

    def eventinit(self):
        self.model.catchEvent(scip.SCIP_EVENTTYPE.LPEVENT, self)

    def eventexit(self):
        self.model.dropEvent(scip.SCIP_EVENTTYPE.LPEVENT, self)

if __name__ == '__main__':
    m = scip.Model()
    m.includeEventhdlr(handler(), 'handler', 'decription')

    x = [m.addVar(obj=i) for i in range(1000)]
    for _ in range(1000):
        left = sum(random.sample(x, random.randint(5, 15)))
        right = random.randint(1, 100)
        m.addCons(left >= right)

    m.optimize()
```

When I run it I get this:

```txt
feasible solution found by trivial heuristic after 0.0 seconds, objective value 4.995000e+10
presolving:
(round 1, fast)       1 del vars, 7 del conss, 0 add conss, 999 chg bounds, 0 chg sides, 0 chg coeffs, 0 upgd conss, 0 impls, 0 clqs
(round 2, fast)       1 del vars, 7 del conss, 0 add conss, 1998 chg bounds, 0 chg sides, 0 chg coeffs, 0 upgd conss, 0 impls, 0 clqs
   (0.0s) running MILP presolver
   (0.0s) MILP presolver found nothing
   (0.0s) symmetry computation started: requiring (bin +, int +, cont +), (fixed: bin -, int -, cont -)
   (0.0s) no symmetry present (symcode time: 0.00)
presolving (3 rounds: 3 fast, 1 medium, 1 exhaustive):
 1 deleted vars, 7 deleted constraints, 0 added constraints, 1998 tightened bounds, 0 added holes, 0 changed sides, 0 changed coefficients
 0 implications, 0 cliques
presolved problem has 999 variables (0 bin, 0 int, 0 impl, 999 cont) and 993 constraints
    993 constraints of type <linear>
Presolving Time: 0.01

[...]

PR #769, opened 2023-12-17
Print names of events and current stage

Joao-Dionisio:
Using ```model.getStage()``` we get an integer, not very informative. We now have ```model.getStageName()```, which gives us the name of the stage. Same for events. This does not solve #766, as directly trying to print an event should still give an error. Similar to PR #484.

Draft because I'm sure there are better ways to do this.


src/pyscipopt/scip.pxi
```diff
@@ -305,6 +310,19 @@ cdef class Event:
...
+            self._getEventNames()
+        return EventNames[self.getType()]
+
+    def _getEventNames(self):
+        """gets event names"""
+        for name in dir(PY_SCIP_EVENTTYPE):
+            attr = getattr(PY_SCIP_EVENTTYPE, name)
+            if isinstance(attr, int):
+                EventNames[attr] = name
+
```
mmghannam, line 325: Would be great if you could implement the __str__ method too (read here for more https://www.digitalocean.com/community/tutorials/python-str-repr-functions)

src/pyscipopt/scip.pxi
```diff
@@ -305,6 +310,19 @@ cdef class Event:
...
+    def _getEventNames(self):
+        """gets event names"""
+        for name in dir(PY_SCIP_EVENTTYPE):
+            attr = getattr(PY_SCIP_EVENTTYPE, name)
+            if isinstance(attr, int):
+                EventNames[attr] = name
+        
     def __repr__(self):
         return self.getType()
```
mmghannam, line 327:
```suggestion
        return str(self.getType())
```
From what I understand __repr__ should always return a string so this behavior is weird.

mmghannam (changes requested):
Thanks João for the useful contribution!

src/pyscipopt/scip.pxi
Joao-Dionisio: I added it to the Events, Mo, but couldn't do it for the Stages, as they're not from a class.

mmghannam: Perfect!

mmghannam (approved):

CHANGELOG.md
```diff
@@ -2,13 +2,7 @@
 
 ## Unreleased
 ### Added
-- Add getConshdlrName to class Constraint
```
mmghannam, line 5: Something seems to have been merged wrong here, this line should be in the 4.4.0 release, and your change should be in a new _Unreleased_ section.

Joao-Dionisio: Oh damn it, I always mess up the CHANGELOG somehow. I think it's alright now. Thanks for looking into this, Mo!

Issue #663, opened 2023-03-23
Model Non Linear Objective function


andreafresa:
Hi, I have the following objective function

$$ \max_{n_{ij}(t)} &&\sum_{i \in I} \sum_{j \in J_i}Q_ib({n_{ij}(t)}) + \sum_{i \in I}V_iw_{ij}\sum_{j\in J_i}\min({b(n_{ij}(t)) , Q_i- \sum_{k <j}b(n_{ik}(t))})a_{ij} $$

Where n_ij is a matrix of the decision variables. My function is not linear, because of min, thus it is a Non-Linear Integer Programming problem (NILP).
I converted min(a,b) = (a + b)/2 + abs(a-b)/2

I implemented my function in the following way:

```
def new_npilp(queue , service_rate , accuracy , V, numcores):
    model = scipop.Model("npilp")
    x = {}
    for i in range(len(service_rate)):
            x[i]=model.addVar("x"+str(i), vtype="I", lb=0, ub=numcores)    
    nonlinexp = {}
    term1 = {}
    term2 = {}
    model = scipop.Model("npilp")
    model.addCons(scipop.quicksum(x[i] for i in range(len(service_rate))) <= numcores)
    for i in range(len(service_rate)):
        term1[i] = accuracy[i]*service_rate[i][1]*x[i]
        term2[i] = accuracy[i]*queue-scipop.quicksum(service_rate[k][1]*x[k] for k in range(len(service_rate)))
        nonlinexp[i] = (term1[i]+term2[i]/2) + abs(term1[i]-term2[i])/2
        print(type(nonlinexp[i]))
    fun = scipop.quicksum(queue*service_rate[i][1]*x[i] + nonlinexp[i] for i in range(len(service_rate)))
    model.setObjective(fun, "maximize")
    model.optimize()
    
```

However I get the following error:

> AssertionError: given coefficients are neither Expr or number but SumExpr

Could someone help me here? How can I solve this using PySCIPOpt? If not, why?

Joao-Dionisio:
Hello, @andreafresa! Can you also paste the data for your problem so that I can try to run it?

In the meantime, I think you can do what I believe is called an epigraph reformulation. If your nonlinear objective is max f(x), you can instead do max z and add the constraint z >= f(x).

andreafresa:
Thank @Joao-Dionisio , I tried your formulation and also this one:


```
def new_npilp(queue , service_rate , accuracy , V, numcores):
[...]

Issue #387, opened 2020-06-09
division in objective: given coefficients are neither Expr or number but ProdExpr


cnglen:
How to add objective which including division expression, such as:

a=model.addVar("a")
b=model.addVar("b")
c=model.addVar("c")

In [354]: model.setObjective((a+b)/c)
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-354-e958238996b9> in <module>
----> 1 model.setObjective((a+b)/c)

src/pyscipopt/scip.pyx in pyscipopt.scip.Model.setObjective()

AssertionError: given coefficients are neither Expr or number but ProdExpr

fserra:
sorry, that is a bad error message. PySCIPOpt doesn't allow for nonlinear objectives. So if you want to minimize `(a + b )/c` you should introduce a new variable, `t`, add a constraint `(a + b )/c <=t` and set as objective function just `t`.

cnglen:
Thanks.

Issue #212, opened 2018-11-01
AssertionError when building model


cmjdxy:
"AssertionError" is raised when I run my program. I have checked my code time and again, but it doesn't work. Can anyone help me out? You have my gratitude.
The code:
`
import numpy as np
from pyscipopt import Model, quicksum

def QP_model(A, w, c, name='QP'):
    '''
    min : x'Ax/2
    s.t.: w'x <= c, 0 <= x <= 1
    '''
    n = len(A)
    model = Model(name)
    x = dict()
    for i in range(n):
        x[i] = model.addVar('x(%d)' % i, vtype='C', lb=0.0, ub=1.0)
    model.addCons(quicksum(w[i]*x[i] for i in range(n)) <= c)
    objfun = quicksum(A[i][j]*x[i]*x[j]/2 for i, j in np.ndindex(n, n))
    model.setObjective(objfun, 'minimize')
    model.data = x
    return model

def main():
    '''
    Main interface.
    '''
    # data
    A = np.random.rand(3, 3)
    w = np.random.rand(3)
    c = np.random.rand()
    # model
    model = QP_model(A, w, c)
    model.optimize()

if __name__ == '__main__':
    print('running...')
    main()
`
The error message:
`
Traceback (most recent call last):

  File "<ipython-input-50-4fc84a02b499>", line 1, in <module>
    runfile('D:/Essays/QMKP/example/temp.py', wdir='D:/Essays/QMKP/example')

  File "d:\ProgramData\Anaconda3\lib\site-packages\spyder\utils\site\sitecustomize.py", line 705, in runfile
    execfile(filename, namespace)

  File "d:\ProgramData\Anaconda3\lib\site-packages\spyder\utils\site\sitecustomize.py", line 102, in execfile
    exec(compile(f.read(), filename, 'exec'), namespace)

  File "D:/Essays/QMKP/example/temp.py", line 41, in <module>
    main()

  File "D:/Essays/QMKP/example/temp.py", line 35, in main
    model = QP_model(A, w, c)

  File "D:/Essays/QMKP/example/temp.py", line 21, in QP_model
    model.setObjective(objfun, 'minimize')

  File "src\pyscipopt\scip.pyx", line 779, in pyscipopt.scip.Model.setObjective

AssertionError
`
N.B. I am using Anaconda3 and Windows 10.

ramabile:
Instead of writing:
[...]

Issue #168, opened 2018-06-02
ValueError: Nonlinear objective functions are not supported!


FCYtheFreeman:
My integer programming has multiple solutions，and I want to find the best solution that has the minimum number of 0. 
Since my variables is non-negative integer, and the sum of variables is confirmed，my first method is set model.setObjective(x^2 + y^2 + z^2 , "maximize"), but it leads to an error: 
ValueError: Nonlinear objective functions are not supported!

How can I solve this problem？Or is there other ways I can achieve my goal？

Thank you !

fserra:
check my answer in #167

PR #781, opened 2024-01-24
Add possibility for nonlinear objective function

Joao-Dionisio:
It's passing all the tests, currently the only issue is that calling ```getObjective``` isn't as nice as with a linear objective.

I deleted an assertion that I'm not sure what its purpose was, hopefully it was only to prevent nonlinear objectives, which should be alright now.

EDIT: just realized that this was on the wrong branch...

### DIFF
--- a/src/pyscipopt/scip.pxi
+++ b/src/pyscipopt/scip.pxi
@@ -1322,12 +1322,8 @@ cdef class Model:
 
         # turn the constant value into an Expr instance for further processing
         if not isinstance(coeffs, Expr):
-            assert(_is_number(coeffs)), "given coefficients are neither Expr or number but %s" % coeffs.__class__.__name__
             coeffs = Expr() + coeffs
 
-        if coeffs.degree() > 1:
-            raise ValueError("Nonlinear objective functions are not supported!")
-        
         if clear:
             # clear existing objective function
             self.addObjoffset(-self.getObjoffset())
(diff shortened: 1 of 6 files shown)


src/pyscipopt/scip.pxi
```diff
@@ -1322,12 +1322,8 @@ cdef class Model:
 
         # turn the constant value into an Expr instance for further processing
         if not isinstance(coeffs, Expr):
-            assert(_is_number(coeffs)), "given coefficients are neither Expr or number but %s" % coeffs.__class__.__name__
```
Joao-Dionisio, line 1325: This is the assertion that I'm not so sure what it's doing.

mattmilten:
I have no clue, really. Maybe we were just too scared back then ;-)

fserra:
I think we might have thought of 2 things:
1. We were always afraid of making pyscipopt deviate too much from SCIP. I guess this qualified as deviating too much
2. I haven't looked at the MR but one would need to add an extra variable. I guess we were afraid about the fact that now we have to deal with this variable somehow... maybe the user doesn't understand why there is an extra variable in the problem. I feel/felt that forcing the user to add the variable themselves would give the user more control.

I think that was it. Anyway, that was back then. I let you guys decide if any of that makes sense.

Opt-Mucca:
@Joao-Dionisio @mmghannam I slept on this, and am now firmly against adding such functionality. It requires keeping track of some hidden overhead, and I think will confuse more users than help. I think we should just add an appropriate error warning when a user tries to add a non-linear objective. 
This then isn't too confusing for new users (if the error warning includes a modelling tip)

Joao-Dionisio:
Hey @Opt-Mucca, fair enough. I think the main point is that the required reformulation isn't obvious for new users, so the modeling tip would be a must, in my opinion, maybe even explicitly saying which constraint to add. I've come across a fair amount of people asking how to have a nonlinear objective: #663, #387, #212, #168, #167, #41, [SO1](https://stackoverflow.com/questions/53458021/division-objective-function-in-pyscipopt), [SO2](https://stackoverflow.com/questions/65796202/pyscipopt-does-not-handle-1-variable-and-variablenegative-number-in-the-objec), [ORSE1](https://or.stackexchange.com/questions/5565/doesnt-pyscipopt-handle-nonlinear-objective-functions).

For posteriority's sake, I'll leave my counterarguments, but if we elaborate on the warning as you said, then I'm fine with closing.

- Having the option to add a nonlinear objective directly doesn't exclude the ability to not do it. So if the user wants more control, he can still model things as he pleases. 
-  I think it's reasonable to assume that someone who is bothered by this change is more likely to know about the reformulation and find ways to do what he wants (which this PR wouldn't impact in any way), than a less experienced user to find out about the reformulation by himself.
- I don't know what restrictions we have in place for naming variables and constraints, but we could name the new variable as something like "Variable modeling the nonlinear objective - see [some sort of reference]". Also, we are only adding one variable and one constraint. We could also add a message saying that we did this so that users aren't unaware.
- PySCIPOpt is used in a couple of classes, I think we can say that inexperienced users have a higher weight here than people who work with SCIP proper (excluding those using it as a black box), so having a slight difference between PySCIPOpt and SCIP doesn't bother me immensely (but I admit it might be my own lack of experience showing).

As for the randomness in the tests, I figured that a less deterministic test would be better, as we would be testing a wider array of problems, and possibly catching an error that we wouldn't otherwise (again, possible lack of experience showing).

An additional question. SCIP already handles nonlinear objectives, right? Can we not just pass it over somehow?

(I know I need to work on my conciseness, Mark, sorry :( )

mmghannam:
I think both your arguments/concerns @Joao-Dionisio, @Opt-Mucca & @fserra make sense. This issue is similar to what's happening in #776 too. On the one hand the feature would help new users get things working quickly, but they are also not mapping to a feature in SCIP and could include judgement calls on how they are implemented. Also, including them as an example, or a warning message would be less convenient for users. So I suggest a somewhat middle ground solution, we could add a sub-package called maybe `recipes`, and it would include these type of functions. If it's implemented as external functions (that are not methods on the model). It will make it clearly separate from SCIP's API while still allowing a place for contributions for convenience methods/formulations.

mmghannam (approved):

PR #791, opened 2024-02-19
Modernize Python metadata

mkoeppe:
Moving them to pyproject.toml


mmghannam (approved):
Thanks a lot @mkoeppe! this looks so much better. I fixed the conflict with your other PR (by removing the cython<3 requirement). Should be good to merge once the tests pass.

PR #792, opened 2024-02-19
Add support for Cython 3

### LINKED ISSUE #790: "python setup.py build" fails
**Describe the bug**

Build fails with the attached log.

[log.txt](https://github.com/scipopt/PySCIPOpt/files/14315766/log.txt)

**To Reproduce**

python3.12 setup.py build

**System**
 - OS: Debian
 - Version 13
 - SCIP version. 8.1.0 
 - How did you install `pyscipopt`?
  from source
**Additional context**
Cython is 3.0.8 from Debian (pip installed cython is the same)
I get the same errors with v4.4.0 or master `5e6a747e8db0`

I reported the build problem from Debian testing. I checked in Debian stable (12), and it seems to work there.
The default version of python changed to 3.12, although "python3.11 setup.py build" seems to yield similar results in Debian testing.

mkoeppe:
Just by adding some `noexcept`

Fixes #790 @bremner

### DIFF
--- a/src/pyscipopt/scip.pxd
+++ b/src/pyscipopt/scip.pxd
@@ -501,7 +501,7 @@ cdef extern from "scip/scip.h":
     ctypedef union SCIP_DOMCHG:
         pass
 
-    ctypedef void (*messagecallback) (SCIP_MESSAGEHDLR *messagehdlr, FILE *file, const char *msg)
+    ctypedef void (*messagecallback) (SCIP_MESSAGEHDLR *messagehdlr, FILE *file, const char *msg) noexcept
     ctypedef void (*errormessagecallback) (void *data, FILE *file, const char *msg)
     ctypedef SCIP_RETCODE (*messagehdlrfree) (SCIP_MESSAGEHDLR *messagehdlr)
 
--- a/src/pyscipopt/scip.pxi
+++ b/src/pyscipopt/scip.pxi
@@ -975,10 +975,10 @@ cdef class Constraint:
                 and self.scip_cons == (<Constraint>other).scip_cons)
 
 
-cdef void relayMessage(SCIP_MESSAGEHDLR *messagehdlr, FILE *file, const char *msg):
+cdef void relayMessage(SCIP_MESSAGEHDLR *messagehdlr, FILE *file, const char *msg) noexcept:
     sys.stdout.write(msg.decode('UTF-8'))
 
-cdef void relayErrorMessage(void *messagehdlr, FILE *file, const char *msg):
+cdef void relayErrorMessage(void *messagehdlr, FILE *file, const char *msg) noexcept:
     sys.stderr.write(msg.decode('UTF-8'))
 
 # - remove create(), includeDefaultPlugins(), createProbBasic() methods
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 ### Added
+- Add support for Cython 3
 - Added methods for getting the names of the current stage and of an event
 ### Fixed
 - Fixed outdated time.clock call in gcp.py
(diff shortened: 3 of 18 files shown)


leoneifler:
Thank you for updating this. Just needs a one-line CHANGELOG entry, looks ready to merge otherwise 👍

mkoeppe:
Note that I have not removed the version constraint, so when build isolation is in use, it continues to use Cython 0.29.x.
If you want me to change this in this PR, let me know

leoneifler:
> Note that I have not removed the version constraint, so when build isolation is in use, it continues to use Cython 0.29.x. If you want me to change this in this PR, let me know

Good point. I think since it is now supported it makes sense to raise the version requirement, as well. Or is there any reason against this that I am not seeing?

mkoeppe:
> is there any reason against this that I am not seeing?

Not really, but I haven't tested anything other than that it compiles.

leoneifler (approved):

src/pyscipopt/scip.pxi
```diff
@@ -975,10 +975,10 @@ cdef class Constraint:
                 and self.scip_cons == (<Constraint>other).scip_cons)
 
 
-cdef void relayMessage(SCIP_MESSAGEHDLR *messagehdlr, FILE *file, const char *msg):
+cdef void relayMessage(SCIP_MESSAGEHDLR *messagehdlr, FILE *file, const char *msg) noexcept:
     sys.stdout.write(msg.decode('UTF-8'))
 
-cdef void relayErrorMessage(void *messagehdlr, FILE *file, const char *msg):
+cdef void relayErrorMessage(void *messagehdlr, FILE *file, const char *msg) noexcept:
```
Joao-Dionisio, line 981: Hey @mkoeppe! Sorry for not looking into your stuff, but it flies a bit over my head... Nevertheless, there are only 3 lines where you add ```noexcept``` where there's no ```with gil```. Is that intentional?

I also find it very sad that the test coverage decreases so much, but that's really not an actual problem :)

src/pyscipopt/scip.pxd
```diff
@@ -501,7 +501,7 @@ cdef extern from "scip/scip.h":
     ctypedef union SCIP_DOMCHG:
         pass
 
-    ctypedef void (*messagecallback) (SCIP_MESSAGEHDLR *messagehdlr, FILE *file, const char *msg)
+    ctypedef void (*messagecallback) (SCIP_MESSAGEHDLR *messagehdlr, FILE *file, const char *msg) noexcept
```
Joao-Dionisio, line 504: Also here.

src/pyscipopt/scip.pxi
```diff
@@ -975,10 +975,10 @@ cdef class Constraint:
                 and self.scip_cons == (<Constraint>other).scip_cons)
 
 
-cdef void relayMessage(SCIP_MESSAGEHDLR *messagehdlr, FILE *file, const char *msg):
+cdef void relayMessage(SCIP_MESSAGEHDLR *messagehdlr, FILE *file, const char *msg) noexcept:
```
Joao-Dionisio, line 978: And here.

src/pyscipopt/scip.pxi
Joao-Dionisio: Okay, looking at the Cython page makes me think this is also good.

mkoeppe: > where you add `noexcept` where there's no `with gil`

I started by adding `noexcept`s one by one to remove the compiler errors. 
I switched to taking care of the `noexcept with gil` by global replace. 
The `noexcept` where there's no `with gil` are all needed.

mmghannam:
Thanks @mkoeppe  for your contribution! For some reason the code coverage tool seems to be unhappy with this although the lines changes were not covered before. But it looks good to me!

PR #801, opened 2024-02-29
Remove bison from brew install since bison is installed by default (on macos:latest at least)

ju-manns:


Opt-Mucca (approved):
From the errors for the Mac jobs this should be fine!

PR #802, opened 2024-03-01
Update _version.py

mmghannam:


Opt-Mucca (approved):
Should be good. At some point we need to update the CHANGELOG too

PR #813, opened 2024-03-12
Fix setLogfile method to handle None path

liangbug:


Opt-Mucca:
Could this not be accomplished with the following code?
```
if path is not None:
    c_path = str_conversion(path)
else:
    c_path = NULL
SCIPset.......
```

Joao-Dionisio:
Or even 

```
c_path = str_conversion(path) if path else NULL
SCIPset ...
```

If we want to keep the same syntax

Opt-Mucca:
I like the suggestion of @Joao-Dionisio !

liangbug:
I added unittest and changelog.


@Joao-Dionisio, @Opt-Mucca, The follownng codes would cause compiling error.
```
if path is not None:
    c_path = str_conversion(path)
else:
    c_path = NULL
SCIPset.......
```
```
c_path = str_conversion(path) if path else NULL
SCIPset ...
```
```
Error compiling Cython file:
src\pyscipopt\scip.pxi:4914:21: Cannot convert 'void *' to Python object
````

Or you prefer the following code instead.
```
if path:
    c_path = str_conversion(path)
    SCIPsetMessagehdlrLogfile(self._scip, c_path)
else:
    SCIPsetMessagehdlrLogfile(self._scip, NULL)
```

**Comment by Opt-Mucca** [comment]
@liangbug I'd prefer the code you suggested!

**Comment by liangbug** [comment]
@Opt-Mucca, I updated the code 
```
if path:
    c_path = str_conversion(path)
    SCIPsetMessagehdlrLogfile(self._scip, c_path)
else:
    SCIPsetMessagehdlrLogfile(self._scip, NULL)
```

PR #815, opened 2024-03-14
Relax stage checking for getObjVal, getVal

Joao-Dionisio:
```getObjVal, getVal``` are too strict in the stage checks. For example, the following code crashes:

```
from pyscipopt import Model

m = Model()

x = m.addVar()
m.setObjective(x)

m.setParam("limits/solutions", 1)
m.optimize()

print("Stage: ", m.getStage())

print(m.getNSols())
print(m.getObjVal()) # crashes here, even though the model does have solutions
```

I changed it to reflect SCIP.

I also removed the ```noexcept``` from the code, because I couldn't compile otherwise. It's probably me being a dummy, but I do have Cython >= 3 (EDIT: It was me being a dummy).  Of course, I imagine that not removing them is desirable. 

EDIT2:

Okay, right now I'm having segfaults in ```test_customizedbenders.py```, and I'm pretty sure it's because ```self._bestSol.sol``` is ```NULL```, but I can't figure out why that is.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,9 @@ __pycache__/
 # C extensions
 *.so
 
+# macOS dir files
+.DS_Store
+
 # Distribution / packaging
 .Python
 build/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 ### Added
+### Fixed
+### Changed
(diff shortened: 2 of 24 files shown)


Joao-Dionisio:
PR closed while there are still problems with it. Opened issue #847 to keep track of it.

PR #823, opened 2024-03-20
Faster compilation in PR pipeline

mmghannam:


Joao-Dionisio (approved):
Thank you thank you thank you thank you!

PR #824, opened 2024-03-20
Speedup: move to macos arm runner for PR pipeline

mmghannam:


mmghannam:
It doesn't seem to make a difference, closing this.

PR #834, opened 2024-04-02
Fix pipelines

mmghannam:
- Use backwards compatible type hints.
- Install bison in macos pipeline.
- Install locales needed by tests in ubuntu integration test.

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -8,9 +8,6 @@ on:
   push:
     branches:
       - 'master'
-  pull_request:
-    branches:
-      - 'master'
 
 jobs:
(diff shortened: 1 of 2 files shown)


mmghannam:
Should be working now with these changes https://github.com/scipopt/PySCIPOpt/actions/runs/8528377932?pr=834

Joao-Dionisio (approved):
It looks good to me, Mo! Genuine question, why do you prefer typing List to regular list?

mmghannam:
> It looks good to me, Mo! Genuine question, why do you prefer typing List to regular list?

I don't 😄 It's just that they work with older python versions.

PR #837, opened 2024-04-03
add support for disjunction constraints

ctdunc:
Adds preliminary support for disjunctive constraints, and new methods for creating constraints without adding them to the problem itself. 

Long-term, it may be nice to have a slate of methods like `createCons<T>()` which create constraints without adding them to the problem, and `.addCons<T>(<T>)` methods which can take constraints as inputs, so that we can add more support for compound constraint creation methods like conjunction,


src/pyscipopt/scip.pxi
```diff
@@ -2503,6 +2334,273 @@ cdef class Model:
...
+                      propagate=propagate, local=local,
+                      modifiable=modifiable, dynamic=dynamic,
+                      removable=removable,
+                      stickingatnode=stickingatnode)
+        #  we have to pass this back to a SCIP_CONS*
+        # object to create a new python constraint & handle constraint release
+        # correctly. Otherwise, segfaults when trying to query information
+        # about the created constraint later.
+        pycons_initial = self.createExprCons(cons, **kwargs)
+        scip_cons = (<Constraint>pycons_initial).scip_cons
```
ctdunc, line 2419: This is where the `free` issue discussed in #837 is currently handled.

Opt-Mucca:
@ctdunc Good changes! I think they add more functionality, and I am all for supporting more constraint classes.

My current concern: The naming of  `createExprCons`. An `ExprCons` is already technically created by the user when calling such a function. The function is rather creating a PySCIPOpt `Constraint` object. I feel like the name `createCons` will lead to a lot of confusion though. I am concerned about users using such a function and then trying to call `addCons` with the resulting object. I will need to discuss with @mmghannam and @Joao-Dionisio 

Aside from that: The code looks good and just has some minor things to tidy up.

Opt-Mucca (approved):

Opt-Mucca:
@ctdunc Thanks for this merge request! This is a really cool feature.

I've changed some minor things that I raised earlier. I renamed the function `createExprCons`, and moved some of what I believe were leftover debug stuff.

I'll now merge the branch! For your comment on adding a cons with a Python object: I agree. Given that users can now create a constraint object easily, they should be able to pass it as an option to `addCons`. This goes on the TODO list.

ctdunc:
Looks great! Thanks for including :) Have been a bit tied up with other project at work, but hoping to get back to this soon.
Having had a bit more time to dig around in getting user-provided decompositions working, I agree with these name changes. Since the underlying SCIP structures typically require a pointer to a SCIP problem, it doesn't really make sense to allow users to create python objects independently of a SCIP Model().

If my understanding of this is wrong, please LMK as I have more features I would like to attempt to contribute & want to avoid this type of confusion going forward. Thanks :)

Opt-Mucca:
You're understanding is correct! A SCIP pointer is in general always required, and therefore does handicap creating Python objects independently. Very happy for any other contributions when you have the time (I will get to the user defined decompositions in the next week or two)

PR #842, opened 2024-04-11
add support for user-provided decompositions

ctdunc:
this commit will add support for user-provided decompositions as dicts of `{var: label}`, `{cons:label}` pairs.

still needs testing & examples, but opening a draft so maintainers can follow along if you have time.


src/pyscipopt/scip.pxi
```diff
@@ -975,6 +975,130 @@ cdef class Constraint:
...
+        self.create()
+    @staticmethod
+    cdef create(SCIP_DECOMP* scip_decomp):
+        if scip_decomp == NULL:
+            raise Warning("cannot create Constraint with SCIP_CONS* == NULL")
+        decomp = Decomposition()
+        decomp.scip_decomp = scip_decomp
+        return decomp
+
+    def isOriginal(self):
```
Opt-Mucca, line 990: We're trying to make sure new functionality isn't added without having tests. So for all available functions you're adding please put them into a simple test (one test can cover multiple functions)

src/pyscipopt/scip.pxi
```diff
@@ -975,6 +975,130 @@ cdef class Constraint:
...
+        return SCIPdecompIsOriginal(self.scip_decomp)
+
+    def getAreaScore(self):
```
Opt-Mucca, line 993: As this isn't standard knowledge (or at least I don't know it), if you can add a one line comment for what the function does then please do. This really helps out users and means they don't need to go through SCIP itself for answers.

Opt-Mucca: This extends to most functions below

Opt-Mucca (commented):
Puhhhhhhh. I misunderstood this MR a bit at the start. Not the smartest bear in the woods.

I'm not an expert with decompositions, and am not sure what general users want from this functionality. I will need to look into how SCIP handles them before I can check if everything correct.

In the meantime: More tests and comments are needed. I am against adding such functionality without making it extremely clear to users what it does. This is just a draft though, so I guess speed isn't a concern right now.

src/pyscipopt/scip.pxi
ctdunc: Will add more tests & docs when I get a moment, sorry. This one got put on the backburner due to other stuff at work.

For our use case, we have a modeling layer on top of PySCIPOpt that basically uses dataframe-like syntax to create groups of constraints using a table that characterizes our variables. So the goal is to break `.groupby` statements into subproblems where each of the groups can be solved ~in parallel. We have some combinatorial constraints as well, so decompositions seem like a natural place to reduce the number of combinations to check.

Joao-Dionisio:
This PR has been inactive for a long time, and the OP mentioned not having the time to work on it anymore. So, to clear the PR page, I'm closing it. It may make sense to revisit in the future, as the feature is interesting.

PR #843, opened 2024-04-24
Fix locale fix

Joao-Dionisio:
We were changing the locale for everything (locale.LC_ALL), when I think we only wanted locale.LC_NUMERIC. Changing it for everything created new problems:

```
from pyscipopt import Model

m = Model()
m.readProblem("inf_model.cip")
m.writeParams("params.set")

m2 = Model()
m.readProblem("inf_model.cip")
```

```
[scip_var.c:766] ERROR: constant in linear sum
[cons_linear.c:17252] ERROR: no luck in parsing linear sum '<r'[0,0,0]>[C] -0.0282485875706215<r[0,0,0]>[C] == 0'
[reader_cip.c:722] ERROR: syntax error when reading constraint (line: 216):
  [linear] <c1>: <r'[0,0,0]>[C] -0.0282485875706215<r[0,0,0]>[C] == 0;
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "src/pyscipopt/scip.pxi", line 5186, in pyscipopt.scip.Model.readProblem
  File "src/pyscipopt/scip.pxi", line 266, in pyscipopt.scip.PY_SCIP_CALL
OSError: SCIP: read error!
```

Problem in question:
[inf_model.txt](https://github.com/scipopt/PySCIPOpt/files/15095760/inf_model.txt)

This way it's still passing the locale check we set up, and this problem disappears.

It might be that this raises other issues, but I think it's better to deal with them as they appear since it's difficult to predict how different locales can cause problems.


mmghannam (changes requested):
Looks good, thanks João! Maybe we should add a test with the instance that's faulty to make sure this doesn't get reintroduced later. What do you think?

Joao-Dionisio:
Sure @mmghannam! Where do you think the instance should be? Should we create an instances subfolder in the test folder, or just put it among the tests?

Joao-Dionisio:
Done @mmghannam! I also realized I forgot to add the locale stuff to write methods, so your suggestion ended up being very helpful :D

mmghannam (approved):

PR #845, opened 2024-05-03
Add getnorigconss

Opt-Mucca:


Joao-Dionisio (commented):
Looks good, @Opt-Mucca!

Looking at [getVars](https://scipopt.github.io/PySCIPOpt/docs/html/scip_8pyx_source.html#l01677), it has an argument ```transformed```, that returns original or transformed variables depending on the argument. I think the syntax should be uniform, but I don't have a strong preference for either one. What do you think, Mark?

(Interestingly, we don't have a way of getting the number of original variables (SCIPgetNOrigVars exists). I can make a PR, but I want to make sure where we're going with this.)

Opt-Mucca:
@Joao-Dionisio That is a great suggestion and a complete oversight by me! I will change the behaviour to match (can also add getnorigvars)

Opt-Mucca:
Added changes and actually made the test work now (is just a simple LP that tests that after presolve there are less constraints than we originally added). 

The only potential issue I see: depending on the function you're using `transformed` has a different default value. I currently set it so the default values stick to the older functionality (don't see any better solution).

Joao-Dionisio (approved):

Joao-Dionisio:
I agree that it's annoying, @Opt-Mucca, but I like your solution, you aren't introducing any problems that weren't there before.

Opt-Mucca:
Thanks for adding the extra docstrings

PR #852, opened 2024-05-15
Minor changes to test_pricer

Joao-Dionisio:
More accurate numerics and displaying of the solution


mmghannam (approved):
Looks good, thanks João!

PR #859, opened 2024-06-13
Add parser to read SCIP statistics

Joao-Dionisio:
I don't think I've ever done a proper parser, so the code might just be too ugly to include. I think it's doing what it's supposed to do, though.


Opt-Mucca (commented):
I'm not sure whether this functionality should be supported.....
I think it's a hacky way to get this information. Am not against users doing it (have done it myself many times), but it shouldn't the official method we provide. We should be providing this through directly wrapped functions.

Joao-Dionisio:
@Opt-Mucca I understand what you're saying (especially about the hacky way), but I don't think this can be done wrapping SCIP functions, unfortunately... What if we called it a convenience function, potentially putting it somewhere else, like we did with the recipes?

src/pyscipopt/scip.pxi
```diff
@@ -5439,6 +5533,76 @@ cdef class Model:
...
+    # Time spent on reading
+    reading_time: float
+    # Time spent on copying
+    copying_time: float
+    # Name of problem
+    problem_name: str
+    # Name of presolved problem
+    presolved_problem_name: str
+    # Dictionary with number of variables by type 
+    _variables: dict
```
mmghannam, line 5553: Why is it private? (goes for the rest of ones prefixed by underscore too. If it's returned by SCIP it should be accessible.

src/pyscipopt/scip.pxi
```diff
@@ -5439,6 +5533,76 @@ cdef class Model:
...
+    # The best dual bound found
+    dual_bound: float
+    # The gap between the primal and dual bounds
+    gap: float
+    # The primal-dual integral 
+    primal_dual_integral: float
+
+    def _populate_remaining(self):
+        # number of variables in the model
+        self.n_vars: int = self._variables["total"]
```
mmghannam, line 5579: These should be computed properties exposed as methods, would make the values more visible also easier to add documentation later.

mmghannam (changes requested):
Great job João! I added some minor comments but looks good overall

src/pyscipopt/scip.pxi
mmghannam: The comment should also say what are the keys and values

Joao-Dionisio: Because of the way .stats is organized, I found it easier to have the variables in a dictionary, which I then unpack in the ```_populate_remaining``` method. Should I just let people access the variables dictionary?

src/pyscipopt/scip.pxi
Joao-Dionisio: Done!

src/pyscipopt/scip.pxi
mmghannam: No it's ok, we can always make it public later. But I don't think it's necessary, probably if more info is added there we can added more computed properties later

src/pyscipopt/scip.pxi
```diff
@@ -5445,6 +5537,155 @@ cdef class Model:
...
+        number of binary variables in the presolved model
+    n_presolved_integer_vars: int
+        number of integer variables in the presolved model
+    n_presolved_implicit_integer_vars: int
+        number of implicit integer variables in the presolved model
+    n_maximal_cons: int
+        number of maximal constraints in the model
+    n_initial_cons: int
+        number of initial constraints in the presolved model
+    n_presolved_maximal_cons
```
mmghannam, line 5605: missing type

src/pyscipopt/scip.pxi
```diff
@@ -5445,6 +5537,155 @@ cdef class Model:
...
+        number of maximal constraints in the presolved model
+    n_presolved_conss
```
mmghannam, line 5607: missing type

src/pyscipopt/scip.pxi
```diff
@@ -5445,6 +5537,155 @@ cdef class Model:
...
+        Name of problem
+    presolved_problem_name: str
+        Name of presolved problem
+    _variables: dict
+        Dictionary with number of variables by type 
+    _presolved_variables: dict
+        Dictionary with number of presolved variables by type 
+    _constraints: dict
+        Dictionary with number of constraints by type
+    _presolved_constraints: dict
```
mmghannam, line 5565: no need to add public documentation for non-public functions, these should probably be normal comments

mmghannam (changes requested):

mmghannam (approved):

PR #860, opened 2024-06-17
Add SCIPprintExternalCodes

Opt-Mucca:
Adds a wrapper for the SCIP functionality to print external libraries (what you see when you run `./scip` from your terminal at the start.). This lets us easily check that everything is packaged correctly inside of the wheel that we release without checking at the intermediate build stage. (Still should check at the intermediate stage, but it is a way for users to now verify what version of IPOPT / nauty they're now using)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 ### Added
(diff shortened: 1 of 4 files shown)


Joao-Dionisio (approved):
Awesome, thank you Mark!

PR #871, opened 2024-06-24
Expand getStatistics to allow unbounded, infeasible, and user-interrupted problems

Joao-Dionisio:
I added default values for some parameters because not all of them exist for every problem (e.g. Primal bound for infeasible problems).


Joao-Dionisio:
Can you take a look at this one, @mmghannam?

mmghannam (approved):
Great, thanks João!

PR #872, opened 2024-07-01
Relax stage for checking getObjVal, getVal

### LINKED ISSUE #847: ```GetObjVal``` can't be called, even though the model has solutions
getObjVal, getVal are too strict in the stage checks. For example, the following code raises an error:

```
from pyscipopt import Model

m = Model()

x = m.addVar()
m.setObjective(x)

m.setParam("limits/solutions", 1)
m.optimize()

print("Stage: ", m.getStage())

print(m.getNSols())
print(m.getObjVal()) # crashes here, even though the model does have solutions
```

```getObjVal``` is a PySCIPOpt-only function, but the SCIP functions it calls don't need the strict requirements that ```getObjVal``` has.

The fix is in PR #815, but I'm having some trouble with it.

### PR DESCRIPTION
**Pull request by Joao-Dionisio**
Fixes #847. See PR #815 for more details.

TLDR: Sometimes the model has solutions and ```getObjVal``` can't be called because the stage requirements are stricter than in SCIP.

### CODE REVIEW

File: src/pyscipopt/scip.pxi [hunk 1]
```diff
@@ -3543,6 +3545,7 @@ cdef class Model:
     def presolve(self):
         """Presolve the problem."""
         PY_SCIP_CALL(SCIPpresolve(self._scip))
+        self._bestSol = Solution.create(self._scip, SCIPgetBestSol(self._scip))
```
Joao-Dionisio, line 3548: this might be the most "dangerous" change, but I don't think it causes problems.

src/pyscipopt/scip.pxi
```diff
@@ -4855,13 +4861,24 @@ cdef class Model:
...
-            raise Warning("method cannot be called before problem is solved")
+
+        if SCIPgetNSols(self._scip) == 0:
+            if self.getStage() != SCIP_STAGE_SOLVING:
+                raise Warning("Without a solution, method can only be called in stage SOLVING.")
+        else:
+            assert self._bestSol.sol != NULL
+            
+            if SCIPsolIsOriginal(self._bestSol.sol):
+                stage_requirement = SCIP_STAGE_PROBLEM
```
Opt-Mucca, line 4875: Would rename to something like `min_stage_requirement`

Opt-Mucca (approved):
I really like this change! Came in ready to disagree, and am now convinced that it is a truly helpful addition. Thank you!

src/pyscipopt/scip.pxi
Opt-Mucca: I don't think it will cause any problems either. We have the same thing after `optimize`, so it should just be `NULL` in the case where it does nothing

PR #873, opened 2024-07-01
Strong Branching

Opt-Mucca:
Opens up code that allows users to use strong branching functionality in SCIP.

I did not end up adding a wrapper for `strongBranchVarWithPropagation`. This would be needed if users ever want to entirely move over SCIP's default branching rule to Python. Currently they would be restricted to vanilla full strong branching.


src/pyscipopt/scip.pxi
```diff
@@ -5533,6 +5556,112 @@ cdef class Model:
...
+    def endStrongBranching(self):
+        """End strong branching. Needs to be called if startStrongBranching was called previously.
+        Between these calls the user can access all strong branching functionality. """
+
+        PY_SCIP_CALL(SCIPendStrongbranch(self._scip))
+
+    def getVarStrongBranchInfoLast(self, Variable var):
+        """Get the results of the last strong branching call on this variable (potentially was called
+        at another node).
+
```
Joao-Dionisio, line 5574: For consistency with the rest of the codebase's sake, add the ```:param Variable var: ....``` to the docstring. I don't dislike having additional info in the docstring (down, up, etc.), but I wonder if they shouldn't be inline comments

src/pyscipopt/scip.pxi
```diff
@@ -5533,6 +5556,112 @@ cdef class Model:
...
+        return down, up, downvalid, upvalid, downinf, upinf, downconflict, upconflict, lperror
+
+    def updateVarPseudoCost(self, Variable var, valdelta, objdelta, weight):
+        """Updates the pseudo costs of the given variable and the global pseudo costs after a change of valdelta
+         in the variable's solution value and resulting change of objdelta in the LP's objective value.
+         Update is ignored if objdelts is infinite. Weight is in range (0, 1], and affects how it updates
+         the global weighted sum.
+
+         :param Variable var: Variable whos psuedo cost will be updated
```
Joao-Dionisio, line 5635: psuedo -> pseudo

I think we're snake case for python variables (so val_delta and obj_delta?). I personally don't mind it staying like this

src/pyscipopt/scip.pxi
```diff
@@ -5533,6 +5556,112 @@ cdef class Model:
         assert isinstance(var, Variable), "The given variable is not a pyvar, but %s" % var.__class__.__name__
         PY_SCIP_CALL(SCIPchgVarBranchPriority(self._scip, var.scip_var, priority))
 
+    def startStrongBranching(self):
+        """Start strong branching. Needs to be called before any strong branching. Must also later end strong branching.
+        If propagation is enabled then strong branching is not done of the LP, but on additionally created nodes (has some overhead)"""
```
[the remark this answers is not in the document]
Opt-Mucca: Changed. Am a big fan of commas too, but I think that one would be too excessive.

src/pyscipopt/scip.pxi
Opt-Mucca: Done. As a reminder: We should talk this Friday about the docstrings

src/pyscipopt/scip.pxi
Opt-Mucca: Changed.

For the snake case, I am a big fan for Python variables. I dislike any other case. I feel like this was never a standard previously though, and all Cython variables are the exact opposite.

tests/test_strong_branching.py
```diff
@@ -0,0 +1,169 @@
...
+                self.scip.updateNodeLowerbound(up_child, up_bounds[best_cand_idx])
+
+        return {"result": SCIP_RESULT.BRANCHED}
+
+
+def create_model():
+    scip = Model()
+    scip.setHeuristics(SCIP_PARAMSETTING.OFF)
+    scip.setSeparating(SCIP_PARAMSETTING.OFF)
+    # scip.setIntParam("presolving/maxrounds", 0)
```
Joao-Dionisio, line 141: This commented code might have been forgotten

Joao-Dionisio:
Damn, Mark, this is super awesome, thank you! I just had those nitpicks, but other than that this all looks good from a cursory glance. I especially like that you made your test a bit like an example!

tests/test_strong_branching.py
Opt-Mucca: Yup. Have removed.

Opt-Mucca:
Thank you! I talked to a friend that complained about this functionality being inaccessible. Thought I'd try and make it possible through Python and easy to understand from the example.

src/pyscipopt/scip.pxi
```diff
@@ -417,6 +417,11 @@ cdef class Column:
         """gets objective value coefficient of a column"""
         return SCIPcolGetObj(self.scip_col)
 
+    def getAge(self):
+        """Gets the age of a column, i.e., the total number of successive times a column was in the LP
```
Joao-Dionisio, line 421: Gets the age of *the* column

src/pyscipopt/scip.pxi
```diff
@@ -1960,6 +1973,16 @@ cdef class Model:
...
 
+    def allColsInLP(self):
+        """checks if all columns, i.e. every variable with non-empty column is present in the LP.
+        This is not True when performing pricing for instance."""
+
+        return SCIPallColsInLP(self._scip)
+
+    # LP Col Methods
+    def getColRedCost(self, Column col):
+        return SCIPgetColRedcost(self._scip, col.scip_col)
```
Joao-Dionisio, line 1984: Docstring missing

src/pyscipopt/scip.pxi
```diff
@@ -5553,6 +5576,118 @@ cdef class Model:
         assert isinstance(var, Variable), "The given variable is not a pyvar, but %s" % var.__class__.__name__
         PY_SCIP_CALL(SCIPchgVarBranchPriority(self._scip, var.scip_var, priority))
 
+    def startStrongBranching(self):
+        """Start strong branching. Needs to be called before any strong branching. Must also later end strong branching.
+        If propagation is enabled then strong branching is not done on the LP, but on additionally created nodes (has some overhead)"""
+
+        PY_SCIP_CALL(SCIPstartStrongbranch(self._scip, False))
```
Joao-Dionisio, line 5583: Is there a reason for not allowing propagation? If there is, then I think the last sentence of the docstring doesn't make sense.

tests/test_strong_branching.py
```diff
@@ -0,0 +1,201 @@
...
+        best_cand_idx = 0
+
+        # Start strong branching and iterate over the branching candidates
+        self.scip.startStrongBranching()
+        for i in range(npriocands):
+
+            # Check the case that the variable has already been strong branched on at this node.
+            # This case occurs when events happen in the node that should be handled immediately.
+            # When processing the node again (because the event did not remove it), there's no need to duplicate work.
+            if self.scip.getVarStrongBranchLastNode(branch_cands[i]) == num_nodes:
```
Joao-Dionisio, line 45: From the comments, I don't understand why we are comparing to the number of nodes instead of the current node number.

tests/test_strong_branching.py
```diff
@@ -0,0 +1,201 @@
...
+
+            # Calculate the gains for each up and down node that strong branching explored
+            if not downinf and downvalid:
+                down_bounds[i] = down
+                downgain = max([down - lpobjval, 0])
+            else:
+                downgain = 0
+            if not upinf and upvalid:
+                up_bounds[i] = up
+                upgain = max([down - lpobjval, 0])
```
Joao-Dionisio, line 76: ```max([up - lpobjval, 0])``` no?

tests/test_strong_branching.py
```diff
@@ -0,0 +1,201 @@
...
+            red_cost_col = self.scip.getColRedCost(cols[0])
+
+            avg_sol = cols[0].getVar().getAvgSol()
+
+        return {"result": SCIP_RESULT.DIDNOTRUN}
+
+
+def create_model():
+    scip = Model()
+    scip.setHeuristics(SCIP_PARAMSETTING.OFF)
```
Joao-Dionisio, line 139: Perhaps justify the need for disabling this?

tests/test_strong_branching.py
```diff
@@ -0,0 +1,201 @@
...
+    u = scip.addVar(vtype="I", lb=-3, ub=99)
+
+    more_vars = []
+    for i in range(100):
+        more_vars.append(scip.addVar(vtype="I", lb=-12, ub=40))
+        scip.addCons(quicksum(v for v in more_vars) <= (40 - i) * quicksum(v for v in more_vars[::2]))
+
+    for i in range(100):
+        more_vars.append(scip.addVar(vtype="I", lb=-52, ub=10))
+        scip.addCons(quicksum(v for v in more_vars[50::2]) <= (40 - i) * quicksum(v for v in more_vars[405::2]))
```
Joao-Dionisio, line 158: Possible typo on ```more_vars[405::2]```. There are only 200 ```more_vars``` right?

tests/test_strong_branching.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+def test_strong_branching():
+    scip = create_model()
+
+    strong_branch_rule = StrongBranchingRule(scip, idempotent=False)
+    scip.includeBranchrule(strong_branch_rule, "strong branch rule", "custom strong branching rule",
+                           priority=10000000, maxdepth=-1, maxbounddist=1)
+
+    scip.optimize()
```
Joao-Dionisio, line 181: Are asserts not needed here? To verify that the objective is correct (we're not cutting off nodes incorrectly, for example).

src/pyscipopt/scip.pxi
Opt-Mucca: Changed

src/pyscipopt/scip.pxi
Opt-Mucca: Added! My bad for missing this.

src/pyscipopt/scip.pxi
Opt-Mucca: I've added a TODO to the docstring. I originally did add the propagation option, but then decided to remove it. I did so because:
- One then needs to add a `bool` argument to `startStrongBranching` and `endStrongBranching` (if not set appropriately then the propagation option raises an error)
- There are many more return arguments of the function when `propagation=True`. This arguments are also useless unless other SCIP functions are then wrapped and made available. 
For these two reasons I just wanted to ignore it for now. I think the majority of users don't need this functionality, and if someone asks for it then we can add it. Or we let them add it themselves.

For the naming conventions: Any opinion on changing the name of the Python wrapper function @mmghannam ? I just couldn't stand the SCIP names for strong branching (found them incredibly unintuitive). I can change them back if needed.

tests/test_strong_branching.py
Opt-Mucca: The number of nodes is the current node number (if there's been a restart or something then only `getNTotalNodes` continues to increase)

tests/test_strong_branching.py
Opt-Mucca: Good spot! Fixed this error

tests/test_strong_branching.py
Opt-Mucca: Added a justification

tests/test_strong_branching.py
Opt-Mucca: This was an oversight by me..... I wonder why it didn't raise an error......
I copied it from `test_branch_probing` and just lowered the number of constraints to make the instance simpler.

Opt-Mucca: Changed to 200

tests/test_strong_branching.py
Opt-Mucca: Hmmmmmmmmmmm. You would calculate the objective of the problem without this and then add an assert statement?

Opt-Mucca: I did that. Added an extra safety layer to the test!

Joao-Dionisio: Yeah, but no need to do the optimization every time, just do it once, get the objective, and then assert it's equal to it

tests/test_strong_branching.py
Joao-Dionisio: Oh, I assumed that SCIP could revisit a prior node during the search. No problem, then

src/pyscipopt/scip.pxi
mmghannam: Sorry for the super late reply. I have thought a bit about this and I think I'm more in favour of keeping the same names as SCIP uses for the following reasons: 

1. SCIP's function names are not really likely to change in the future, so having the same names keeps a direct link with SCIP making it easy to transfer your knowledge to directly work in SCIP or easily use PySCIPOpt if you're already used to the C-API. 
2. Regarding the clarity issue, the new names could be clearer than the old ones but what would be more important in understanding what the function does is having a clear documentation, or maybe in this case a strong branching guide since the functionality spans multiple functions (and a link to it in the docstrings). 

So, I think changing the names of direct "wrapper" methods would do more harm than good.  However, I'm in favour of coming up with new names if the function does something more than a direct call to a SCIP function, and I think we should mark these functions somehow, maybe in the naming style; we could use snake case for these more elaborate methods.

Opt-Mucca: Summary: Agreed with Mo.

Joao-Dionisio (approved):

PR #874, opened 2024-07-02
Adds bipartite graph function

Opt-Mucca:
This branch shouldn't be merged until the strong branch one is complete. A potential user raised concerns about such functionality not being present in PySCIPOpt directly and that he had to use third party software. After thinking about it a bit, it is just a glorified feature retrieval function. So I thought I'd spin up a quick version.


tests/test_bipartite.py
```diff
@@ -0,0 +1,118 @@
...
+    scip.addCons(r1 >= x0)
+    scip.addCons(r2 >= -x0)
+    scip.addCons(y0 == r1 + r2)
+    scip.addCons(t + l + 7 * u <= 300)
+    scip.addCons(t >= quicksum(v for v in more_vars[::3]) - 10 * more_vars[5] + 5 * more_vars[9])
+    scip.addCons(more_vars[3] >= l + 2)
+    scip.addCons(7 <= quicksum(v for v in more_vars[::4]) - x0)
+    scip.addCons(quicksum(v for v in more_vars[::2]) + l <= quicksum(v for v in more_vars[::4]))
+
+    scip.setObjective(t - quicksum(j * v for j, v in enumerate(more_vars[20:-40])))
```
Joao-Dionisio, line 78: I can't look into it in more detail now, Mark, but are these tests quick? 400 integer variables, 400 constraints, how long does it take to solve 4 times? Also, you have ```more_vars[405::2]``` in line 67, but I think there are only 200 more_vars.

Opt-Mucca: Each test takes approximately 1s on my machine. Can make them shorter if we need to and just lower the node limit.

Opt-Mucca: Also: There is no rush on this

Joao-Dionisio: I asked purely because it might have delayed ```pytest .``` too much, but it's not the case. 

As for whether this should or should not be exposed, I think you have much stricter requirements than I do :)

src/pyscipopt/scip.pxi
```diff
@@ -2039,8 +2062,8 @@ cdef class Model:
         return SCIPgetRowObjParallelism(self._scip, row.scip_row)
 
     def getRowParallelism(self, Row row1, Row row2, orthofunc=101):
-        """Returns the degree of parallelism between hyplerplanes. 1 if perfectly parallel, 0 if orthognal.
-        101 in this case is an 'e' (euclidean) in ASCII. The other accpetable input is 100 (d for discrete)."""
+        """Returns the degree of parallelism between hyplerplanes. 1 if perfectly parallel, 0 if orthogonal.
+        101 in this case is an 'e' (euclidean) in ASCII. The other acceptable input is 100 (d for discrete)."""
```
Joao-Dionisio, line 2066: I would also add how this is calculated, degree of parallelism is too vague. The SCIP documentation has "returns the degree of parallelism between the hyperplanes defined by the two row vectors v, w: p = |v*w|/(|v|*|w|); the hyperplanes are parallel, iff p = 1, they are orthogonal, iff p = 0"

src/pyscipopt/scip.pxi
```diff
@@ -5553,10 +5576,371 @@ cdef class Model:
...
+        return score
+
     def getTreesizeEstimation(self):
         """Get an estimation of the final tree size """
         return SCIPgetTreesizeEstimation(self._scip)
 
+
+    def getBipartiteGraphRepresentation(self, prev_col_features=None, prev_edge_features=None, prev_row_features=None,
+                                        static_only=False, suppress_warnings=False):
+        """This function generated the bipartite graph representation of an LP, which was first used in
```
Joao-Dionisio, line 5698: generates instead of generated. Or if you want to say that it generated the representation first used in the paper, I think you need to say something like "... of an LP used in the following paper". Not a grammar expert though, so might be wrong.

tests/test_bipartite.py
```diff
@@ -0,0 +1,118 @@
...
+    u = scip.addVar(vtype="I", lb=-3, ub=99)
+
+    more_vars = []
+    for i in range(100):
+        more_vars.append(scip.addVar(vtype="I", lb=-12, ub=40))
+        scip.addCons(quicksum(v for v in more_vars) <= (40 - i) * quicksum(v for v in more_vars[::2]))
+
+    for i in range(100):
+        more_vars.append(scip.addVar(vtype="I", lb=-52, ub=10))
+        scip.addCons(quicksum(v for v in more_vars[50::2]) <= (40 - i) * quicksum(v for v in more_vars[405::2]))
```
Joao-Dionisio, line 67: Same thing with ```more_vars[405::2]```

Joao-Dionisio (approved):

PR #880, opened 2024-08-05
Add prototype for utils functions

Joao-Dionisio:
When creating tests it can be annoying having to create the same models over and over again. In this ```utils.py``` file, we aggregate some models that might show up more than once. 

Note: There is already an ```util.py``` file in the tests folder, might want to rename it.

### DIFF
--- a/tests/helpers/utils.py
+++ b/tests/helpers/utils.py
@@ -1,0 +1,109 @@
+from pyscipopt import Model, quicksum, SCIP_PARAMSETTING, exp, log, sqrt, sin
+from typing import List
+
+def random_MIP_1():
+    scip = Model()
+    scip.setHeuristics(SCIP_PARAMSETTING.OFF)
+    scip.setSeparating(SCIP_PARAMSETTING.OFF)
+    scip.setLongintParam("limits/nodes", 250)
+    scip.setParam("presolving/maxrestarts", 0)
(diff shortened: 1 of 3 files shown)


tests/helpers/utils.py
```diff
@@ -0,0 +1,109 @@
+from pyscipopt import Model, quicksum, SCIP_PARAMSETTING, exp, log, sqrt, sin
+from typing import List
+
+def random_MIP_1():
```
mmghannam, line 4: the `_1`s look strange when there's no `_2` 😄

Joao-Dionisio: In my defense, it's still a draft :p

mmghannam: Fair enough 😄

Joao-Dionisio: Okay, my defense stopped working @mmghannam :D 

My new defense is that this can be something that we keep updating as we come across or come up with interesting models. What do you think? This PR is something that should only affect the developers.

Opt-Mucca (approved):

Opt-Mucca:
I will merge this. The models look good.
Things still to do in some other MR:
- Call them in the actual tests to remove duplicate code
- Merge `util.py` into `utils.py`

PR #894, opened 2024-09-09
Update README with new docs. Move CONTRIBUTING to docs

Opt-Mucca:
I moved some information from README.md and CONTRIBUTION.md to the online documentation. I also removed duplicate information that I had previously copied over, e.g. FAQs / Gotchas.

I currently have not changed the basic example on the README, and I have left the potentially unnecessary subsection on plug-ins. 

I have also added some deprecation warning to INSTALL.md that we will remove it at some later date and added a recommendation to use the online documentation instead


README.md
```diff
@@ -13,54 +13,20 @@ This project provides an interface from Python to the [SCIP Optimization Suite](
...
 
-**Using Conda**
-
-[![Conda version](https://img.shields.io/conda/vn/conda-forge/pyscipopt?logo=conda-forge)](https://anaconda.org/conda-forge/pyscipopt)
-[![Conda platforms](https://img.shields.io/conda/pn/conda-forge/pyscipopt?logo=conda-forge)](https://anaconda.org/conda-forge/pyscipopt)
-
-***DO NOT USE THE CONDA BASE ENVIRONMENT TO INSTALL PYSCIPOPT.***
-
-Conda will install SCIP automatically, hence everything can be installed in a single command:
+Thr recommended installation method is via PyPI
```
Joao-Dionisio, line 23: The

README.md
```diff
@@ -13,54 +13,20 @@ This project provides an interface from Python to the [SCIP Optimization Suite](
 Documentation
 -------------
 
-Please consult the [online documentation](https://scipopt.github.io/PySCIPOpt/docs/html) or use the `help()` function directly in Python or `?` in IPython/Jupyter.
+Please consult the [online documentation](https://pyscipopt.readthedocs.io/en/latest/) or use the `help()` function directly in Python or `?` in IPython/Jupyter.
```
Joao-Dionisio, line 16: Until we have all the methods and the pop-down menus and stuff, I think we still need to link to the old documentation

Opt-Mucca: I strongly disagree with this. I think the API is only a small fraction of the documentation, and I don't think pop-down menus are even necessary.

Opt-Mucca: It should be said: The old documentation will stay online, so people can continue to use it

Joao-Dionisio: Yeah, but there are some things in the old documentation that seem important. The docstrings, for example, which explain what the arguments of the methods are. This is really the only page I think should be referenced: [Model class reference](https://scipopt.github.io/PySCIPOpt/docs/html/classpyscipopt_1_1scip_1_1Model.html)

Are you against using a sentence like

> We are currently migrating the documentation to [online documentation](https://pyscipopt.readthedocs.io/en/latest/). It should be prefered to the old documentation (can still be found in ...) which will be deprecated.

Opt-Mucca: You're completely correct on the docstrings......
I've added a strong sentence immediately after the current documentation link.

Joao-Dionisio (approved):

PR #895, opened 2024-09-09
Easier way to attach event handlers

mmghannam:


Joao-Dionisio:
Thanks for this, Mo! I think we should also add some tests, just to be safe.

Also, you and Mark are currently in a race to see which of you manages to do commit #2000 :)

mmghannam:
Done 😄 thanks for the quick review!

Joao-Dionisio:
Thanks, Mo! Now all that's missing is the CHANGELOG entry, I think

Joao-Dionisio (approved):

PR #911, opened 2024-10-07
Change theme of documentation

Opt-Mucca:


docs/conf.py
```diff
@@ -75,7 +75,8 @@
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
 #
-html_theme = "sphinx_nefertiti"
+# html_theme = "sphinx_nefertiti"
```
mmghannam, line 78:
```suggestion
```

No need to leave this comment :)

mmghannam (changes requested):

docs/conf.py
Opt-Mucca: Booooooooooooooooooooo. I am having regret because the current one is so beautiful

Joao-Dionisio (approved):

Joao-Dionisio (approved):

mmghannam (approved):

PR #912, opened 2024-10-08
Add sphinxcontrib.jquery for new theme

Opt-Mucca:


Opt-Mucca:
I'm getting a build error on readthedocs for not having this line. Not sure why I can avoid having it locally.

PR #914, opened 2024-10-11
Add comments from dominik

Opt-Mucca:


mmghannam (approved):
Looks good! Thanks @Opt-Mucca and thanks @DominikKamp :)

PR #916, opened 2024-10-18
Primal-dual evolution event handler recipe

Joao-Dionisio:
Based on #881, I think it's helpful to include the answer as a recipe.

It's a bit scuffed, I think one needs to capture the final state of the problem, as the event handler is not called when the solving stops.

I'm also unsure about the best way to separate things, I decided just to include the event handler in the recipe and the user would need to optimize the model themselves, as well as plot the solutions afterward. I'd be happy with better suggestions, though.


src/pyscipopt/recipes/primal_dual_evolution.py
```diff
@@ -0,0 +1,50 @@
...
+                if self.model.isGT(self.model.getDualbound(), self.model.data["dual_solutions"][-1][1]):
+                    self.model.data["dual_solutions"].append((self.model.getSolvingTime(), self.model.getDualbound()))
+            else:
+                if self.model.isLT(self.model.getDualbound(), self.model.data["dual_solutions"][-1][1]):
+                    self.model.data["dual_solutions"].append((self.model.getSolvingTime(), self.model.getDualbound()))
+
+    if not hasattr(model, "data"):
+        model.data = {}
+
+    model.data["primal_solutions"] = []
```
mmghannam, line 31: maybe dual_log and primal_log are better names for what it is?

src/pyscipopt/recipes/primal_dual_evolution.py
```diff
@@ -0,0 +1,50 @@
+from pyscipopt import Model, Eventhdlr, SCIP_EVENTTYPE, Eventhdlr
+
+def get_primal_dual_evolution(model: Model):
+
+    class gapEventhdlr(Eventhdlr):
```
mmghannam, line 5:
```suggestion
    class GapEventhdlr(Eventhdlr):
```

tests/helpers/utils.py
```diff
@@ -77,19 +77,15 @@ def random_nlp_1():
     return model
 
 
-def knapsack_model(weights=[4, 2, 6, 3, 7, 5], costs=[7, 2, 5, 4, 3, 4]):
+def knapsack_model(weights=[4, 2, 6, 3, 7, 5], costs=[7, 2, 5, 4, 3, 4], knapsackSize = 15):
```
mmghannam, line 80:
```suggestion
def knapsack_model(weights=[4, 2, 6, 3, 7, 5], costs=[7, 2, 5, 4, 3, 4], knapsack_size = 15):
```

tests/test_recipe_primal_dual_evolution.py
```diff
@@ -0,0 +1,36 @@
...
+    
+    for i in range(1, len(model.data["dual_solutions"])):
+        if model.getObjectiveSense() == "minimize":
+            assert model.data["dual_solutions"][i][1] >= model.data["dual_solutions"][i-1][1]
+        else:
+            assert model.data["dual_solutions"][i][1] <= model.data["dual_solutions"][i-1][1]
+
+    # how to get a simple plot of the data
+    #from pyscipopt.recipes.primal_dual_evolution import plot_primal_dual_evolution
+    #plot_primal_dual_evolution(model)
```
mmghannam, line 36: these should be removed right? 😄

src/pyscipopt/recipes/primal_dual_evolution.py
```diff
@@ -0,0 +1,50 @@
...
+        model.data = {}
+
+    model.data["primal_solutions"] = []
+    model.data["dual_solutions"] = []
+    hdlr = gapEventhdlr()
+    model.includeEventhdlr(hdlr, "gapEventHandler", "Event handler which collects primal and dual solution evolution")
+
+    return model
+
+def plot_primal_dual_evolution(model: Model):
```
mmghannam, line 38: Maybe this method is unnecessary, one should provide the data not assume what it would be used for. I feel this plotting is more for an example

src/pyscipopt/recipes/primal_dual_evolution.py
```diff
@@ -0,0 +1,50 @@
+from pyscipopt import Model, Eventhdlr, SCIP_EVENTTYPE, Eventhdlr
+
+def get_primal_dual_evolution(model: Model):
```
mmghannam, line 3: maybe the name should be `attach_primal_dual_evolution_eventhdlr`? longer but more descriptive of what it is

src/pyscipopt/recipes/primal_dual_evolution.py
```diff
@@ -0,0 +1,50 @@
...
+
```
mmghannam, line 4: I think we need a good docstring here. Describing most importantly where the data is saved and how one would read it

mmghannam:
> Based on #881, I think it's helpful to include the answer as a recipe.
> 
> It's a bit scuffed, I think one needs to capture the final state of the problem, as the event handler is not called when the solving stops.

you can capture the final state using the `exitsol` callback. 

> I'm also unsure about the best way to separate things, I decided just to include the event handler in the recipe and the user would need to optimize the model themselves, as well as plot the solutions afterward. I'd be happy with better suggestions, though.

I think in recipes is ok but without the plotting and with good documentation :)  

and thanks a lot João!

tests/test_recipe_primal_dual_evolution.py
Joao-Dionisio: Ah yeah, I wanted to give an example of use. Can it go on the doctstring?

src/pyscipopt/recipes/primal_dual_evolution.py
Joao-Dionisio: What do you think of this, Mo?

>     """
    Attaches an event handler to a given SCIP model that collects primal and dual solutions,
    along with the solving time when they were found.
    The data is saved in model.data["primal_log"] and model.data["dual_log"]. They consist of
    a list of tuples, each tuple containing the solving time and the corresponding solution.
    A usage example can be found in examples/finished/plot_primal_dual_evolution.py. The
    example takes the information provided by this recipe and uses it to plot the evolution
    of the dual and primal bounds over time. 
    """

Joao-Dionisio:
Thanks for the suggestions, Mo!

I ended up copying the gas transportation model for the plotting example because importing stuff with python from different folders is somewhat finicky. I could have added a couple of lines using sys, but I feel like it would potentially complicate things for inexperienced users.

Also Mo and @Opt-Mucca, what do you think of creating an issue asking users to give us some of their pyscipopt models so that we can expand on the utils?

tests/helpers/utils.py
```diff
@@ -1,14 +1,6 @@
 from pyscipopt import Model, quicksum, SCIP_PARAMSETTING, exp, log, sqrt, sin
 from typing import List
 
-from pyscipopt.scip import is_memory_freed
-
-
-def is_optimized_mode():
```
Opt-Mucca, line 7: Why are we removing this test?

tests/test_heur.py
```diff
@@ -106,7 +106,7 @@ def test_heur():
     assert round(sol[y]) == 0.0
 
 def test_heur_memory():
-    if is_optimized_mode():
+    if is_memory_freed():
```
Opt-Mucca, line 109: These should not be completely equivalent?

Opt-Mucca:
@Joao-Dionisio I strongly support asking users to contribute models! That would be super helpful for our testing.

For this MR: I like the recipe and the example's fine (even if it does have a lot of duplicate code). My issue is with the changes in the tests, and I'm a bit curios from the design POV.

src/pyscipopt/recipes/primal_dual_evolution.py
```diff
@@ -0,0 +1,60 @@
...
+            self.model.catchEvent(SCIP_EVENTTYPE.LPSOLVED, self)
+            self.model.catchEvent(SCIP_EVENTTYPE.NODESOLVED, self)
+            
+
+        def eventexec(self, event):
+            # if a new best primal solution was found, we save when it was found and also its objective
+            if event.getType() == SCIP_EVENTTYPE.BESTSOLFOUND:
+                self.model.data["primal_log"].append((self.model.getSolvingTime(), self.model.getPrimalbound()))
+            
+            if not self.model.data["dual_log"]:
```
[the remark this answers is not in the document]
Joao-Dionisio: I'm not sure I understand the comment, but Python evaluates `0`, empty list, `None`, as `False` in conditions. So I'm just checking if the dual log is empty, since otherwise I will want to compare the current value with the previous one.

src/pyscipopt/recipes/primal_dual_evolution.py
```diff
@@ -0,0 +1,60 @@
...
+            
+            if self.model.getObjectiveSense() == "minimize":
+                if self.model.isGT(self.model.getDualbound(), self.model.data["dual_log"][-1][1]):
+                    self.model.data["dual_log"].append((self.model.getSolvingTime(), self.model.getDualbound()))
+            else:
+                if self.model.isLT(self.model.getDualbound(), self.model.data["dual_log"][-1][1]):
+                    self.model.data["dual_log"].append((self.model.getSolvingTime(), self.model.getDualbound()))
+
+
+    if not hasattr(model, "data"):
```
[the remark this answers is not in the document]
Joao-Dionisio: That's a good point, but I'm not super sure what the best approach is. Personally, I like having a data dictionary in my model to store everything, so that it's easily accessible everywhere, but I understand that having things tidier might be preferable.

tests/helpers/utils.py
Joao-Dionisio: It annoyed me having this test in utils.py, since at the time the test seemed useless. I only realized after I read your comment that just constructing the model might somehow alter the result of `is_memory_free`. Still, I would prefer to have this test in a different place.

src/pyscipopt/recipes/primal_dual_evolution.py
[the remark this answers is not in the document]
Opt-Mucca: This was just a question out of curiosity. I'm not against storing it in the `Model` object. I just think it's tidier and more intuitive to be in the event handler (this does not seem to be universal though). Happy with the current state.

tests/helpers/utils.py
Opt-Mucca: All for it being moved if you want, but I don't see if it actually being copied somewhere else.

PR #917, opened 2024-10-18
Add _createConsBasicLinear

Joao-Dionisio:
Added _createConsBasicLinear based on #910. 

Also added stage check for freeTransform, freeReoptSolve. I don't know if there's a better way to do this than to do it manually for every method.

Tried to add transformProb. No compilation errors, but the test seemed like it didn't do anything, I'm not sure what's going on.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 ### Added
+- Added stage checks to presolve, freereoptsolve, freetransform
 - Added primal_dual_evolution recipe and a plot recipe
 ### Fixed
 ### Changed
--- a/src/pyscipopt/scip.pxi
+++ b/src/pyscipopt/scip.pxi
@@ -2057,6 +2057,15 @@ cdef class Model:
     def freeTransform(self):
         """Frees all solution process data including presolving and
         transformed problem, only original problem is kept."""
+        if self.getStage() not in [SCIP_STAGE_INIT,
+                                 SCIP_STAGE_PROBLEM,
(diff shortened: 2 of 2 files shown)


Opt-Mucca:
Things I'm for in this MR:
- Adding stage checks to the functions. I think there is no better way to do this than what you've done.
- The transform functionality. What is the difference between this and calling presolve however? I would need to this distinction made clear.

Things I don't like:
- Why do users need to add basic linear constraint? Shouldn't this be done with `_createConsLinear`? The basic include is supposed to save users implementing certain callbacks. This is easy in Python though and doesn't take up much code, and I think it's more confusing to provide this functionality.

Joao-Dionisio:
Hey @Opt-Mucca, thanks for looking into this.

After `transformProb` the problem is in stage `TRANSFORMED`, so presolving wasn't called yet. I'm not entirely sure what presolving does after the problem is transformed though, but agreed that we should make it clear.  

The idea for adding `_createConsBasicLinear` was because it might help out in other projects. There is this [ML4LB repo](https://github.com/pandat8/ML4LB?tab=readme-ov-file#installation) that asks users to install PySCIPOpt and then compile their own version with these added methods. I'm not entirely sure that this is actually needed (they use it [here](https://github.com/pandat8/ML4LB/blob/001839ace3506c8410a30d1f4d3188a3cd95e2dd/localbranching.py#L787)), but it's still an annoyance for the users, as described in #910.

To be fair, in the documentation we do have this line, but we might want to remove it:
> Ideally speaking, we want every SCIP function to be wrapped in PySCIPOpt.

PR #926, opened 2024-11-27
Allow writeProblem to write to stdout

Joao-Dionisio:
If `writeProblem` is called with `filename=None`, the model is printed to the standard output.

Also added `verbosity` option (default True), because it can be annoying to many messages when you copy many models.

EDIT: I'm defaulting it to `.cip`. We can add an option for the extension, though, but we'd need to change `writeProblem` as well.

EDIT2: Something going on with the MacOS integration tests.


DominikKamp:
Could this be renamed to `printProblem()` to make it consistent to `printSol()` (there are no methods starting with display yet)?

Joao-Dionisio:
> Could this be renamed to `printProblem()` to make it consistent with `printSol()` (there are no methods starting with display yet)?

It can yeah, the idea was that in the SCIP terminal one would use `display problem` for the same thing. I have no strong opinion either way, do you have a preference @DominikKamp ?

DominikKamp:
I just mean that to achieve the same as `display solution` we have to call `printBestSol()` in the presence of `writeBestSol()`, so it would feel more natural if we can `display problem` by `printProblem()` in the presence of `writeProblem()`, unless you plan to rename all print methods.

Opt-Mucca:
Happy with the change!

Opt-Mucca (approved):

Issue #896, opened 2024-09-10
If compile SCIPOpt suite using git repository of every projects, output does not come out to console in real time


adover134:
**Describe the bug**
https://github.com/scipopt/PySCIPOpt/issues/887#issue-2505564663
This is extension from the link.

I installed Ubuntu server 24.04.
I downloaded SCIPsuite tarball from the following link.
https://www.scipopt.org/index.php#download

I unzipped tarball, then removed gcg, soplex, papilo, zimpl, scip folders.
Then I cloned up-to-date version of each folders from git hub.

I installed Ipopt by cloning from git hub.
This is because MUMPS was required by PySCIPOpt.

After installing Ipopt, I installed SCIPsuite with TPI=tny option.
Then I installed PySCIPOpt by cloning git repository.

After successful(?) installation, I tried to solve my model.
However, when I was not using redirectOutput(), the console did not showed output in real time.

**To Reproduce**
I used master branch of each projects given by scipopt.

**Expected behavior**
The output should be printed on terminal (console) in real time while not using redirectOutput().

**Screenshots**
![image](https://github.com/user-attachments/assets/fe79f287-8d7e-4481-9d13-aa2902d58e39)

**System**
 - OS: Ubuntu
 - Version: 24.04.1 LTS
 - SCIP version: 10.0.0
 - How did you install `pyscipopt`? cloned from git repository

**Additional context**
I used `m.setParam('display/dualbound/active', 0)`.
However, it worked as `m.setParam('display/nnodes/active', 0)` and also not showing completion percentage.
This one is more easy to fix, right?

Joao-Dionisio:
Hey, @adover134! Wait, is the output not showing in your screenshot? Or is that you only showing the desired output? In your code snippet you are using `redirectOutput`.

adover134:
If not use `redirectOutput()`, then The console does not show something; just save the file per some minutes. I cannot take the picture of it right now, because I am running the model.
The purpose of uploading screenshot is, I wanted to show you that the output does not shows number of nodes and completion percentage but showing dualbounds when I deactivated only showing dualbound.

Joao-Dionisio:
I understand you not wanting to share your code on a public forum, but would it be possible to invite me to a google colab or something, where I would be able to run the code? My email is up201606210[at]up[dot]pt

adover134:
[...]

PR #929, opened 2024-11-29
Correct output redirection

### LINKED ISSUE #243: "write" methods don't work after `redirectOutput()`
After calling `redirectOutput()`, e.g. `writeProblem()` does not produce a file but prints the problem description to the outstream.

Apparently, we redirect *everything* to the new messagehandler.

DominikKamp:
This resolves #243 by restricting method `redirectOutput()` to the streams `stdout` and `stderr`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Added stage checks to presolve, freereoptsolve, freetransform
 - Added primal_dual_evolution recipe and a plot recipe
 ### Fixed
+- Only redirect stdout and stderr in redirectOutput() so that file output still works afterwards
 ### Changed
 ### Removed
 
--- a/src/pyscipopt/scip.pxi
+++ b/src/pyscipopt/scip.pxi
@@ -12,7 +12,7 @@ cimport cython
 from cpython cimport Py_INCREF, Py_DECREF
 from cpython.pycapsule cimport PyCapsule_New, PyCapsule_IsValid, PyCapsule_GetPointer
 from libc.stdlib cimport malloc, free
-from libc.stdio cimport fdopen, fclose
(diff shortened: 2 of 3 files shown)


Joao-Dionisio:
Hey hey, @DominikKamp, glad to see you here :smile: 

We had some problems with printing and people having weird locales [#830], [#843]. Do you think we might need to be concerned with this here as well?

Also, can you please add some tests?

Joao-Dionisio:
Yeah, the output itself should be hard to test, but just that it's being printed to the correct place. We have a bunch of `test_XX.py` files, so it's mostly a matter of seeing in which it makes more sense, defaulting to [`test_model.py`](https://github.com/scipopt/PySCIPOpt/blob/6876af9db7fc3eb3d224c030ba8b53975fee2b20/tests/test_model.py). I think it makes sense to add it here. There's also the possibility of creating a new test file, if you think it is justified (it also needs to be prefixed by `test_`).

The test function needs to be prefixed by `test_`, and you can use some models we have stored in `helpers/utils.py`, in case you don't want to build a new one.

DominikKamp:
Tests added but please verify that the console output is still as expected with and without `redirectOutput()`.

DominikKamp:
Before doing anything else, it seems like the mac pipeline has to be fixed because it seems like the github runner is not able to compile the SCIPOptSuite before eternity ends (why not use the precompiled version like in the Windows-test?).

Opt-Mucca:
Changes look good. I will test before merging.

Opt-Mucca:
This does fix the issue on `writeProblem`. It also seems to help with #896

Opt-Mucca (approved):

Opt-Mucca:
Let's wait until #933 is merged as then the pipelines can run faster

PR #930, opened 2024-11-30
Generic indicator names

Joao-Dionisio:
Indicator constraints all had the same name by default. They now have the same default naming convention as other constraints. 

I did discover a big bug that I'll fix later. When a model has indicator constraints and is optimized, all the constraints disappear. Not good. The tests are failing because of the assertion I added detecting this. EDIT: I'm very dumb, by default `getNConss` returns the number of constraints of the transformed problem.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Added stage checks to presolve, freereoptsolve, freetransform
 - Added primal_dual_evolution recipe and a plot recipe
 ### Fixed
+- Added default names to indicator constraints
 ### Changed
 - GitHub actions using Mac now use precompiled SCIP from latest release
 ### Removed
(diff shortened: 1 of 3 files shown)


Opt-Mucca:
Two things:
- This branch was built on your older branch that you changed (I assume). All the display problem stuff should be removed.
- You shouldn't have a default name for the indicator cons. I like the two line change on changing the name if `name==""` though and think it's needed

Joao-Dionisio:
@Opt-Mucca yep, I made an oopsie when branching.

As for your second point, the indicator constraints now have the same behavior as normal constraints - default `name = ""`, and changing it to `name = 'c'+str(SCIPgetNConss(self._scip)+1)` if this is the case. You are saying that the name should be a default argument? also for the constraints?

Opt-Mucca:
It is not the same as normal constraints. You have changed the default name to "indicatorCons", which will bypass your change to `cX` as a default name.

Joao-Dionisio:
@Opt-Mucca I don't think so, I think I changed the default name _from_ `indicatorCons`, which was there before, to `""`.

image

Opt-Mucca (approved):

PR #933, opened 2024-12-03
Updates mac script to use precompiled version

Opt-Mucca:


Opt-Mucca:
This should make our pipelines run faster. It just downloads the precompiled SCIP from Stefan's GitHub release for MacOS

Joao-Dionisio (approved):

PR #934, opened 2024-12-09
Add transformed option to getVarDict

Joao-Dionisio:
Allows `getVarDict` to be used on the transformed problem.

I don't know which should be the default value for the `transformed` flag, as this doesn't seem consistent in the rest of the source code. `getVars`, `getObjVal`, `getSolObjVal` default to original the original problem, `getConss`, `getNConss`, `getNVars` to the transformed problem. I don't know if we should change them, though, because it would have big consequences.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 ### Added
+- Added transformed option to getVarDict, updated test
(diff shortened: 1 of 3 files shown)


Opt-Mucca:
Happy with the changes!

Opt-Mucca (approved):

PR #935, opened 2024-12-09
Added option for getting Lhs, Rhs of nonlinear constraints

Joao-Dionisio:
Did as the title says, plus removed some mentions of quadratic constraints, as PySCIPOpt now covers more general constraints (following the open todo [here](https://github.com/orgs/scipopt/projects/3?pane=issue&itemId=61794879)).

Some of the documentation was wrong (`getSlack`, `getActivity`), since it incorrectly said some methods could be used with nonlinear constraints. I could adapt these functions to also cover nonlinear constraints, but it might be a bit hacky compared with what's going on on the linear side.

Draft because I don't know what should be done with these two methods.


Opt-Mucca:
I like all the changes!

For the reason for it being draft: I'd ask Stefan how this is in general handled on the non-linear side for activity. I couldn't find a single function in my 30s search, but I imagine that there is one. That can be a separate MR though.

Opt-Mucca (approved):

PR #939, opened 2024-12-11
Fixed categorical data example #932

Jiunixo:
Adding constraints so that each employee must be assigned to exactly one shift.


Joao-Dionisio (approved):

Joao-Dionisio:
Thank you so much, @Jiunixo!

PR #940, opened 2024-12-16
add support for [get|set]HeurTiming

pfetsch:
We needed to control the timing of heuristics. Since this cannot be changed via parameters, I added an API support her.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 ### Added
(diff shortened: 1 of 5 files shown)


pfetsch:
I added a test and tried to extend the documentation. Better?

Joao-Dionisio (approved):

pfetsch:
@Joao-Dionisio Thanks for the fast review.

PR #947, opened 2025-01-24
Add type to loop variables only

Joao-Dionisio:
Similar to #946, but without adding types to the arguments, as it was controversial.

We should probably also add a performance test at some point. Naturally, SCIP should be where the majority of time is spent, but maybe someone is doing a lot of stuff on the Python side.

The performance gain of this is probably marginal, but faster is better than slower,  I suppose.


src/pyscipopt/benders.pxi
```diff
@@ -175,6 +175,7 @@ cdef SCIP_RETCODE PyBendersSolvesub (SCIP* scip, SCIP_BENDERS* benders, SCIP_SOL
 cdef SCIP_RETCODE PyBendersPostsolve (SCIP* scip, SCIP_BENDERS* benders, SCIP_SOL* sol,
         SCIP_BENDERSENFOTYPE type, int* mergecands, int npriomergecands, int nmergecands, SCIP_Bool checkint,
         SCIP_Bool infeasible, SCIP_Bool* merged) noexcept with gil:
+    cdef int i
```
DominikKamp, line 178: Please move index declarations consistently to the back of a cdef block.

src/pyscipopt/lp.pxi
```diff
@@ -63,6 +63,7 @@ cdef class LP:
         lb      -- lower bound (default 0.0)
         ub      -- upper bound (default infinity)
         """
+        cdef int i
         nnonz = len(entries)
```
DominikKamp, line 67: This must then also be an `int`.

src/pyscipopt/scip.pxi
```diff
@@ -981,6 +984,7 @@ cdef class Solution:
 
     def __getitem__(self, Expr expr):
         # fast track for Variable
+        cdef SCIP_Real coeff
```
DominikKamp, line 987: Maybe add a new line after each cdef block.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 ### Added
+- Added cdef to loop variables for slight speedup
```
DominikKamp, line 5:
```suggestion
- Added cdef type declaration of loop variables for slight speedup
```

src/pyscipopt/scip.pxi
Joao-Dionisio: Did this, but it does look a bit unnatural in some places

src/pyscipopt/lp.pxi
Joao-Dionisio: I can do this (did so for the examples you pointed out), but I think there's no gain from Cython's POV.

src/pyscipopt/benders.pxi
Joao-Dionisio: Done.

src/pyscipopt/conshdlr.pxi
```diff
@@ -168,6 +168,8 @@ cdef SCIP_RETCODE PyConsFree (SCIP* scip, SCIP_CONSHDLR* conshdlr) noexcept with
     return SCIP_OKAY
 
 cdef SCIP_RETCODE PyConsInit (SCIP* scip, SCIP_CONSHDLR* conshdlr, SCIP_CONS** conss, int nconss) noexcept with gil:
+    cdef int i
+    
     PyConshdlr = getPyConshdlr(conshdlr)
     cdef constraints = []
```
DominikKamp, line 174: What does this mean?

src/pyscipopt/lp.pxi
```diff
@@ -63,14 +63,15 @@ cdef class LP:
         lb      -- lower bound (default 0.0)
         ub      -- upper bound (default infinity)
         """
-        nnonz = len(entries)
+        cdef int nnonz = len(entries)
```
DominikKamp, line 67:
```suggestion
```

src/pyscipopt/lp.pxi
```diff
@@ -95,9 +96,10 @@ cdef class LP:
         lbs   -- lower bounds (default 0.0)
         ubs   -- upper bounds (default infinity)
         """
-
-        ncols = len(entrieslist)
-        nnonz = sum(len(entries) for entries in entrieslist)
+        
+        cdef ncols = len(entrieslist)
```
DominikKamp, line 100:
```suggestion
        cdef int ncols = len(entrieslist)
```

src/pyscipopt/lp.pxi
```diff
@@ -95,9 +96,10 @@ cdef class LP:
...
+        cdef nnonz = sum(len(entries) for entries in entrieslist)
+        cdef int i
```
DominikKamp, line 103:
```suggestion
```

src/pyscipopt/reader.pxi
```diff
@@ -46,8 +46,11 @@ cdef SCIP_RETCODE PyReaderWrite (SCIP* scip, SCIP_READER* reader, FILE* file,
                                  SCIP_CONS** conss, int nconss, int maxnconss, int startnconss,
                                  SCIP_Bool genericnames, SCIP_RESULT* result) noexcept with gil:
     cdef SCIP_READERDATA* readerdata
+    cdef int i 
+
     readerdata = SCIPreaderGetData(reader)
     cdef int fd = fileno(file)
+
```
DominikKamp, line 53:
```suggestion
    cdef int fd = fileno(file)
    cdef int i

    readerdata = SCIPreaderGetData(reader)
```

src/pyscipopt/scip.pxi
DominikKamp: If the function block is not divided by newlines already, this can also be omitted.

src/pyscipopt/scip.pxi
```diff
@@ -1280,10 +1291,14 @@ cdef class Node:
 
         """
         cdef int addedconsssize = SCIPnodeGetNAddedConss(self.scip_node)
+        cdef int i
+
         if addedconsssize == 0:
             return []
+
         cdef SCIP_CONS** addedconss = <SCIP_CONS**> malloc(addedconsssize * sizeof(SCIP_CONS*))
         cdef int nconss
```
DominikKamp, line 1300:
```suggestion

        if addedconsssize == 0:
            return []

- 

        cdef SCIP_CONS** addedconss = <SCIP_CONS**> malloc(addedconsssize * sizeof(SCIP_CONS*))
        cdef int nconss
        cdef int i
```

src/pyscipopt/scip.pxi
```diff
@@ -1356,6 +1372,8 @@ cdef class Node:
 
         """
         cdef int nbranchvars = self.getNParentBranchings()
+        cdef int i
```
DominikKamp, line 1375: Please move this below boundtypes.

src/pyscipopt/scip.pxi
```diff
@@ -4705,6 +4740,9 @@ cdef class Model:
             The created and added Constraint objects.
 
         """
+        cdef int idx 
+        cdef int i
```
DominikKamp, line 4744: Please reuse `i` also here.

src/pyscipopt/scip.pxi
```diff
@@ -4880,6 +4920,7 @@ cdef class Model:
         """
         cdef SCIP_Bool success
         cdef int _nvars
```
DominikKamp, line 4922: Why use `_nvars` instead of `nvars` here? Why below not? Could this be made consistent as well?

src/pyscipopt/lp.pxi
DominikKamp: Depends on how `range` works. For me it would be just more consistent. At some point it needs to be converted and maybe it is better to do it only once.

src/pyscipopt/scip.pxi
DominikKamp: If the scope allows, `nvars`. At some spots `_nvars` seems to be not even used afterwards.

Joao-Dionisio: I suppose the same argument goes for `_vars`.

Joao-Dionisio: I think everything should be done now, @DominikKamp

src/pyscipopt/scip.pxi
DominikKamp: I have adapted the newlines.

src/pyscipopt/scip.pxi
DominikKamp: Alright, I have proposed some extension.

DominikKamp:
@Joao-Dionisio From my side this would be fine but some pyscipopt tests seem to fail although there is a green button so please have a look.

Joao-Dionisio:
@DominikKamp yeah, everything seems to be fine, you just need to approve the PR (I can't do it because I'm the one who created it).

DominikKamp (approved):

PR #953, opened 2025-02-10
Add node cutoff

### LINKED ISSUE #949: Questions about Implementing a Node Heuristic in PySCIPOpt
Dear PySCIPOpt Developers,

I am currently working on a heuristic that is applied at a node in a Branch & Bound process. The goal of this heuristic is to determine a lower bound for the node. If this lower bound is greater than the best known feasible solution, I would like to remove this node from the Branch & Bound tree.

I have a few questions regarding how to implement this in PySCIPOpt:

    What is the best way to retrieve the current variable bounds at a given node?
    Can I directly prune the node myself after computing the lower bound?
    If direct pruning is not possible, is there a way to pass the computed lower bound to SCIP so that it removes the node accordingly?

I appreciate your support and look forward to your response.

Best regards,
Mario

Joao-Dionisio:
Should fix #949

### DIFF
--- a/tests/test_node.py
+++ b/tests/test_node.py
@@ -1,0 +1,20 @@
+from pyscipopt import SCIP_RESULT, Eventhdlr, SCIP_EVENTTYPE
+from helpers.utils import random_mip_1
+
+class cutoffEventHdlr(Eventhdlr):
+    def eventinit(self):
+        self.model.catchEvent(SCIP_EVENTTYPE.NODEFOCUSED, self)
+
+    def eventexec(self, event):
+        self.model.cutoffNode(self.model.getCurrentNode())
+        return {'result': SCIP_RESULT.SUCCESS}
+
+def test_cutoffNode():
+    m = random_mip_1()
+    hdlr = cutoffEventHdlr()
+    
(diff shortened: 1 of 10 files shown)


tests/test_node.py
```diff
@@ -0,0 +1,20 @@
...
+
+def test_cutoffNode():
+    m = random_mip_1()
+    hdlr = cutoffEventHdlr()
+    
+    m.includeEventhdlr(hdlr, "test", "test")
+
+    m.optimize()
+
+    assert m.getNSols() == 0
```
mmghannam, line 20: This assertion seems a bit flaky. If heuristics could run before the node is focused, this would not work. Maybe disable heuristics, and presolving. You could maybe instead also check the number of LP iterations, it should be zero I think in this case.

Joao-Dionisio: Maybe I can be a bit more explicit, @mmghannam, but `random_MIP_1()` already disables heuristics, presolving, and separators.

mmghannam: Thank you!

mmghannam (approved):

PR #957, opened 2025-02-16
New matrix variable functionality

Joao-Dionisio:
## What is this?

This PR will track the implementation of the matrix variable functionality. With the current changes, users can do the following:

```python

from pyscipopt import Model, sin, log
import numpy as np

m = Model()
matrix_variable1 = m.addMatrixVar( shape=(3,3), vtype="B", name="test", obj=np.ones((3,3)) )

obj = np.asarray([ [2,1,3], [1,1,0], [1,2,3] ])
matrix_variable2 = m.addMatrixVar( shape=(3,3), vtype="C", name="", obj=obj )
# # note that depending on the argument type, what you pass to the function may be broadcasted by numpy (intentional decision on our part)
# # e.g., passing vtype="C" makes every variable have type "C", but obj is taking an iterable, and so every variable will be assigned its own value

m.addMatrixCons( log(log(matrix_variable1)) - sin(matrix_variable2) >= 32 ) # same here. This means every entry should be >= 32

m.optimize() # should be running without errors
``` 

## Motivation
In some optimization sub-fields and other adjacent fields, it is sometimes standard to have a matrix view of the problem. For example, in machine learning one often gets a matrix, does some things to it, and passes it on.
It is also something that general modeling frameworks tend to support, so adding this functionality should make integrating PySCIPOpt into them easier, thus possibly increasing the visibility of PySCIPOpt.

## Status
It is now possible to add matrix variables, create expression involving matrix varaibles, and create linear and general nonlinear expressions involving matrix variables. So, in a sense, most of the basic functionality is already here. What is yet missing is proper documentation, more thorough testing (especially on the nonlinear side), and a focus on performance optimization when it comes to building *big* models. We also want to allow ```addCons``` to accept matrix variables, as some users might find it more intuitive.

---

Please let us know your thoughts and suggestions regarding this functionality.

This is joint work by @Opt-Mucca, @mmghannam, and @Joao-Dionisio.


docs/tutorials/matrix.rst
```diff
@@ -0,0 +1,133 @@
...
+and there to purely understand the API):
+
+.. code-block:: python
+
+    x = scip.addMatrixVar(shape=(2, 2), vtype="B", name="x")
+    y = scip.addMatrixVar(shape=(2, 2), vtype="C", name="y", ub=5)
+    z = scip.addVar(vtype="C", name="z", ub=7)
+
+    scip.addMatrixCons(x + y <= z)
+    scip.addMatrixCons(exp(x) + sin(sqrt(y)) == z + y)
```
mmghannam, line 82: Should mention here that the function is applied element wise.

docs/tutorials/matrix.rst
```diff
@@ -0,0 +1,133 @@
...
+.. code-block:: python
+
+    x = scip.addMatrixVar(shape=(2, 2))
+    assert(isinstance(x, MatrixVariable))
+    assert(isinstance(x[0][0], Variable))
+    cons = x <= 2
+    assert(isinstance(cons, MatrixConstraint))
+    assert(isinstance(cons[0][0]), Constraint)
+
+
```
mmghannam, line 132: We should add a section here about querying solutions using matrix variables.

src/pyscipopt/scip.pxi
```diff
@@ -3132,6 +3548,123 @@ cdef class Model:
...
+            is the matrix variable a pricing candidate? (Default value = False)
+        pricedVarScore : float or np.ndarray, optional
+            score of matrix variable in case it is priced, the higher the better (Default value = 1.0)
+
+        Returns
+        -------
+        MatrixVariable
+
+        """
+        # assert has_numpy, "Numpy is not installed. Please install numpy to use matrix variables."
```
mmghannam, line 3589:
```suggestion
```

src/pyscipopt/scip.pxi
```diff
@@ -3132,6 +3548,123 @@ cdef class Model:
...
+            ndim = len(shape)
+
+        matrix_variable = np.empty(shape, dtype=object)
+
+        if isinstance(name, str):
+            matrix_names = np.full(shape, name, dtype=object)
+            if name != "":
+                # with np.nditer(matrix_names, flags=["multi_index", "refs_ok"], op_flags=["writeonly"]) as it:
+                #   for x in it:
+                #        x[...] = f"{name}_{'_'.join(map(str, it.multi_index))}"
```
mmghannam, line 3618: Do we need this?

src/pyscipopt/scip.pxi
```diff
@@ -3132,6 +3548,123 @@ cdef class Model:
...
+            matrix_variable[idx] = self.addVar(name=matrix_names[idx], vtype=matrix_vtypes[idx], lb=matrix_lbs[idx],
+                                               ub=matrix_ubs[idx], obj=matrix_objs[idx], pricedVar=matrix_priced_vars[idx],
+                                               pricedVarScore=matrix_priced_var_scores[idx])
+
+        # with np.nditer(matrix_names, flags=["multi_index"], op_flags=["writeonly"]) as it:
+        #     for x in it:
+        #        x[...] = self.addVar(name=matrix_names[idx], vtype=matrix_vtypes[idx], lb=matrix_lbs[idx],
+        #                             ub=matrix_ubs[idx], obj= matrix_objs[idx], pricedVar=matrix_priced_vars[idx],
+        #                             pricedVarScore=matrix_priced_var_scores[idx])
+
```
mmghannam, line 3664: this too?

src/pyscipopt/scip.pxi
```diff
@@ -4811,6 +5347,156 @@ cdef class Model:
 
         return constraints
 
+    def addMatrixCons(self, cons: MatrixExprCons, name: Union[str, np.ndarray] ='',
```
mmghannam, line 5350: Please split each argument on its own line.

docs/tutorials/matrix.rst
Joao-Dionisio: As an FYI, `getBestSol` returns the solution in terms of the single variables, not the matrix variables. This is to be expected, but is probably something we want to convert back. Would need to think a bit, but if people are putting MatrixVariables into the model, they will expect MatrixVariables back.

Opt-Mucca: I'm confused. `getBestSol` should return a `Solution` object? We changed it so a user can now do `sol[matrix_var]` and get a `np.ndarray` of solution values returned. What else do we need to change?

Opt-Mucca: Ahhhhhhhhhhh. This is to do with the documentation. I can add.

docs/tutorials/matrix.rst
```diff
@@ -81,8 +81,9 @@ and there to purely understand the API):
 
     scip.addMatrixCons(x + y <= z)
     scip.addMatrixCons(exp(x) + sin(sqrt(y)) == z + y)
-    scip.addMatrixCons(y <= x @ y <= x)
-    scip.addMatrixCons(x.sum() <= 2)
+    scip.addMatrixCons(y <= x @ y)
+    scip.addMatrixCons(x @ y <= x)
+    scip.addCons(x.sum() <= 2) # Matrix variables can also appear in standard constraints, if the result expression as type Expr
```
Opt-Mucca, line 86: This comment is incorrect? Or at least I don't understand it. I will remove it.

docs/tutorials/matrix.rst
Opt-Mucca: Added

docs/tutorials/matrix.rst
Opt-Mucca: Ahhhhhhh. Gotcha. Will leave it as is.

Opt-Mucca (approved):

docs/tutorials/matrix.rst
mmghannam: *is type Expr

docs/tutorials/matrix.rst
```diff
@@ -0,0 +1,148 @@
...
+
+    x = scip.addMatrixVar(shape=(2, 2), vtype="B", name="x")
+    y = scip.addMatrixVar(shape=(2, 2), vtype="C", name="y", ub=5)
+    z = scip.addVar(vtype="C", name="z", ub=7)
+
+    scip.addMatrixCons(x + y <= z)
+    scip.addMatrixCons(exp(x) + sin(sqrt(y)) == z + y)
+    scip.addMatrixCons(y <= x @ y)
+    scip.addMatrixCons(x @ y <= x)
+    scip.addCons(x.sum() <= 2) # Matrix variables can also appear in standard constraints, if the result expression as type Expr
```
mmghannam, line 86:
```suggestion
    scip.addCons(x.sum() <= 2) # Matrix variables can also appear in standard constraints, if the result expression is type Expr
```

PR #961, opened 2025-03-06
Add getLinearConsIndicator

Opt-Mucca:


Joao-Dionisio:
So, using this on $z=1 \implies ax \leq b$ returns $ax \leq b$, right?

In the SCIP documentation, there's a `(may be NULL)`, should we say something about it?

PR #965, opened 2025-03-19
add parameter set and get functions for lpi

lidingxu:
Sometimes an LP is solved through pyscipopt, and SCIP has several parameters for LPI,  the pull request add support to get and set these parameters. The types of parameters are encoded in an enum, the set and get function accept them as arguments.

### DIFF
--- a/tests/test_lp.py
+++ b/tests/test_lp.py
@@ -11,10 +11,8 @@ def test_lp():
     defaultLPParPricing = myLP.getIntParam(SCIP_LPPARAM.PRICING)
     defaultLPParLpinfo = myLP.getIntParam(SCIP_LPPARAM.LPINFO)
     defaultLPParLpitlim = myLP.getIntParam(SCIP_LPPARAM.LPITLIM)
-    # defaultLPParFastmip = myLP.getIntParam(SCIP_LPPARAM.FASTMIP)
     defaultLPParFeastol = myLP.getRealParam(SCIP_LPPARAM.FEASTOL)
     defaultLPParDualfeastol = myLP.getRealParam(SCIP_LPPARAM.DUALFEASTOL)
-    # defaultLPParBarrierconvtol = myLP.getRealParam(SCIP_LPPARAM.BARRIERCONVTOL)
     defaultLPParObjlim = myLP.getRealParam(SCIP_LPPARAM.OBJLIM)
     defaultLPParLptilim = myLP.getRealParam(SCIP_LPPARAM.LPTILIM)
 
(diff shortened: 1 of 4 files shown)


tests/test_lp.py
```diff
@@ -1,9 +1,41 @@
...
     # create LP instance, minimizing by default
     myLP = LP()
 
+    # get default int parameters.
+    lpParFromScratch = myLP.getIntParam(SCIP_LPPARAM.FROMSCRATCH)
+    lpParScaling = myLP.getIntParam(SCIP_LPPARAM.SCALING)
+    lpParPricing = myLP.getIntParam(SCIP_LPPARAM.PRICING)
+    lpParLpinfo = myLP.getIntParam(SCIP_LPPARAM.LPINFO)
+    lpParLpitlim = myLP.getIntParam(SCIP_LPPARAM.LPITLIM)
+    # lpParFastmip = myLP.getIntParam(SCIP_LPPARAM.FASTMIP)
```
Joao-Dionisio, line 14: Why are some lines commented? Also, can you please add some asserts? Just tiny things, that setting an integer/real parameter is actually doing what it's supposed to, same for the getter functions

Joao-Dionisio (commented):
Thank you so much, Liding! And thank you for removing white space from `scip.pxi, scip.pxd` :)

Just have a minor remark on the tests, otherwise it looks good!

src/pyscipopt/lp.pxi
```diff
@@ -476,3 +476,76 @@ cdef class LP:
...
+        float
+
+        """
+        cdef SCIP_Real value
+
+        PY_SCIP_CALL(SCIPlpiGetRealpar(self.lpi, param, &value))
+
+        return value
+
+    def getIntParam(self, param):
```
Joao-Dionisio, line 532: Can you bring this one before `getRealParam`? So that it is

```
setInt
setReal
getInt
getReal
```

tests/test_lp.py
lidingxu: some parameters are solver specific, see the new file.

Joao-Dionisio: Is it possible to detect if the solver is present? Something like

```
if specific_solver:
  do solver specific thing
```

Otherwise, I would just remove them, because they'll never get tested.

Opt-Mucca:
Two things for this:

- Why are all the functions on the python side missing `lpi` in their definitions? `setIntParam` would make more sense as `setLpiIntParam` or something along those lines (maybe play around with the capitalisation)
- Do we actually want this to be available to the user? What use cases are we trying to accommodate here? In general SCIP should be protecting users from having to play around with the LPI.

Joao-Dionisio:
Regarding the names, it's mostly mirroring the SCIP names, I don't know if confusion would arise. So a user would do

```
from pyscipopt import Model, LP

m = Model()
lp = LP()
m.setIntParam(...)
lp.setIntParam(...)
```

But we're already changing `SCIPlpiSetIntPar` into `setIntParam`, so maybe you're right, it wouldn't hurt.


As for the benefit of having the LPI, I'm not really sure, but `lp.pxi` has been there from the near beginning. Maybe it makes LP research easier somehow, I've never used it.

lidingxu:
Users always call lp.setIntParam(), call lp.lpiSetIntParam() is a bit redundant. The functions defined in lp.pxi also has no lpiXXXX(). Sometimes, a user just want to solve an LP using lightweighted LPI, and currently, we have no parameters for controlling it directly,

Joao-Dionisio (approved):
Yeah, looks good Liding! At some point I'll add a test for all the numerical comparisons, but not needed for now.

What do you think, @Opt-Mucca ?

Opt-Mucca:
My bad on this. I didn't look closely enough. Also simply wasn't aware we even supported this functionality...... No blocker from me now.

PR #970, opened 2025-03-25
Fixed bug when accessing matrix variable attributes

### LINKED ISSUE #969: Matrix API easy access to LocalBounds
Hello Pyscipopt Team,

Is your feature request related to a problem? Please describe.
I'm currently unable to access the individual bounds of variables that I created using the Matrix API. While setting the bounds during variable creation is very convenient, reading them afterwards is not as straightforward.

Describe the solution you'd like
When I create variables using the Matrix API, I can set their bounds like this:

`x = scip.addMatrixVar(shape, vtype='C', name='x', ub=np.array([[5, 6], [2, 8]]))`

It would be great if I could also retrieve the bounds in a similar matrix-style format, for example:

`ub = x.getUbLocal()  # or a similar method`

Additional context
This feature would make it easier to inspect and debug models that rely heavily on matrix-based variable definitions.

Thank you all for your great work and for considering this request!

Joao-Dionisio:
Fix #969 

There's still a bug with `getCol`, but that one is also a bug in the normal variables. I'm thinking it may just be about adding an assertion.

@DominikKamp do you know when `SCIPvarGetCol` returns NULL?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Added SCIP_LPPARAM, setIntParam, setRealParam, getIntParam, getRealParam, isOptimal, getObjVal, getRedcost for lpi
 - Added isFeasPositive
 ### Fixed
+- Fixed bug when accessing matrix variable attributes
 ### Changed
 ### Removed
 
--- a/src/pyscipopt/scip.pxi
+++ b/src/pyscipopt/scip.pxi
@@ -1685,7 +1685,7 @@ class MatrixVariable(MatrixExpr):
 
         """
         vtypes = np.empty(self.shape, dtype=object)
-        for idx in np.ndindex(self):
+        for idx in np.ndindex(self.shape):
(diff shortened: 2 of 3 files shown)


DominikKamp:
Ideally, if there is no column to the variable for example if the variable is deleted already or no LP created yet.

Joao-Dionisio:
@DominikKamp would it make sense to add a precondition to the method, even though it takes in `SCIP_VAR*` and not `SCIP*`? Or is returning NULL something to be expect? In the latter case, something should be said in the documentation, IMO.

DominikKamp:
Maybe you meant to check the upper bounds of the columns after optimizing but you might need to disable presolving before.

DominikKamp:
`NULL` is okay but there is also an assertion that it must be a column, which would fail before.

mmghannam (approved):
Looks good, thanks @Joao-Dionisio!

PR #973, opened 2025-04-14
Add deactivatePricer

Joao-Dionisio:
This is useful if one wants to go for the Price and Branch heuristic in the middle of pricing

### DIFF
--- a/src/pyscipopt/scip.pxi
+++ b/src/pyscipopt/scip.pxi
@@ -4157,6 +4157,34 @@ cdef class Model:
         """
         return Node.create(SCIPgetBestChild(self._scip))
 
+    def getSiblings(self):
+        """
+        Gets the siblings of the focus node.
+
+        Returns
+        -------
+        list of Nodes
+
+        """
+        cdef SCIP_NODE** _siblings
+        cdef int n_siblings
+        cdef int i
+
+        PY_SCIP_CALL(SCIPgetSiblings(self._scip, &_siblings, &n_siblings))
+
+        return [Node.create(_siblings[i]) for i in range(n_siblings)]
+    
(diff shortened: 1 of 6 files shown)


tests/test_pricer.py
```diff
@@ -124,6 +128,7 @@ def test_cuttingstock():
     pricer.data['rollLength'] = rollLength
     pricer.data['patterns'] = patterns
     pricer.data['redcosts'] = []
+    pricer.data["deactivate"] = False
```
mmghannam, line 131: You could deactivate the price directly here instead of holding the data. Would that cause an issue?

Joao-Dionisio: I wanted to test deactivating the pricer inside the pricer. Deactivating it before it doing anything feels a bit of a lame test

src/pyscipopt/scip.pxi
```diff
@@ -7593,6 +7593,7 @@ cdef class Model:
         PY_SCIP_CALL(SCIPactivatePricer(self._scip, scip_pricer))
         pricer.model = <Model>weakref.proxy(self)
         Py_INCREF(pricer)
+        pricer._name = name
```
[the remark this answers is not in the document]
mmghannam: Why not get the pricer's name when you need it, using `SCIPpricerGetName`?

tests/test_pricer.py
mmghannam: ok cool

src/pyscipopt/scip.pxi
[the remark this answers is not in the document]
Joao-Dionisio: not my finest moment...

mmghannam (approved):

PR #979, opened 2025-04-22
Don't run tests on draft PRs

Joao-Dionisio:


Joao-Dionisio:
Well, there's the test-coverage (3.11) guy over there that seems to be waiting on the other test, but I'm happy with this for now.

PR #981, opened 2025-04-26
Add support for indicator constraints

Joao-Dionisio:
Missing tests and a bunch of `create/add generic indicator etc.` methods.

Didn't try to compile. Also, should we interface things like `setLinearConsIndicator`, `setBinaryVarIndicator`, when things are done automatically when we call `addConsIndicator`?


Joao-Dionisio:

PR #982, opened 2025-04-26
Add support for cardinality constraints

Joao-Dionisio:
Did not compile, still need to make tests. Need to check if `NULL, NULL` is the correct behavior in `addVarCardinality`.


Joao-Dionisio:

PR #984, opened 2025-04-26
Add support for SOS1 constraints

Joao-Dionisio:
Missing tests and did not compile. Also added methods to the variable class directly.


Joao-Dionisio:

PR #986, opened 2025-05-05
Add access functions for row duals

chrhansk:
Could you please add those and ideally release a new version? I need those functions for a project.


Joao-Dionisio:
Thank you, @chrhansk! As for the new release, let me just wait on a review for #973, I think it should be done by the end of the week.

Joao-Dionisio (approved):

Joao-Dionisio:
@chrhansk the release is still not done, but you can try out the test release with

`pip install --index-url https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple/ PySCIPOpt==5.5.0`

Actually, can you please first try `pip install -i https://test.pypi.org/simple/ PySCIPOpt` and see whether it install v5.5.0 or v5.3.0? Wondering if it's a problem with my machine.

chrhansk:
Thanks, that works. The second command does however install v5.3.0, I also need to specify the version as 5.5.0

PR #988, opened 2025-05-06
Fix master compatibility

DominikKamp:
This makes PySCIPOpt work for latest master branch on Mac with Homebrew.

### DIFF
--- a/src/pyscipopt/scip.pxi
+++ b/src/pyscipopt/scip.pxi
@@ -2339,7 +2339,7 @@ cdef class Model:
         """
         if self.getMajorVersion() < MAJOR:
             raise Exception("linked SCIP is not compatible to this version of PySCIPOpt - use at least version", MAJOR)
-        if self.getMinorVersion() < MINOR:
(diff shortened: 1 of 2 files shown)


src/pyscipopt/scip.pxi
```diff
@@ -2339,7 +2339,7 @@ cdef class Model:
         """
         if self.getMajorVersion() < MAJOR:
             raise Exception("linked SCIP is not compatible to this version of PySCIPOpt - use at least version", MAJOR)
-        if self.getMinorVersion() < MINOR:
+        if self.getMajorVersion() == MAJOR and self.getMinorVersion() < MINOR:
```
Joao-Dionisio, line 2342: Oh, this was a pretty dumb error, thank you!

Joao-Dionisio:
It looks like it makes sense, @DominikKamp. But how would one test this? It's something that will impact everyone, that's why I'm a bit concerned about sending it untested.

DominikKamp:
It affects Linux and Mac systems where there could be a brewed GMP. So better check this location as well. I have tested the updated branch and the tests seem to pass, otherwise you can just blame it on me.

PR #993, opened 2025-05-14
Fixes potentially outdated value of getVal

mmghannam:


Joao-Dionisio:
Can you add a test that fails without this, please?

I'm a bit worried about scripts that call `getVal` very often, though. Sure, it's a tiny tiny additional effort, people using things like `getVarDict`  on big models might feel the difference. We would be creating a `Solution` object for every variable.

Can we somehow circumvent this? A terrible idea I just had was to attach an even handler to every model, that catches `BESTSOLFOUND` and the `execmethod` calls `getBestSol`.

Joao-Dionisio:
image

This is cracking me up. If not raise error, then raise error.
We were the ones who added this, Mo ahah

mmghannam:
> Can you add a test that fails without this, please?
> 
> I'm a bit worried about scripts that call `getVal` very often, though. Sure, it's a tiny tiny additional effort, but people using things like `getVarDict` on big models might feel the difference. We would be creating a `Solution` object for every variable.
> 
> Can we somehow circumvent this? A terrible idea I just had was to attach an even handler to every model, that catches `BESTSOLFOUND` and the `execmethod` calls `getBestSol`.

We could maybe prevent it in the solving stage, and raise and error guiding the user to the other methods.

Joao-Dionisio:
what was it that we decided was the better way to deal with this @mmghannam ?

marouane-f:
`if not stage_check or self._bestSol.sol == NULL and SCIPgetStage(self._scip) != SCIP_STAGE_SOLVING:`

The check allows to call `getVal` during solving, if a solution is available. Is this intended behavior?

Joao-Dionisio:
@mmghannam I remember we had a discussion about this, with a better way to do this. Do you remember what it was?

 I'm not too fond of preventing it in the solving stage, since the intended behavior seems clear, and crashing the user's code seems somewhat frustrating.

mmghannam:
I unfortunately don't remember. I would also lean on the side of allowing it in the solving stage. Regarding the eventhandler idea, this would add an overhead to all models, which is not so nice. But either this solution or the eventhandler, any overhead is worth not returning an outdated value.

Joao-Dionisio:
Came up with a happy compromise. Before creating a new Solution object, check whether it's needed.
Some experiments showed that the original fix was 30% slower, while this is only 7% slower.

Joao-Dionisio:
Would rather have a test for which the original implementation failed. @marouane-f do you have a MRE with you?
If not, then we can just merge as well.

PR #1005, opened 2025-05-28
Require binary variables in logical constraints

Joao-Dionisio:
Fix 1004

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Added isPositive(), isNegative(), isFeasLE(), isFeasLT(), isFeasGE(), isFeasGT(), isHugeValue(), and tests
 - Added SCIP_LOCKTYPE, addVarLocksType(), getNLocksDown(), getNLocksUp(), getNLocksDownType(), getNLocksUpType(), and tests
 ### Fixed
+- Fixed segmentation fault when logical constraints were used with expressions rather than variables.
 ### Changed
 ### Removed
 
--- a/src/pyscipopt/scip.pxi
+++ b/src/pyscipopt/scip.pxi
@@ -6308,6 +6308,9 @@ cdef class Model:
 
         _resvar = (<Variable>resvar).scip_var
         for i, var in enumerate(vars):
+            if not isinstance(var, Variable):
+                raise TypeError("Expected Variable, got %s." % type(var))
+
             _vars[i] = (<Variable>var).scip_var
 
         if name == '':
@@ -6373,6 +6376,9 @@ cdef class Model:
 
         _resvar = (<Variable>resvar).scip_var
         for i, var in enumerate(vars):
+            if not isinstance(var, Variable):
+                raise TypeError("Expected Variable, got %s." % type(var))
+
             _vars[i] = (<Variable>var).scip_var
 
         if name == '':
@@ -6437,6 +6443,9 @@ cdef class Model:
 
         assert type(rhsvar) is type(bool()), "Provide BOOLEAN value as rhsvar, you gave %s." % type(rhsvar)
         for i, var in enumerate(vars):
+            if not isinstance(var, Variable):
+                raise TypeError("Expected Variable, got %s." % type(var))
+
             _vars[i] = (<Variable>var).scip_var
 
         if name == '':
@@ -8763,7 +8772,7 @@ cdef class Model:
         return Node.create(downchild), Node.create(eqchild), Node.create(upchild)
 
 
-    def branchVarVal(self, variable, value):
+    def branchVarVal(self, Variable variable, value):
         """
         Branches on variable using a value which separates the domain of the variable.
 
--- a/tests/test_cons.py
+++ b/tests/test_cons.py
@@ -72,6 +72,25 @@ def test_cons_logical():
     assert m.isEQ(m.getVal(result1), 1)
     assert m.isEQ(m.getVal(result2), 0)
 
+@pytest.mark.xfail(
+    reason="Logical constraints expect lists of binary variables"
+)
+def test_cons_logical_fail():
+    m = Model()
+    x1 = m.addVar(vtype="B")
+    x2 = m.addVar(vtype="B")
+    x3 = m.addVar(vtype="B")
+    x4 = m.addVar(vtype="B")
+    result1 = m.addVar(vtype="B")
+
+    m.addCons(x3 == 1 - x1)
+    m.addCons(x4 == 1 - x2)
+
+    # result1 true
+    m.addConsOr([x1*x3, x2*x4], result1)
+
+    m.optimize()
+
 def test_SOScons():
     m = Model()
     x = {}


src/pyscipopt/scip.pxi
```diff
@@ -6308,6 +6308,9 @@ cdef class Model:
 
         _resvar = (<Variable>resvar).scip_var
         for i, var in enumerate(vars):
+            if not isinstance(var, Variable):
+                raise TypeError("Expected Variable, got %s." % type(var))
```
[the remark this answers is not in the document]
DominikKamp: Wrong and not the task of PySCIPOpt.

src/pyscipopt/scip.pxi
```diff
@@ -6308,6 +6308,9 @@ cdef class Model:
 
         _resvar = (<Variable>resvar).scip_var
         for i, var in enumerate(vars):
+            if not isinstance(var, Variable):
```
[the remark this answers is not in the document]
DominikKamp: Maybe rather refactor the whole conversion into a helper function.

DominikKamp:
The suppressed suggestions seem to make more sense than the unsuppressed suggestions.

src/pyscipopt/scip.pxi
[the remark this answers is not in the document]
DominikKamp: One could indeed additionally add a function for checking a single variable since the same needs to be done for the resultant.

src/pyscipopt/scip.pxi
```diff
@@ -2397,6 +2397,18 @@ cdef void relayErrorMessage(void *messagehdlr, FILE *file, const char *msg) noex
             fputs(msg, file)
         fflush(file)
 
+cdef class VarArrayWrapper:
```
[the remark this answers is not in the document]
Joao-Dionisio: Like this?

```python
def __cinit__(self, object vars):
        cdef int size = len(vars)
        self.ptr = <SCIP_VAR**> malloc(size * sizeof(SCIP_VAR*))

        for i, var in enumerate(vars):
            if not isinstance(var, Variable):
                raise TypeError("Expected Variable, got %s." % type(var))

            self.ptr[i] = (<Variable>var).scip_var
```

src/pyscipopt/scip.pxi
```diff
@@ -2397,6 +2397,30 @@ cdef void relayErrorMessage(void *messagehdlr, FILE *file, const char *msg) noex
...
+cdef class _VarArrayWrapper:
+    cdef SCIP_VAR** ptr
+    cdef int size
+
+    def __cinit__(self, object vars):
+        cdef int size = len(vars)
+        self.ptr = <SCIP_VAR**> malloc(size * sizeof(SCIP_VAR*))
+
+        if size == 1:
+            if not isinstance(vars, Variable):
```
DominikKamp, line 2409: What is the length of a variable?

src/pyscipopt/scip.pxi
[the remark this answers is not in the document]
DominikKamp: Since it contains a Cython type, should this not be private?

src/pyscipopt/scip.pxi
Joao-Dionisio: Yeah, was too hasty with the push, should be working now.

src/pyscipopt/scip.pxi
```diff
@@ -2397,6 +2397,33 @@ cdef void relayErrorMessage(void *messagehdlr, FILE *file, const char *msg) noex
             fputs(msg, file)
         fflush(file)
 
+cdef class _VarArray:
+    cdef SCIP_VAR** ptr
+    cdef int size
+
+    def __cinit__(self, object vars):
+        cdef int size
+
```
DominikKamp, line 2406:
```suggestion
```

src/pyscipopt/scip.pxi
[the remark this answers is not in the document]
Joao-Dionisio: So, something like `[2+x, 3*y]`? Since it's an internal function, I say that we can always easily get `[x,y]`

src/pyscipopt/scip.pxi
```diff
@@ -8233,15 +8268,16 @@ cdef class Model:
 		solution : Solution
 			The corresponding solution in the main model
         """
-        cdef SCIP_VAR** vars = <SCIP_VAR**> malloc(self.getNVars() * sizeof(SCIP_VAR*))
+        cdef SCIP_VAR** vars = <SCIP_VAR**> malloc(sub_model.getNVars() * sizeof(SCIP_VAR*))
```
Joao-Dionisio, line 8276: This was actually another bug that was hidden in here

src/pyscipopt/scip.pxi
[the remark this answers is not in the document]
DominikKamp: I mean `[x, y]` where x and y are expressions, nevermind.

src/pyscipopt/scip.pxi
DominikKamp: Yes, it should at least be asserted that the number of subvariables is at least the number of main variables. Normally, a variable map is maintained so that the main SCIP gets the correct values, but here it is probably assumed that the variable sequence is the same. However, if the subscip is still in solving, `getNVars()` might return the number of transformed subvariables, so that we can consider fewer variables here. Actually, this allocation should be removed as the pointer is overwritten below. This could mean that there is a pipeline to detect double free but not to detect memory leaks.

CHANGELOG.md
```diff
@@ -5,7 +5,9 @@
 - Added support for knapsack constraints
 - Added isPositive(), isNegative(), isFeasLE(), isFeasLT(), isFeasGE(), isFeasGT(), isHugeValue(), and tests
 - Added SCIP_LOCKTYPE, addVarLocksType(), getNLocksDown(), getNLocksUp(), getNLocksDownType(), getNLocksUpType(), and tests
+- Added internal class _VarArray to get the scip variables from PySCIPOpt variables
```
DominikKamp, line 8: Since it is internal, consider removing it from the changelog.

CHANGELOG.md
```diff
@@ -5,7 +5,9 @@
...
 ### Fixed
+- Fixed segmentation fault when logical constraints were used with expressions rather than variables.
```
DominikKamp, line 10: The segmentation fault is not really fixed, rather an exception is raised instead.

src/pyscipopt/scip.pxi
```diff
@@ -2397,6 +2397,39 @@ cdef void relayErrorMessage(void *messagehdlr, FILE *file, const char *msg) noex
...
+                self.ptr = NULL
+            else:
+                self.ptr = <SCIP_VAR**> malloc(size * sizeof(SCIP_VAR*))
+
+                for i, var in enumerate(vars):
+                    print(i)
+                    if not isinstance(var, Variable):
+                        raise TypeError("Expected Variable, got %s." % type(var))
+
+                    self.ptr[i] = (<Variable>var).scip_var
```
DominikKamp, line 2427:
```suggestion
                    self.ptr[i] = var.scip_var
```

src/pyscipopt/scip.pxi
```diff
@@ -2397,6 +2397,39 @@ cdef void relayErrorMessage(void *messagehdlr, FILE *file, const char *msg) noex
...
+
+            if not isinstance(vars, (list, tuple)):
+                raise TypeError("Expected Variable or list of Variable, got %s." % type(vars))
+
+            size = len(vars)
+            print(size)
+            if size == 0:
+                self.ptr = NULL
+            else:
+                self.ptr = <SCIP_VAR**> malloc(size * sizeof(SCIP_VAR*))
```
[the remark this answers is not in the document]
Joao-Dionisio: Why do you prefer to use `self.size`?

DominikKamp:
Please replace also the other occurrences of variable array creations. For consistency make use of the size inside the introduces VarArray class. Fix the memory leak in `translateSubSol()`. Also add an assertion at least on the number of variables. Otherwise, this looks fine.

src/pyscipopt/scip.pxi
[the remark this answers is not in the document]
Joao-Dionisio: I reverted these changes this afternoon because the code doesn't compile with them. Will try to fix it for a bit.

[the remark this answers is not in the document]
DominikKamp: To define it in the variable array object.

src/pyscipopt/scip.pxi
DominikKamp: Why is `scip_var` not declared in `Variable`?

DominikKamp: I mean that `var.scip_var` should work here. `var` is a `Variable` and `var.scip_var` is a `SCIP_VAR*`. I believe this becomes an error because there is no `cdef` member declaration of `scip_var` in the `Variable` class so that it is just considered some `object`.

src/pyscipopt/scip.pxi
DominikKamp: Now there are more memory leaks (do not call `malloc()` when overwriting the returned memory pointer by another allocated array).

src/pyscipopt/scip.pxi
```diff
@@ -8233,21 +8268,22 @@ cdef class Model:
...
         """
-        cdef SCIP_VAR** vars = <SCIP_VAR**> malloc(self.getNVars() * sizeof(SCIP_VAR*))
         cdef SCIP_SOL* real_sol
         cdef SCIP_SOL* subscip_sol = sol.sol
         cdef SCIP_HEUR* _heur
         cdef SCIP_Bool success
         cdef int i
+        cdef _VarArray wrapper
+        
+        assert sub_model.getNVars(False) >= self.getNVars(False), "The sub_model must have at least as many variables as the main model."
```
DominikKamp, line 8278:
```suggestion
        assert sub_model.getNVars(False) >= self.getNVars(True), "The sub_model must have at least as many variables as the main model."
```

src/pyscipopt/scip.pxi
Joao-Dionisio: You meanthe problem is with `wrapper = _VarArray(sub_model.getVars())`? What's the proper way to fix this? After this one, I think it should be merged

DominikKamp: Remove the `malloc()`s.

src/pyscipopt/scip.pxi
DominikKamp: Does it work within cdef (for the `__cinit__` of `VarArray`)?

src/pyscipopt/scip.pxi
Joao-Dionisio: Ahh, I thought you were still talking about `translateSubSol`. So, if I understood correctly, the problem comes when we do 

```python

cdef SCIP_VAR** _vars = malloc(...)

for i in ...:
    vars[i] = ....
 
cdef _VarArray wrapper = _VarArray(vars)
_vars = wrapper.ptr
```

because even though we free the allocated memory inside _VarArray, we don't free the original memory, even if we were to call `free(_vars)`, since it would be a double free. Should be alright now

DominikKamp:
There are also other test files where this landed in the source, could you check?

Joao-Dionisio:
Yeah, had one in the numerical tests, thank you! Okay, so everything is done now, right?

DominikKamp:
Well, now that the logic seems fixed, the code review can begin...

src/pyscipopt/scip.pxi
```diff
@@ -1042,17 +1042,18 @@ cdef class Solution:
 
         # fast track for Variable
         cdef SCIP_Real coeff
+        cdef _VarArray wrapper
+
```
DominikKamp, line 1046:
```suggestion

```

src/pyscipopt/scip.pxi
```diff
@@ -1042,17 +1042,18 @@ cdef class Solution:
...
-            return SCIPgetSolVal(self.scip, self.sol, var.scip_var)
+            return SCIPgetSolVal(self.scip, self.sol, _VarArray(expr).ptr[0])
         return sum(self._evaluate(term)*coeff for term, coeff in expr.terms.items() if coeff != 0)
 
     def _evaluate(self, term):
         self._checkStage("SCIPgetSolVal")
         result = 1
         for var in term.vartuple:
-            result *= SCIPgetSolVal(self.scip, self.sol, (<Variable> var).scip_var)
+            result *= SCIPgetSolVal(self.scip, self.sol, _VarArray(var).ptr[0])
```
DominikKamp, line 1056: There was a warning at another location that this inline variable access is bad because the garbage collector might have freed the memory before. I do not see an urgent need to check `var` to be a `Variable` as it is an internal function. So the previous version should be fine.

DominikKamp: Everywhere.

tests/test_bipartite.py
```diff
@@ -40,8 +40,6 @@ def branchexeclp(self, allowaddcons):
 
 
 
-
-
```
DominikKamp, line 44: Is there any spacing convention?

src/pyscipopt/scip.pxi
```diff
@@ -5249,7 +5278,9 @@ cdef class Model:
         for i, (term, coef) in enumerate(terms.items()):
             termvars = <SCIP_VAR**> malloc(len(term) * sizeof(SCIP_VAR*))
             for j, var in enumerate(term):
-                termvars[j] = (<Variable>var).scip_var
+                wrapper = _VarArray(var)
+                termvars[j] = wrapper.ptr[0]
+
             PY_SCIP_CALL( SCIPcreateExprMonomial(self._scip, &monomials[i], <int>len(term), termvars, NULL, NULL, NULL) )
             termcoefs[i] = <SCIP_Real>coef
             free(termvars)
```
DominikKamp, line 5286:
```suggestion
            wrapper = _VarArray(term)
            PY_SCIP_CALL( SCIPcreateExprMonomial(self._scip, &monomials[i], wrapper.size, wrapper.ptr, NULL, NULL, NULL) )
            termcoefs[i] = <SCIP_Real>coef
```
Make more use of `size`.

src/pyscipopt/scip.pxi
```diff
@@ -5332,8 +5364,9 @@ cdef class Model:
             if opidx == Operator.varidx:
                 assert len(node[1]) == 1
                 pyvar = node[1][0] # for vars we store the actual var!
-                PY_SCIP_CALL( SCIPcreateExprVar(self._scip, &scipexprs[i], (<Variable>pyvar).scip_var, NULL, NULL) )
-                vars[varpos] = (<Variable>pyvar).scip_var
+                wrapper = _VarArray(pyvar)
+                vars[varpos] = wrapper.ptr[0]
+                PY_SCIP_CALL( SCIPcreateExprVar(self._scip, &scipexprs[i], vars[varpos], NULL, NULL) )
                 varpos += 1
```
DominikKamp, line 5370: Please remove unnecessary `pyvar`, `vars`, and `varpos`.

src/pyscipopt/scip.pxi
```diff
@@ -6104,14 +6137,16 @@ cdef class Model:
...
         assert nvars == len(weights), "Number of variables and weights must be the same."
 
         if name == '':
             name = 'c'+str(SCIPgetNConss(self._scip)+1)
 
         for i in range(nvars):
-            vars_array[i] = (<Variable>vars[i]).scip_var
+            wrapper = _VarArray(vars[i])
+            vars_array[i] = wrapper.ptr[0]
             weights_array[i] = <SCIP_Longint>weights[i]
```
DominikKamp, line 6150: Wrap the whole variable array and replace `vars_array` by `wrapper.ptr`.

src/pyscipopt/scip.pxi
```diff
@@ -6366,14 +6396,15 @@ cdef class Model:
...
         cdef SCIP_CONS* scip_cons
         cdef int i
+        cdef _VarArray wrapper = _VarArray(vars)
+        cdef _VarArray resvar_wrapper = _VarArray(resvar)
 
-        _resvar = (<Variable>resvar).scip_var
-        for i, var in enumerate(vars):
-            _vars[i] = (<Variable>var).scip_var
+        _resvar = resvar_wrapper.ptr[0]
+        _vars = wrapper.ptr
```
DominikKamp, line 6407: Make code consistent to `addConsAnd()`.

src/pyscipopt/scip.pxi
```diff
@@ -6430,14 +6459,14 @@ cdef class Model:
...
-        cdef SCIP_VAR** _vars = <SCIP_VAR**> malloc(nvars * sizeof(SCIP_VAR*))
+        cdef SCIP_VAR** _vars
         cdef SCIP_CONS* scip_cons
         cdef int i
-
-        assert type(rhsvar) is type(bool()), "Provide BOOLEAN value as rhsvar, you gave %s." % type(rhsvar)
-        for i, var in enumerate(vars):
-            _vars[i] = (<Variable>var).scip_var
+        cdef _VarArray wrapper = _VarArray(vars)
+        _vars = wrapper.ptr
```
DominikKamp, line 6469:
```suggestion

        _vars = wrapper.ptr
```

src/pyscipopt/scip.pxi
```diff
@@ -8233,21 +8268,22 @@ cdef class Model:
 		solution : Solution
 			The corresponding solution in the main model
         """
```
DominikKamp, line 8270:
```suggestion
        """
        cdef SCIP_VAR** vars
```

src/pyscipopt/scip.pxi
```diff
@@ -8233,21 +8268,22 @@ cdef class Model:
...
+        assert sub_model.getNVars(False) >= self.getNVars(True), "The sub_model must have at least as many variables as the main model."
 
-        for i, var in enumerate(sub_model.getVars()):
-            vars[i] = (<Variable>var).scip_var
+        wrapper = _VarArray(sub_model.getVars())
+        vars = wrapper.ptr
 
         name = str_conversion(heur.name)
         _heur = SCIPfindHeur(self._scip, name)
         PY_SCIP_CALL( SCIPtranslateSubSol(self._scip, sub_model._scip, subscip_sol, _heur, vars, &real_sol) )
```
DominikKamp, line 8285:
```suggestion
        wrapper = _VarArray(sub_model.getVars(False))
        _heur = SCIPfindHeur(self._scip, str_conversion(heur.name))
        PY_SCIP_CALL( SCIPtranslateSubSol(self._scip, sub_model._scip, subscip_sol, _heur, wrapper.ptr, &real_sol) )
```

src/pyscipopt/scip.pxi
Joao-Dionisio: How are you getting these warnings?

tests/test_bipartite.py
Joao-Dionisio: Not really, but that one looked particularly bad

src/pyscipopt/scip.pxi
```diff
@@ -2397,6 +2398,34 @@ cdef void relayErrorMessage(void *messagehdlr, FILE *file, const char *msg) noex
...
+        else:
+            if not isinstance(vars, (list, tuple)):
+                raise TypeError("Expected Variable or list of Variable, got %s." % type(vars))
+
+            self.size = len(vars)
+            if self.size == 0:
+                self.ptr = NULL
+            else:
+                self.ptr = <SCIP_VAR**> malloc(self.size * sizeof(SCIP_VAR*))
+
```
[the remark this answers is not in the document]
Joao-Dionisio: hmmm slightly debatable, but okay okay

Joao-Dionisio:
That should be it, then. Thank you for the awesome review, Dominik!

PR #1008, opened 2025-05-30
Remove bogus execute permission and shebang line from sudoku example

musicinmybrain:
The sudoku example `examples/finished/sudoku.py` has the execute bit set in its filesystem permissions, and it has a shebang line:

https://github.com/scipopt/PySCIPOpt/blob/eb32a20b1ed3f3300025c57354af6b7cab904081/examples/finished/sudoku.py#L4

However, the shebang line is not the first line of the file, so it has no effect, and the execute bit is useless. Besides, no other examples try to be directly executable in this way. This PR removes both the shebang and the execute bit.


Joao-Dionisio:
Thank you! I think most of the early examples were directly translated from a gurobipy book, maybe they were initially used in a different context. ATM I can't review, but this should be merged

Joao-Dionisio (approved):

PR #1009, opened 2025-06-05
SCIP 10

### LINKED ISSUE #955: How can I use Python to obtain the relationship between the time and gap of each log print during the solving process of Scip solver
I just came into contact with the SCIP solver. I found that the SCIP solver may be blocking all processes

The proof is as follows: 
1) I used the SCIP solver to solve the instance in the main thread, and opened another thread to obtain the gap of SCIP every 1 second, which can only obtain the gap value of the last time; 
2) I placed the SCIP solver in one thread and obtained the gap of SCIP every 1 second in another thread. I opened two threads in the main thread, but the result was that only the last gap value could be obtained.

I want to obtain the relationship between time and gap as recorded in the log of the solver during the solving process. Using a handler with event type `BESTSOLFOUND` can only retrieve a portion of the time and gap values. If I want to obtain all the time and gap data like the log output, how can I do it? If anyone can help solve it, I would be extremely grateful

The log information recorded during the execution process is as follows：

```
 time | node  | left  |LP iter|LP it/n|mem/heur|mdpt |vars |cons |rows |cuts |sepa|confs|strbr|  dualbound   | primalbound  |  gap   | compl. 
t 0.0s|     1 |     0 |     0 |     - | trivial|   0 | 412 | 176 |   0 |   0 |  0 |   0 |   0 | 1.000156e+05 | 2.460007e+03 |3965.66%| unknown
p 0.0s|     1 |     0 |     0 |     - |  clique|   0 | 412 | 176 | 176 |   0 |  0 |   0 |   0 | 1.000156e+05 | 6.292663e+03 |1489.40%| unknown
p 0.0s|     1 |     0 |     8 |     - | vbounds|   0 | 412 | 176 | 176 |   0 |  0 |   0 |   0 | 1.000156e+05 | 6.341714e+03 |1477.11%| unknown
  0.0s|     1 |     0 |   851 |     - |  9350k |   0 | 412 | 176 | 176 |   0 |  0 |   0 |   0 | 7.746556e+03 | 6.341714e+03 |  22.15%| unknown
  0.1s|     1 |     0 |   879 |     - |    19M |   0 | 412 | 176 | 178 |   2 |  1 |   0 |   0 | 7.731929e+03 | 6.341714e+03 |  21.92%| unknown
  0.1s|     1 |     0 |   910 |     - |    29M |   0 | 412 | 176 | 183 |   7 |  2 |   0 |   0 | 7.714161e+03 | 6.341714e+03 |  21.64%| unknown
  0.1s|     1 |     0 |   948 |     - |    39M |   0 | 412 | 176 | 185 |   9 |  3 |   0 |   0 | 7.700883e+03 | 6.341714e+03 |  21.43%| unknown
  0.2s|     1 |     0 |   995 |     - |    44M |   0 | 412 | 176 | 186 |  10 |  4 |   0 |   0 | 7.693293e+03 | 6.341714e+03 |  21.31%| unknown
  0.2s|     1 |     0 |  1013 |     - |    49M |   0 | 412 | 176 | 187 |  11 |  5 |   0 |   0 | 7.690795e+03 | 6.341714e+03 |  21.27%| unknown
  0.2s|     1 |     0 |  1035 |     - |    57M |   0 | 412 | 176 | 189 |  13 |  6 |   0 |   0 | 7.688735e+03 | 6.341714e+03 |  21.24%| unknown
  0.3s|     1 |     0 |  1058 |     - |    60M |   0 | 412 | 176 | 191 |  15 |  7 |   0 |   0 | 7.687740e+03 | 6.341714e+03 |  21.22%| unknown
  0.3s|     1 |     0 |  1085 |     - |    66M |   0 | 412 | 176 | 193 |  17 |  8 |   0 |   0 | 7.686455e+03 | 6.341714e+03 |  21.20%| unknown
  0.3s|     1 |     0 |  1109 |     - |    69M |   0 | 412 | 176 | 197 |  21 |  9 |   0 |   0 | 7.685850e+03 | 6.341714e+03 |  21.20%| unknown
  0.4s|     1 |     0 |  1144 |     - |    78M |   0 | 412 | 176 | 199 |  23 | 10 |   0 |   0 | 7.684444e+03 | 6.341714e+03 |  21.17%| unknown
  0.4s|     1 |     0 |  1165 |     - |    78M |   0 | 412 | 176 | 200 |  24 | 11 |   0 |   0 | 7.683881e+03 | 6.341714e+03 |  21.16%| unknown
```

### LINKED ISSUE #1006: Preparing SCIP 10
To make the transition to SCIP 10 smoother, I figured I'd take some notes on what needs to be done to ensure compatibility and which functions need to be wrapped.

- [x] New event types `SCIP_EVENTTYPE_IMPLTYPECHANGED`, ` SCIP_EVENTTYPE_DUALBOUNDIMPROVED`, `SCIP_EVENTTYPE_GAPUPDATED`
- [ ] IISFINDER functions (almost done)
- [x] Statistics Json functions
- [x] `SCIPvarGetImplType()`, `SCIPvarIsBinary()`, `SCIPvarIsIntegral()`, `SCIPvarIsImpliedIntegral()`, and `SCIPvarIsNonimpliedIntegral()`
- [ ] Exact SCIP functions
- [ ] Many many new API functions 

Before merging these things, would like to add a bunch of tests as well.

Joao-Dionisio:
Fix #1006 
Fix #955 

Naturally the tests won't pass, since we need to update the SCIP version. The IIS stuff still needs more work.

Interesting behaviors:

It seems that an empty relaxator now does not raise an error (or at least pyscipopt is not catching it). Would like to find the explanation before removing the test.

The pricer test is also failing, even with heuristics disabled, it seems that SCIP is finding solutions outside of the pricer.

Even though implied integers are only gonna be removed in SCIP 11, adding a variable with an `M` vtype appears to add a continuous variable already.


DominikKamp:
> It's probably on purpose, but feels a bit strange to have a variable with a CONTINUOUS vtype and var.isIntegral() returning True.

The variable type is just more independent from the integrality property now, so on purpose. Here it means that integrality can be assumed but will not be enforced, maybe worth a small comment. Good that this works!

Joao-Dionisio:
> Good that this works!

Found this comment interesting. It wasn't something that was tested when merging? This gives me an idea. What about having a PySCIPOpt branch always up-to-date with SCIP master, and then people keep adding tests to it? Of course, some things would still need to be tested in SCIP proper, but new/changed API functions could perhaps be tested directly on PySCIPOpt.

DominikKamp:
Not sure what should have been tested. The implint type just does no longer exist. In principle having a permanent development branch sounds fine, just be aware of that on `master` API functions can change at any time.

Joao-Dionisio:
Yeah, I meant mostly as a way to more easily test out new features. At least for me, not having to deal with Criterion, Gdb, and recompiling the tests after every little change is a godsend. Maybe the other devs would also be more willing to create tests this way. And there are cases where it doesn't make a difference where the test is in C-SCIP or PySCIPOpt.

DominikKamp:
I also believe that if there were a regular development branch, it would be more possible that developers contribute to it, however, some delay makes sense as long as something is under development on the SCIP master branch, which is normally the case until it is released.

DominikKamp:
Nowadays, `SCIPiisGreedyMinimize()` is rather called `SCIPiisGreedyMakeIrreducible()`.

DominikKamp:
For exact SCIP an interface between `fractions.Fraction` and `SCIP_RATIONAL*` would be nice, but maybe rather in the long run because the memory handling might be involved.

DominikKamp:
Extended the description by remaining fails.

DominikKamp:
Build error updated.

Joao-Dionisio:
@Opt-Mucca, in your opinion, which IIS methods should be wrapped? Currently, I have

```
SCIP_RETCODE SCIPincludeIISfinder()
SCIPiisfinderGetData()
SCIPincludeIISfinderGreedy()
SCIPiisGreedyMakeIrreducible();
SCIPiisIsSubscipInfeasible();
SCIPiisIsSubscipIrreducible();
SCIPiisGetSubscip();
```
I don't know if `SCIPiisfinderGetData` makes much sense, since I guess people will just handle data on the python side.

Opt-Mucca:
@Joao-Dionisio looks good aside from the most important one: `SCIPgenerateIIS`
I don't think the include greedy IIS is needed either.

mmghannam:
Closing this until SCIP 10 is released.

src/pyscipopt/recipes/infeasibilities.py
```diff
@@ -36,7 +36,7 @@ def get_infeasible_constraints(orig_model: Model, verbose: bool = False):
 
     n_infeasibilities_detected = 0
     for c in binary:
-        if model.isGT(model.getVal(binary[c]), 0):
+        if model.isInfinity(binary[c]) or model.isGT(model.getVal(binary[c]), 0):
```
Joao-Dionisio: Just to make debug mode happy

Joao-Dionisio:
That should be it, apart from interfacing `SCIP_RATIONAL*`, which will be done in a future version, most likely.

Now just need to wait for the new releases to hit github, to then build SCIP on scipoptsuite-deploy, and then we have a release.

Opt-Mucca:
@Joao-Dionisio Great work! Everything looks good to me.
Minor questions:
- Is the CHANGELOG updated in the wrong section?
- Should we put `exact` somewhere in the name of all functions associated with exact SCIP?

Joao-Dionisio:
@Opt-Mucca oops, thanks for the heads up. Yeah, you're right on the CHANGELOG.

For the exact methods, I think the only one missing is `allowNegSlack`, but that's also missing from the SCIP side. Not the sort of thing you want to find on release day :)
But we can add the exact keyword on the python front, of course!

Joao-Dionisio:
We also need to update the link where we're downloading SCIP from. From now on, the binaries will not be uploaded to SCIP's GitHub, so we should get them directly from SCIP's website. I'll do the change later.

Opt-Mucca:
> From now on, the binaries will not be uploaded to SCIP's GitHub, so we should get them directly from SCIP's website

@svigerske Created a fully static build????

svigerske:
> > From now on, the binaries will not be uploaded to SCIP's GitHub, so we should get them directly from SCIP's website
> 
> @svigerske Created a fully static build????

I only made tarballs and zip archives that contain all (shared/dynamic) libs necessary to run on systems with Glibc 2.28 (Linux) / macOS 13 (arm64) / MSVS-redistributables (Windows).

DominikKamp:
Please revert the last commit, breaks the build with
~~~
tomllib.TOMLDecodeError: Cannot overwrite a value (at line 107, column 40).
~~~

Joao-Dionisio:
I'm having some problems with setting up the release, was just trying out the removal of the cached SCIP, in case something outdated got stored. Just reverted.

Joao-Dionisio:
We can now build the wheels [successfully](https://github.com/scipopt/PySCIPOpt/actions/runs/19729751021). We just need to check naming conventions for the IIS stuff, and we might want to interface a few more methods. Other than this, the documentation for IIS is also missing, but it's not strictly required for the release, although it would be nice.

Joao-Dionisio:
I think that is mostly it. Custom IISfinders are still not working properly. I need to find a way to access the IIS object from inside the IISfinder. Problem is that the problem might not even be infeasible, so I will need to think about this later. But, for now, I think we can just do the release.
The presolver plugin will probably warrant a new release, so we can just try to improve it by then.

Joao-Dionisio:
@Opt-Mucca, can you please take a look at the IIS stuff?
I think the interfacing is now somewhat okay, but maybe you want to have more functionality.
The thing with the `test_iisGreddyMakeIrreducible()` is just confusing me, though, because it seems that the priorities work as intended, just not the irreducible flag (or something in the interfacing, quite likely).

So, things that happen with the `setParam("iis/irreducible", False)`.
- The custom iisfinder with a higher priority and without the skip option actually finds an IIS, and doesn't get to the greedy one. 
- The custom iisfinder with a lower priority gets to the greedy iisfinder first, which actually finds an IIS
- The custom iisfinder with a higher priority and with the skip option also gets to the greedy iisfinder <- this is the behavior we don't expect

We were hoping to isolate the behavior of a custom iis finder by setting the `irreducible` flag to false.

DominikKamp:
It really does not make sense that the custom iisfinder with highest priority finds an IIS without stopafterone option but not with stopafterone option. Did the problem or the custom iisfinder change in between? The stopafterone code looks fine as far as I can see (the greedy irreducibilitization is independent as mentioned in the parameter description).

Joao-Dionisio:
@DominikKamp it's not just the stopafterone that changes, I also change the behavior of the iisfinder. If stopafterone (rather the irreducible flag) the iisfinder doesn't do anything, but it's still called, so I would say it's still expected that the greedy is not called

PR #1012, opened 2025-06-15
Fix tutorial model

aapeliv:
The tutorial model as implemented in the code block differs from the equations above and was unbounded. This fixes it to be as in the display equation.


Joao-Dionisio:
Oh, thank you @aapeliv!

Joao-Dionisio (approved):

PR #1013, opened 2025-06-27
Add enableDebugSol following discussion on mailing list

Joao-Dionisio:
See details here: [https://listserv.zib.de/pipermail/scip/2025-June/005038.html](https://listserv.zib.de/pipermail/scip/2025-June/005038.html)

Want to add tests before merging. In general, it might be good to have some of the tests run on debug mode. Any opinion @DominikKamp?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 ### Added
-- Added enableDebugSol, disableDebugSol, and check for WITH_DEBUG_SOLUTION flag
+- Added enableDebugSol() and disableDebugSol() for controlling the debug solution mechanism if DEBUGSOL=true
 - More support for AND-Constraints
 - Added support for knapsack constraints
 - Added isPositive(), isNegative(), isFeasLE(), isFeasLT(), isFeasGE(), isFeasGT(), isHugeValue(), and tests
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 ### Added
-- Added enableDebugSol
+- Added enableDebugSol, disableDebugSol, and check for WITH_DEBUG_SOLUTION flag
 - More support for AND-Constraints
 - Added support for knapsack constraints
 - Added isPositive(), isNegative(), isFeasLE(), isFeasLT(), isFeasGE(), isFeasGT(), isHugeValue(), and tests
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 ### Added
+- Added enableDebugSol
 - More support for AND-Constraints
 - Added support for knapsack constraints
 - Added isPositive(), isNegative(), isFeasLE(), isFeasLT(), isFeasGE(), isFeasGT(), isHugeValue(), and tests
--- a/docs/build.rst
+++ b/docs/build.rst
@@ -161,6 +161,7 @@ To use debug information in PySCIPOpt you need to build it with the following co
 .. code-block::
   
   export CFLAGS="-UNDEBUG"
+  export CXXFLAGS="-UNDEBUG"
   python -m pip install .
 
 .. note:: Be aware that you will need the debug library of the SCIP Optimization Suite for this to work
--- a/docs/build.rst
+++ b/docs/build.rst
@@ -159,8 +159,9 @@ Build with Debug
 To use debug information in PySCIPOpt you need to build it with the following command:
 
 .. code-block::
-
-  python -m pip install --install-option="--debug" .
+  
+  export CFLAGS="-UNDEBUG"
+  python -m pip install .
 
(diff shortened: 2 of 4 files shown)


DominikKamp:
In my opinion the debug solution mechanism could be enabled in `Model.__init__()` like in `SCIPrunShell()` maybe controlled by an additional parameter. Interfacing `SCIPenableDebugSol()` is also okay. But then `SCIPdisableDebugSol()` should be added as well.

Joao-Dionisio:
Sounds good, Dominik, thank you! Mark raised a valid point, however, of having code working only for specific SCIP builds. I'm going to add an assert that SCIP is indeed in debug mode, just need to figure out what to query from SCIP.

DominikKamp:
I guess you actually want to check whether `WITH_DEBUG_SOLUTION` is defined.

Joao-Dionisio:
I've never used `WITH_DEBUG_SOL`, but I feel more comfortable not adding more stuff to the Model class that might confuse newcomers. I'll ask Benjamin for the test, and then I think we're good to go.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 ### Added
+- Added enableDebugSol, disableDebugSol, and check for WITH_DEBUG_SOLUTION flag
```
```suggestion
- Added enableDebugSol() and disableDebugSol() for controlling the debug solution mechanism if DEBUGSOL=true
```

docs/build.rst
```diff
@@ -159,8 +159,9 @@ Build with Debug
 To use debug information in PySCIPOpt you need to build it with the following command:
 
 .. code-block::
-
-  python -m pip install --install-option="--debug" .
+  
+  export CFLAGS="-UNDEBUG"
```
DominikKamp, line 172:
```suggestion
  export CFLAGS="-UNDEBUG"
  export CXXFLAGS="-UNDEBUG"
```

src/pyscipopt/scip.pxi
```diff
@@ -56,6 +56,16 @@ else:
...
 
+cdef extern from "scip/config.h":
+    """
+    #ifndef WITH_DEBUG_SOLUTION
+    #define WITH_DEBUG_SOLUTION 0
+    #endif
+    """
+    bint WITH_DEBUG_SOLUTION
+
+cdef bint with_debug_solution = WITH_DEBUG_SOLUTION
```
DominikKamp: If we already have `WITH_DEBUG_SOLUTION`, why do we need `with_debug_solution`?

src/pyscipopt/scip.pxi
```diff
@@ -7495,6 +7502,24 @@ cdef class Model:
...
         PY_SCIP_CALL(SCIPsetRelaxSolVal(self._scip, NULL, var.scip_var, val))
 
+    def enableDebugSol(self):
+        """
+        Enables the debug solution mechanism, which allows tracing back the invalidation of
+        a debug solution during the solution process of SCIP. It must be explicitly
+        enabled for the SCIP data structure.
+        """
+        if not with_debug_solution:
+            raise RuntimeError("SCIP was not built with the WITH_DEBUG_SOLUTION flag. Please rebuild SCIP with this flag enabled.")
```
```suggestion
            raise RuntimeError("SCIP must be built with `DEBUGSOL=true` to enable the debug solution mechanism.")
```

src/pyscipopt/scip.pxi
```diff
@@ -7495,6 +7502,24 @@ cdef class Model:
...
+        SCIPenableDebugSol(self._scip)
+    
+    def disableDebugSol(self):
+        """
+        Disables the debug solution mechanism.
+        """
+        if not with_debug_solution:
+            raise RuntimeError("SCIP was not built with the WITH_DEBUG_SOLUTION flag. Please rebuild SCIP with this flag enabled.")
```
```suggestion
            raise RuntimeError("SCIP must be built with `DEBUGSOL=true` to disable the debug solution mechanism.")
```

DominikKamp:
A test for the added functions is difficult without a way to get the current state, and testing the debug solution mechanism itself is also involved since an assertion fail would need to be identified, so I would not insist on it.

src/pyscipopt/scip.pxi
Joao-Dionisio: Figured having something to differentiate from the actual flag would be nice, but doesn't make sense, you're right.

Joao-Dionisio:
Alright, Dominik, thank you for the review, as always. I'll see what Benjamin ends up sending, and decide later if it can be turned into a test.

Joao-Dionisio:
@DominikKamp can you please approve?

DominikKamp:
@Joao-Dionisio Have you checked locally that `WITH_DEBUG_SOLUTION` is set to `True` when linking SCIP with `DEBUGSOL=true` (I am not sure where the initialization happens)?

DominikKamp (approved):
If the define works, this is fine.

Joao-Dionisio:
Thanks for the help @gtjusila!

PR #1023, opened 2025-07-12
Add SCIPvarMarkRelaxationOnly

### LINKED ISSUE #1019: consinitlp return options
Hello,

I am looking to add a custom linear constraint to the LP relaxation prior to the initial solve. The SCIP documentation is useful in outlining that this can be achieved through a constraint handler, using the CONSINITLP callback method. However, I do not see any documentation on what the appropriate return results should be reported back to SCIP. For now I am using SCIP_RESULT.FEASIBLE and it seems to work ok, but reporting feasibility prior to solving the LP relaxation seems out of place. Is there a more appropriate result to report? Or perhaps does it not matter? 
Thank you!
Kody

Joao-Dionisio:
Fix #1019 

Could you provide a test, please, @kkazda?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 - Added isPositive(), isNegative(), isFeasLE(), isFeasLT(), isFeasGE(), isFeasGT(), isHugeValue(), and tests
 - Added SCIP_LOCKTYPE, addVarLocksType(), getNLocksDown(), getNLocksUp(), getNLocksDownType(), getNLocksUpType(), and tests
 - Added addMatrixConsIndicator(), and tests
-- Added SCIPvarMarkRelaxationOnly, SCIPvarIsRelaxationOnly, and tests
+- Added SCIPvarMarkRelaxationOnly, SCIPvarIsRelaxationOnly, SCIPvarMarkDeletable, SCIPvarIsDeletable, and tests
 ### Fixed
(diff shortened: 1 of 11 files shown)


kkazda:
i am getting the following error when running var.markRelaxationOnly()
"*** TypeError: an integer is required"

kkazda:
i don't believe PY_SCIP_CALL should call SCIPvarMarkRelaxationOnly as its does not return a RET_CODE.
also, i believe it would be valuable to also implement SCIPvarIsRelaxationOnly so a check to see if a variable is properly marked as relaxation only can be performed.

Joao-Dionisio:
Oops! Yeah that's what you get for trusting Copilot a bit too much :) 
I'll edit it tomorrow morning, sorry

Joao-Dionisio:
@kkazda should be good now, works locally

Joao-Dionisio:
In SCIP in debug mode, we're getting a segfault. I figure it's because the variable is not marked as deletable.
To add `SCIPvarMarkDeletable`, we'd also need to interface `SCIPcreateVar` or add another optional argument to `addVar` (`deletable=False`, which would mark it deletable before adding the variable to the problem). I'm not too keen on either option, so if this is working for you @kkazda, then I would merge as is.

Joao-Dionisio:
After some thought, I decided to go with the optional argument approach, and also decided against adding this argument to the matrix variables, at least for now.

PR #1027, opened 2025-07-17
Fix debug command

Joao-Dionisio:
To use pyscipopt with debug information, the documentation suggests using a command which is in the process of being deprecated. Currently we get this warning:

```
DEPRECATION: --build-option and --global-option are deprecated. pip 24.2 will enforce this behaviour change. A possible replacement is to use --config-settings. Discussion can be found at https://github.com/pypa/pip/issues/11859
```

This PR attempts an alternative.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -40,8 +40,9 @@ jobs:
 
       - name: Install PySCIPOpt
         run: |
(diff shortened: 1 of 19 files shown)


Joao-Dionisio:
As usual, I forgot to switch to master before branching...

Issue #1050, opened 2025-08-07
Adding stage checks to all the methods


Joao-Dionisio:
For a long time, we've had problems related to calling methods outside the allowed Stages. (For more information on the stages, see [here](https://www.scipopt.org/doc/html/type__set_8h.php#ac8ea6cabb843c57e1232f3132959e197))
SCIP does the stage check internally, but only if it's compiled in Debug mode. 

The following code seems to work:

```python
m = Model()
m.restartSolve()
```

Even though `restartSolve` cannot be called in the `PROBLEM` stage.

With SCIP in debug mode, we get:
```
[debug.c:2345] ERROR: cannot call method <SCIPrestartSolve> in problem creation stage
[scip_solve.c:3660] ERROR: Error <-8> in function call
```

This needs to be fixed, as we cannot expect users to know about debug mode. (Issues in #1001, #907, #971, #610)

So, I think we need to add stage checks to all the methods. This is a lot of work, so it should be an ongoing process. For now, I'll just fix the issues I linked here, but I will keep the issue open for the other methods.

There is also a question of performance. If a user is adding 10 thousand variables in a loop, checking for the stage in all these variables seems a bit unnecessary. Perhaps we can have a global `CHECK_STAGE` variable which can be set/unset. Suggestions on how to deal with this are welcome.

DominikKamp:
I believe, if this is helpful for PySCIPOpt users by default, we should replace the `#ifndef NDEBUG` around `SCIPcheckStage()` by another define, which is automatically set if `NDEBUG` is not set and can additionally be set by cmake parameter so that PySCIPOpt can be built by default with this SCIP configuration, while being able to turn off the stage checks by a PySCIPOpt parameter.

Joao-Dionisio:
Ah, I re-added the labels without reading your comment properly.

So, something like

```c
#ifndef NStageCheck
SCIP_RETCODE SCIPcheckStage(...)
...
#endif
```

and users would do something like `pip install pyscipopt -StageCheck=True/False`?
I'd rather the users had the ability to turn it on and off without the need to recompile.

Either with an argument at model creation (`m = Model(checkStage=True/False)`), or even better with a method.
`m.enableStageCheck()`

DominikKamp:
If we do it in SCIP, they have to recompile SCIP to switch between these configurations because we will not add a SCIP runtime parameter for it (too expensive).

DominikKamp:
[...]

PR #1051, opened 2025-08-07
Adding stage checks

### LINKED ISSUE #971: Segfault when providing initial solutions
**Describe the bug**

When calling `model.trySol`to provide an initial solution before solving the problem, a segfault occurs

**To Reproduce**

```
from pyscipopt import Model


def _init():
    model = Model()
    model.hideOutput()
    x = model.addVar("x", "B")
    y = model.addVar("y", "B")
    z = model.addVar("z", "B")
    return model, x, y, z


def _optimize(name, m):
    m.optimize()
    print("* %s constraint *" % name)
    objSet = bool(m.getObjective().terms.keys())
    print("* Is objective set? %s" % objSet)
    if objSet:
        print("* Sense: %s" % m.getObjectiveSense())
    status = m.getStatus()
    print("* Model status: %s" % status)
    if status == 'optimal':
        for v in m.getVars():
            if v.name != "n":
                print("%s: %d" % (v, round(m.getVal(v))))
    else:
        print("* No variable is printed if model status is not optimal")
    print("")


def and_constraint(sense="minimize"):
    """ AND constraint """
    model, x, y, z = _init()
    r = model.addVar("r", "B")
    
    model.addConsAnd([x, y, z], r)

    model.addCons(x == 1)

    model.setObjective(r, sense=sense)

    sol = model.createSol()

    model.setSolVal(sol, x, 1)
    model.setSolVal(sol, y, 1)
    model.setSolVal(sol, z, 1)
    model.setSolVal(sol, r, 1)

    print("Storing solution")
    stored = model.trySol(sol)
    print(f"Stored solution: {stored}")

    _optimize("AND", model)


if __name__ == "__main__":
    and_constraint()

```

**Expected behavior**

The correct method is `model.addSol(sol)`, which works as intended. I realize that you are not supposed to call `model.trySol(sol)` during the construction of the model, but a corresponding error message would be very much appreciated instead of a segfault. I recall that SCIP generally checks the current stage (at least when assertions are enabled) and bails out otherwise. It would be great if segfaults could be avoided in general when using python...

**System**
 - OS: Linux
 - SCIP version: 9.2.1
 - How did you install `pyscipopt`? pip

Joao-Dionisio:
Fix #971

### DIFF
--- a/src/pyscipopt/scip.pxi
+++ b/src/pyscipopt/scip.pxi
@@ -1804,37 +1804,6 @@ cdef class Variable(Expr):
             mayround = SCIPvarMayRoundUp(self.scip_var)
         return mayround
 
-    def getNBranchings(self, branchdir):
-        """
-        returns the number of times a bound of the variable was changed in given direction due to branching
-
-        Parameters
-        ----------
-        branchdir : PY_SCIP_BRANCHDIR
-            branching direction (downwards, or upwards)
-
-        Returns
-        -------
-        int
-        """
-        return SCIPvarGetNBranchings(self.scip_var, branchdir)
-
-    def getNBranchingsCurrentRun(self, branchdir):
-        """
-        returns the number of times a bound of the variable was changed in given direction due to branching in the
-        current run
-
-        Parameters
-        ----------
-        branchdir : PY_SCIP_BRANCHDIR
-            branching direction (downwards, or upwards)
-
-        Returns
-        -------
-        int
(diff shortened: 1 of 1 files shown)


Joao-Dionisio:
@DominikKamp, assigning you just to get your input on the broader issue (#1050)

Joao-Dionisio:
A better fix comes with #1056

PR #1056, opened 2025-08-16
Release 5.6.0

### LINKED ISSUE #971: Segfault when providing initial solutions

### LINKED ISSUE #1018: Docker build error on platform linux/arm64
**Describe the bug**
Error when building docker image on platform linux/arm64

**To Reproduce**
Create a dockerfile
```
FROM python:3.12
RUN pip install PySCIPOpt
```
Run `docker build -f Dockerfile . --platform linux/arm64`. Failure
Run `docker build -f Dockerfile . --platform linux/amd64`. Pass

**Expected behavior**
The behaviour between the two platform should be the same.

**Screenshots**

**System**
 - OS: Max
 - Version M2 ship
 - SCIP version
 - How did you install `pyscipopt`? Pip or uv install

**Additional context**

[log.txt](https://github.com/user-attachments/files/21137562/log.txt)

### LINKED ISSUE #1050: Adding stage checks to all the methods

Joao-Dionisio:
fix #971 
fix #1050 
fix #1018 

Just leaving this here for the release we'll probably do next week or the one after. It's going to use SCIP with the following hash: [scip@747acc1e](https://github.com/scipopt/scip/commit/747acc1eddb76397842619f0a30f96692315d50d)
This commit in particular splits stage checks from debug mode, and will be enabled by default on PySCIPOpt. Performance degradation is minimal, and it solves a bunch of user-issues. We'll also add the option to remove these stage checks, for the small performance gain.

Plus, the plan is to also start shipping PySCIPOpt with PaPILO and try to make it compatible with Linux ARM.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 ### Added
+### Fixed
+### Changed
+### Removed
+
+## 5.6.0 - 2025.XX.YY
+### Added
 - More support for AND-Constraints
 - Added support for knapsack constraints
 - Added isPositive(), isNegative(), isFeasLE(), isFeasLT(), isFeasGE(), isFeasGT(), isHugeValue(), and tests


mmghannam:
This is outdated, right? @Joao-Dionisio

Joao-Dionisio:
@mmghannam Yes, but this comment is important:

> This commit in particular splits stage checks from debug mode, and will be enabled by default on PySCIPOpt. Performance degradation is minimal, and it solves a bunch of user-issues. We'll also add the option to remove these stage checks, for the small performance gain.

We need to alter `setup.py` slightly, and add a line to the installation instructions. The relevant Gitlab issue: [https://git.zib.de/integer/scip/-/merge_requests/3938](https://git.zib.de/integer/scip/-/merge_requests/3938)

PR #1093, opened 2025-10-24
Use optional dependencies instead of dependency groups

jonathanberthias:
The dependency groups are only available with `pip` versions which are not compatible with Python 3.8, so to keep supporting that version, switch to dependency extras.

Follow-up to #1082, #1087, #1088, #1090

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -3,11 +3,10 @@ name: Integration test
 env:
   version: 9.2.0
 
-# runs on branches and pull requests; doesn't run on tags.
 on:
   push:
-    # branches:
-      # - 'master'
+    branches:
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -26,17 +26,16 @@ dynamic = ["version"]
 [project.urls]
 Homepage = "https://github.com/SCIP-Interfaces/PySCIPOpt"
 
-[dependency-groups]
+[project.optional-dependencies]
 test = [
     "pytest",
     "pytest-xdist",
-]
-test-full = [
```
jonathanberthias, line 34: I removed one group because optional dependency groups can't depend on one another, so it led to a lot of repetition. There are now only 2 groups: `test` and `coverage`

PR #1094, opened 2025-10-27
Specify test dependencies in requirements.txt files

jonathanberthias:
A new attempt since the previous one #1093 broke the MacOS integration tests :disappointed: 

It wasn't related to the dependencies themselves, but rather to the fact that the optional dependency method forces us to install PySCIPOpt at the same time as the dependencies, whereas there are separate steps in the GitHub workflow. I think the separation is a nice thing to keep, so the optional dependencies can't really be used.

Instead, I suggest using the oldest method in the books: `requirements.txt` files. It can do everything the dependency groups can, and is supported by all `pip` versions.

NB: I did a couple extra changes to the integration test file, let me know if you want me to revert them


Joao-Dionisio:
Thanks @jonathanberthias :)

I'll likely ask one of the others to do a proper review, but in the meantime, do you know of a way to run the pipelines locally, without having to keep accepting and reverting PRs? (I'm asking for myself as well btw)

.github/workflows/coverage.yml
```diff
@@ -1,6 +1,6 @@
 name: Run tests with coverage
 env:
-  version: 9.2.1
+  version: 9.2.4
```
Joao-Dionisio:
```suggestion
  version: 9.2.3
```

I would rather have the SCIP version doing the testing being the one that's shipped when pip installing pyscipopt. Until [this](https://github.com/scipopt/scipoptsuite-deploy/pull/29) gets merged, and we have a new release, the scip version should be 9.2.3 (same elsewhere).

It's something we need to add to `RELEASE.md`, though, didn't realize this was being forgotten.

.github/workflows/integration-test.yml
```diff
@@ -1,7 +1,7 @@
 name: Integration test
 
 env:
-  version: 9.2.0
+  version: 9.2.4
```
Joao-Dionisio:
```suggestion
  version: 9.2.3
```

.github/workflows/integration-test.yml
```diff
@@ -96,7 +94,7 @@ jobs:
           wget --quiet --no-check-certificate https://github.com/scipopt/scip/releases/download/$(echo "v${{env.version}}" | tr -d '.')/SCIPOptSuite-${{ env.version }}-Darwin.sh
           chmod +x SCIPOptSuite-${{ env.version }}-Darwin.sh
           ./SCIPOptSuite-${{ env.version }}-Darwin.sh --skip-license --include-subdir
-          mv SCIPOptSuite-${{ env.version }}-Darwin ${{ runner.workspace }}/scipoptsuite
+          mv SCIPOptSuite-${{ env.version }}-Darwin ${{ github.workspace }}/scipoptsuite
```
Joao-Dionisio, line 98: What's the difference here?

.github/workflows/coverage.yml
jonathanberthias: Sure that makes sense, I saw your PR and thought it might be merged by the time I opened this one :laughing:

.github/workflows/integration-test.yml
jonathanberthias: My editor was saying `runner.workspace` doesn't exist, and indeed it isn't in the [documentation](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#runner-context). I'm not sure if it exists but isn't documented or if its value is just undefined.

In any case, I think using the standard `github.workspace` is safer in the long run since it is:
> The default working directory on the runner

jonathanberthias:
> do you know of a way to run the pipelines locally, without having to keep accepting and reverting PRs?

I tried modifying the trigger to run on all commits in #1093 (in this [commit](https://github.com/scipopt/PySCIPOpt/pull/1093/commits/42ead97ea861ab2b3510c0ab99d8e0f1b35fedb6)) but it didn't work. I did make it work in some of my repos, so I guess it's because the PR comes from a fork and not the repo itself.

A better way would actually be to add a [`workflow_dispatch`](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_dispatch) trigger so that you can run the tests on-demand. I don't know if it will work on PRs from forks though.

Do you want me to try to add it in this PR?

.github/workflows/coverage.yml
Joao-Dionisio: Oh, I really wouldn't hold my breath on that one, people tend to be pretty busy :)

Joao-Dionisio:
> Do you want me to try to add it in this PR?

That would be great, yes please!

jonathanberthias:
Ok so I think I figured it out. The workflow_dispatch can only be used if it is in the `master`, that's a protection from GitHub to prevent forks from triggering any workflow they want. I opened #1095 for that.

In the meantime, I forced the workflow to run. I thought it didn't work last time I tried, but actually it did, but it ran in my fork, not in this repo. So here is the run for this PR: https://github.com/jonathanberthias/PySCIPOpt/actions/runs/18887608453 :green_circle: 

It remains to see whether the workflow_dispatch will trigger the tests in this repo or in the fork. I assume it will be in the fork, but at least it will be easier than having to push changes to force the workflow to run :smile:

Joao-Dionisio:
Okay, then that should be it! By merging this, I guess we'll test two PRs at once :)

Joao-Dionisio:
image

Awesome!

PR #1097, opened 2025-10-29
Release 570

Joao-Dionisio:
A final release before SCIP 10, also to update to SCIP 9.2.4.

Something went wrong with the previous release, it seems that the changes didn't make it to pip. Let's see if we can fix things now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 ### Changed
 ### Removed
 
-## 5.6.0 - 2025.11.01
+## 5.7.0 - 2025.11.17
 ### Added
 - Added possibility of having variables in exponent.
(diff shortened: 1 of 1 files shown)


Joao-Dionisio:
I don't know what's going on, MacOS 13 is raising this error when building the wheels:

```
  Installing collected packages: numpy, pyscipopt
  Successfully installed numpy-1.24.4 pyscipopt-5.6.0
  + /bin/sh -c 'pytest /Users/runner/work/PySCIPOpt/PySCIPOpt/tests'
  /bin/sh: pytest: command not found
```

leaving it here to come back later

Joao-Dionisio:
This last commit might be a bit controversial, but since SCIP will do the stage checking by default, I wouldn't worry too much. Kept them in a couple, though.

PR #1098, opened 2025-10-30
skip SCIP_ERROR line in coverage

Joao-Dionisio:
Every coverage run yielded the following annotation error:

image

It's not critical, and it was annoying me, so I'll try to suppress it.


Joao-Dionisio:
ah, it seems that we either have this annotation error, or we need to change the tests. Oh well, I'll live with it, for now.

PR #1099, opened 2025-10-30
Add wrappers for methods used in node selection

mmghannam:
I'm planning to port SCIP's default node selection rule as an example here. But we need all these methods to be there first.

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -39,15 +39,15 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
-    
+
       - name: Build wheels
         uses: pypa/cibuildwheel@v2.21.1
         env:
           CIBW_ARCHS: ${{ matrix.arch }}
-          CIBW_TEST_GROUPS: test
+          CIBW_BEFORE_TEST: "pip install -r {project}/requirements/test.txt"
           CIBW_TEST_COMMAND: "pytest {project}/tests"
           CIBW_MANYLINUX_*_IMAGE: manylinux_2_28
-        
+
       - uses: actions/upload-artifact@v4
         with:
           name: wheels-${{ matrix.os}}-${{ matrix.arch }}


src/pyscipopt/scip.pxi
```diff
@@ -3025,6 +3025,72 @@ cdef class Model:
...
+        Gets current plunging depth (successive selections of child/sibling nodes).
+
+        Returns
+        -------
+        int
+
+        """
+        return SCIPgetPlungeDepth(self._scip)
+
+    def getLowerbound(self):
```
Joao-Dionisio, line 3050: In the docstring, can you please state explicitly that this refers to the transformed problem?

src/pyscipopt/scip.pxi
```diff
@@ -3025,6 +3025,72 @@ cdef class Model:
         """
         return SCIPgetDepth(self._scip)
 
+    def getMaxDepth(self):
+        """
+        Gets maximal depth of the branch-and-bound tree reached during solving.
```
```suggestion
        Gets maximal depth of the branch-and-bound tree processed during solving (excluding probing nodes).
```

src/pyscipopt/scip.pxi
```diff
@@ -3025,6 +3025,72 @@ cdef class Model:
...
+        Returns
+        -------
+        float
+
+        """
+        return SCIPgetLowerbound(self._scip)
+
+    def getCutoffbound(self):
+        """
+        Gets the cutoff bound.
```
Joao-Dionisio: Can this be a bit more descriptive please?

tests/test_node_methods.py
```diff
@@ -0,0 +1,171 @@
...
+    nodes = m.getNNodes()
+
+    print(f"Max depth after solving: {max_depth}")
+    print(f"Number of nodes explored: {nodes}")
+    print(f"Optimization status: {m.getStatus()}")
+
+
+    assert max_depth >= 0, f"Expected max_depth >= 0, got {max_depth}"
+
+    if nodes > 1:
```
Joao-Dionisio, line 46: can you instead iterate over the nodes and assert that max depth is equal to the maximum depth of them?

tests/test_node_methods.py
```diff
@@ -0,0 +1,171 @@
...
+    initial_lb = m.getLowerbound()
+    print(f"Initial lower bound: {initial_lb}")
+
+    m.optimize()
+
+    lower_bound = m.getLowerbound()
+    obj_val = m.getObjVal()
+
+    print(f"Lower bound after solving: {lower_bound}")
+    print(f"Objective value: {obj_val}")
```
Joao-Dionisio: Since `getLowerbound` refers to the transformed problem, it's possible this will yield unexpected results.

tests/test_node_methods.py
```diff
@@ -0,0 +1,171 @@
...
+        disable_sepa=True,
+        disable_heur=True,
+        disable_presolve=True,
+        node_lim=10000,
+        small=True
+    )
+
+    m.setIntParam("limits/solutions", 1)
+
+    m.optimize()
```
Joao-Dionisio, line 117: Can you also explicitly do `m.setObjlimit()`, to ensure no shenanigans are going on?

tests/test_node_methods.py
```diff
@@ -0,0 +1,171 @@
...
+
+    print(f"Strong branching LP iterations: {sb_iters}")
+    print(f"Total LP iterations: {total_lp_iters}")
+    print(f"Number of nodes: {nodes}")
+
+    assert sb_iters >= 0, f"Expected non-negative strong branching iterations, got {sb_iters}"
+    assert sb_iters <= total_lp_iters, f"Strong branching iterations {sb_iters} should not exceed total LP iterations {total_lp_iters}"
+
+
+if __name__ == "__main__":
```
Joao-Dionisio: This should probably be removed before merging, for consistency.

Joao-Dionisio (changes requested):

tests/test_node_methods.py
mmghannam: done

src/pyscipopt/scip.pxi
mmghannam: done

tests/test_node_methods.py
mmghannam: can do! good idea

src/pyscipopt/scip.pxi
mmghannam: done

tests/test_node_methods.py
mmghannam: done, added an event handler to track it

tests/test_node_methods.py
mmghannam: The test only checks against the previous queried lowerbound, I don't understand what are you referring to.

tests/test_node_methods.py
mmghannam: what shenanigans haha, wouldn't then an objlimit be more dependent on the run details?

tests/test_node_methods.py
Joao-Dionisio: Ahh right, I forgot that there isn't a `getNodes` method. Wasn't understanding the reason for the event handler.

tests/test_node_methods.py
Joao-Dionisio: I didn't mean that it would break the tests, only that for a given instance, you might be printing a lowerbound that's larger than the objective value (former refers to the transformed problem, latter to the original problem)

Joao-Dionisio: But ultimately, it's no biggie

mmghannam: ahaa, I can print the cutoff bound then

tests/test_node_methods.py
Joao-Dionisio: Yeah, you're right

Joao-Dionisio (approved):

tests/test_node_methods.py
mmghannam: I just removed it, doesn't add much anyway

Joao-Dionisio:
Thanks, Mo!

PR #1101, opened 2025-10-30
Update stubs with node selection methods

jonathanberthias:
Tiny changes to account for the recently merged PR.

It's in draft as if #1099 gets merged, there would be more changes. But if you cut the 5.7 release before merging it, this PR should be included as-is :smile:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
 - MatrixVariable comparisons (<=, >=, ==) now support numpy's broadcast feature.
 - Added methods: getMaxDepth(), getPlungeDepth(), getLowerbound(), getCutoffbound(), getNNodeLPIterations(), getNStrongbranchLPIterations().
 - setup.py now automatically detects conda environments when SCIPOPTDIR is not defined.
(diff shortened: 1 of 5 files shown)


src/pyscipopt/scip.pyi
```diff
@@ -236,12 +236,10 @@ class Expr:
     def __next__(self): ...
     def __pow__(self, other): ...
     def __radd__(self, other): ...
-    def __reduce_cython__(self, *args, **kwargs): ...
```
Joao-Dionisio, line 239: what were these and why don't we need them anymore?

jonathanberthias: These are symbols generated by Cython that exist at runtime so `stubgen` picked them up, but I don't think they are useful to have in the stubs, as I wouldn't expect anyone to try to use them. But maybe I'm wrong, in which case we can add them back :shrug: 

I removed most of these symbols in #1073, these are a few leftovers I missed back then

Joao-Dionisio: alrighty :)

PR #1102, opened 2025-11-05
Fix the case of returning none from the nodeselect callback

mmghannam:


src/pyscipopt/nodesel.pxi
```diff
@@ -87,8 +87,12 @@ cdef SCIP_RETCODE PyNodeselSelect (SCIP* scip, SCIP_NODESEL* nodesel, SCIP_NODE*
   nodeseldata = SCIPnodeselGetData(nodesel)
   PyNodesel = <Nodesel>nodeseldata
   result_dict = PyNodesel.nodeselect()
-  selected_node = <Node>(result_dict.get("selnode", None))
-  selnode[0] = selected_node.scip_node
+  selected_node = <Node>result_dict.get("selnode", None)
+  if selected_node == None:
+    selnode[0] = NULL
```
Joao-Dionisio, line 92: In which cases would the `result_dict` not have `selnode`? Trying to figure out if it should return `NULL` or raise an error.

mmghannam: `selnode` can be None. This is how node selectors communicate the fact that there is no node to select.

mmghannam: The old code was weird anyway. In case selnode is None it was casting it as Node object then getting the scip_node from it. Smells like undefined behavior to me

Joao-Dionisio: 100% on the old code. Tyty, will merge after the other one

Joao-Dionisio (approved):

PR #1103, opened 2025-11-05
Detect SCIP installation in conda environment if SCIPOPTDIR is not defined

mmghannam:
To make it a bit easier to develop pyscipopt for some people. Including me :)


Joao-Dionisio:
I assume you tested this, so all good on my end! Just bringing copilot over, just in case.

mmghannam:
> I assume you tested this, so all good on my end! Just bringing copilot over, just in case.

Yup tested against the latest conda package of scip 9.2.4

PR #1104, opened 2025-11-05
Add the hybrid estimate in C-SCIP as an example

mmghannam:
Needed to port this for another project, thought could be useful for others too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Added basic type stubs to help with IDE autocompletion and type checking.
 - MatrixVariable comparisons (<=, >=, ==) now support numpy's broadcast feature.
 - Added methods: getMaxDepth(), getPlungeDepth(), getLowerbound(), getCutoffbound(), getNNodeLPIterations(), getNStrongbranchLPIterations().
+- setup.py now automatically detects conda environments when SCIPOPTDIR is not defined.
 ### Fixed
 - Implemented all binary operations between MatrixExpr and GenExpr
 - Fixed the type of @ matrix operation result from MatrixVariable to MatrixExpr.
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -32,6 +32,8 @@ you need to specify the install location using the environment variable
     `set SCIPOPTDIR=<path_to_install_dir>` (**cmd**, **Cmder**, **WSL**)\
     `$Env:SCIPOPTDIR = "<path_to_install_dir>"` (**powershell**)
 
+**Note:** If `SCIPOPTDIR` is not set, the setup script will automatically attempt to detect a conda environment (via the `CONDA_PREFIX` environment variable) and use it if available. If no conda environment is detected, it will fall back to searching global installation paths.
+
 `SCIPOPTDIR` needs to have a subdirectory `lib` that contains the
 library, e.g. `libscip.so` (for Linux) and a subdirectory `include` that
 contains the corresponding header files:
--- a/setup.py
+++ b/setup.py
@@ -7,16 +7,28 @@ scipoptdir = os.environ.get("SCIPOPTDIR", "").strip('"')
 extra_compile_args = []
 extra_link_args = []
 
-# if SCIPOPTDIR is not set, we assume that SCIP is installed globally
+# if SCIPOPTDIR is not set, try to detect conda environment, otherwise assume global installation
 if not scipoptdir:
-    if platform.system() == "Darwin":
-        includedir = "/usr/local/include"
-        libdir = "/usr/local/lib"
+    # check if we're in a conda environment
+    conda_prefix = os.environ.get("CONDA_PREFIX", "").strip('"')
+
+    if conda_prefix and os.path.exists(os.path.join(conda_prefix, "include")):
+        includedir = os.path.join(conda_prefix, "include")
+        libdir = os.path.join(conda_prefix, "lib")
+        libname = "libscip" if platform.system() == "Windows" else "scip"
+        print(f"Detected conda environment at {conda_prefix}.")
+        print(f"Using include path {includedir}.")
+        print(f"Using library directory {libdir}.\n")
     else:
-        includedir = "."
-        libdir = "."
-    libname = "libscip" if platform.system() in ["Windows"] else "scip"
-    print("Assuming that SCIP is installed globally, because SCIPOPTDIR is undefined.\n")
+        # fall back to global installation
+        if platform.system() == "Darwin":
+            includedir = "/usr/local/include"
+            libdir = "/usr/local/lib"
+        else:
+            includedir = "."
+            libdir = "."
+        libname = "libscip" if platform.system() == "Windows" else "scip"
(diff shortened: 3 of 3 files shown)


examples/finished/nodesel_hybridestim.py
```diff
@@ -0,0 +1,307 @@
...
+            if self.scip.getNNodes() % self.bestnodefreq == 0:
+                selnode = self.scip.getBestboundNode()
+            else:
+                selnode = self.scip.getBestNode()
+        else:
+            # Get global lower and cutoff bound
+            lowerbound = self.scip.getLowerbound()
+            cutoffbound = self.scip.getCutoffbound()
+
+            # If we didn't find a solution yet, use only 20% of the gap as cutoff bound
```
Joao-Dionisio: can you call this difference something other than gap? or is it a standard term in node selection? (mentioning gap in the same sentence where you say you have no solutions feels iffy)

examples/finished/nodesel_hybridestim.py
```diff
@@ -0,0 +1,307 @@
...
+            -1 if node1 is better than node2
+            0 if both nodes are equally good
+            1 if node1 is worse than node2
+        """
+        score1 = self._get_nodesel_score(node1)
+        score2 = self._get_nodesel_score(node2)
+
+        # Check if scores are equal or both infinite
+        if (self.scip.isEQ(score1, score2) or
+            (self.scip.isInfinity(score1) and self.scip.isInfinity(score2)) or
```
Joao-Dionisio, line 166: I think in debug mode, SCIP throws an assertion when using things like `isEQ` with non-finite values, so maybe consider passing it to the end of the `or`, just to reduce the chances of this happening

examples/finished/nodesel_hybridestim.py
```diff
@@ -0,0 +1,307 @@
...
+                # Calculate maximal plunging bound
+                maxbound = lowerbound + self.maxplungequot * (cutoffbound - lowerbound)
+
+            # We want to plunge again: prefer children over siblings, and siblings over leaves
+            # but only select a child or sibling if its estimate is small enough
+            selnode = None
+
+            # Try priority child first
+            node = self.scip.getPrioChild()
+            if node is not None and node.getEstimate() < maxbound:
```
Joao-Dionisio: would it look nicer if returning the `selnode` right away, instead of multiple nested conditionals?

examples/finished/nodesel_hybridestim.py
```diff
@@ -0,0 +1,307 @@
...
+                else:
+                    return 0
+
+        # Compare scores
+        if score1 < score2:
+            return -1
+        else:
+            return 1
+        
+def random_mip_1(disable_sepa=True, disable_heur=True, disable_presolve=True, node_lim=2000, small=False):
```
Joao-Dionisio: can this not be imported instead of copied here? or would you rather just have everything centralized for clarity?

examples/finished/nodesel_hybridestim.py
mmghannam: I agree, it's copied from the C counterpart haha

mmghannam: I'm not so sure when is the cutoff bound not infinity in this case

examples/finished/nodesel_hybridestim.py
mmghannam: I refactored a bit, no nesting now :)

examples/finished/nodesel_hybridestim.py
mmghannam: Yeah I wanted people to just need the file to run the example. I would be happy with reading a non trivial instance also, or removing the whole test function too I'm fine with that.

examples/finished/nodesel_hybridestim.py
mmghannam: changed the condition

examples/finished/nodesel_hybridestim.py
Joao-Dionisio: Then what about reading that one big instance you uploaded a while back?

(On mobile, can give details later)

mmghannam: sure, done.

Joao-Dionisio:
Can you please also add a link for this example in the tutorial docu? The example there is simple, people may want to take a look at something more complex. Plus, it can bring attention to the github page, for users who are not aware.

examples/finished/nodesel_hybridestim.py
Joao-Dionisio: what about

```python
any_finite = self.scip.isInfinity(abs(score1)) or self.scip.isInfinity(abs(score2))`
```

just to simplify the code?

examples/finished/nodesel_hybridestim.py
Joao-Dionisio: You might have an objective limit, I suppose

Joao-Dionisio: Or say you are minimizing a variable `x`, and you start deriving integrally-valid lower bounds on `x`.

examples/finished/nodesel_hybridestim.py
Joao-Dionisio: can you re-run the code, just to check that the assertion passes?

mmghannam: I did

examples/finished/nodesel_hybridestim.py
mmghannam: sure

examples/finished/nodesel_hybridestim.py
mmghannam: makes sense

mmghannam:
> Can you please also add a link for this example in the tutorial docu? The example there is simple, people may want to take a look at something more complex. Plus, it can bring attention to the github page, for users who are not aware.

Done

examples/finished/nodesel_hybridestim.py
mmghannam: done

examples/finished/nodesel_hybridestim.py
```diff
@@ -0,0 +1,270 @@
...
+            if self.scip.getNNodes() % self.bestnodefreq == 0:
+                selnode = self.scip.getBestboundNode()
+            else:
+                selnode = self.scip.getBestNode()
+        else:
+            # Get global lower and cutoff bound
+            lowerbound = self.scip.getLowerbound()
+            cutoffbound = self.scip.getCutoffbound()
+
+            # If we didn't find a solution yet, use only 20% of the gap as cutoff bound
```
Joao-Dionisio:
```suggestion
            # If we didn't find a solution yet, tighten the cutoff bound to 20% of the range between it and the lowerbound.
```

It can be interpreted as only `0.2 * (cutoffbound - lowerbound)`, but I don't know how to improve it. but people can also just look at the freakin code

mmghannam: 😂

PR #1106, opened 2025-11-10
update setup.py w.r.t. SCIP 10

pfetsch:
The path to config.h in SCIP 10 is given by `lib/shared/include/scip/config.h` now.
The file `config.h` is generated depending on the build flags of make and is placed next to the `.o` files.
But then there is was no connection from `libscip` to `config.h`.
In SCIP 10 this is now resolved by using the generic path mentioned above.
The change now also gets rid of `NO_CONFIG_HEADER` which was a work-around.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,10 +22,10 @@ if not scipoptdir:
     else:
         # fall back to global installation
         if platform.system() == "Darwin":
-            includedir = "/usr/local/include"
+            includedirs = ["/usr/local/include"]
             libdir = "/usr/local/lib"
         else:
-            includedir = "."
+            includedirs = ["."]
             libdir = "."
         libname = "libscip" if platform.system() == "Windows" else "scip"
         print("Assuming that SCIP is installed globally, because SCIPOPTDIR is undefined.\n")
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -34,39 +34,36 @@
 
     # check whether SCIP is installed in the given directory
     if os.path.exists(os.path.join(scipoptdir, "include")):
-        includedir = os.path.abspath(os.path.join(scipoptdir, "include"))
+        includedirs = [os.path.abspath(os.path.join(scipoptdir, "include"))]
     else:
-        print(f"SCIPOPTDIR={scipoptdir} does not contain an include directory; searching for include files in src or ../src directory.")
+        print("SCIPOPTDIR=%s does not contain an include directory; searching for include files in src or ../src directory.\n" % scipoptdir)
```
Joao-Dionisio, line 39: Are you doing this for people trying to compile PySCIPOpt with Python < 3.6?

pfetsch: No particular reason, I just like the more `printf` like notation. I can revert this change, if you like.

Joao-Dionisio: Since the strings are small, I don't mind either way. Thank you!

PR #1107, opened 2025-11-10
Add three functions for variables

pfetsch:
This pull request adds three function concerning variables:
- `vstatus` for obtaining the variable status;
- `isActive` to check whether a variable is active;
- `markDoNotMultaggrVar` to mark a variable to not allowed to be multi-aggregated.

The functions are needed in a secondary project.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,10 +22,10 @@ if not scipoptdir:
     else:
         # fall back to global installation
         if platform.system() == "Darwin":
-            includedir = "/usr/local/include"
+            includedirs = ["/usr/local/include"]
             libdir = "/usr/local/lib"
         else:
-            includedir = "."
+            includedirs = ["."]
             libdir = "."
         libname = "libscip" if platform.system() == "Windows" else "scip"
         print("Assuming that SCIP is installed globally, because SCIPOPTDIR is undefined.\n")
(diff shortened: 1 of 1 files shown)


src/pyscipopt/scip.pxi
```diff
@@ -4413,6 +4450,17 @@ cdef class Model:
         if infeasible:
             print('could not change variable type of variable %s' % var)
 
+    def markDoNotMultaggrVar(self, Variable var):
+        """
+        Marks a variable to not allowed to be multi-aggregated.
```
```suggestion
        Marks a variable preventing it from being multi-aggregated.
```

Original is a bit awkward, but also not super happy with this one.

Joao-Dionisio:
Thank you @pfetsch ! Can you please also add `SCIPmarkDoNotAggrVar()` and a basic test to the added methods and variable type?

src/pyscipopt/scip.pxi
Joao-Dionisio: Maybe `Forbids a variable from being multi-aggregated."

src/pyscipopt/scip.pxi
```diff
@@ -1538,6 +1538,32 @@ cdef class Variable(Expr):
         elif vartype == SCIP_VARTYPE_IMPLINT:
             return "IMPLINT"
 
+    def vstatus(self):
```
```suggestion
    def getStatus(self):
```

Just a suggestion, based on similar examples in the code. But if you believe confusion might arise from this (having a `getStatus` in Model and Variable), then you can leave it as is.

pfetsch: I used the function `vtype` directly above as a "role model". We can of course change this.

pfetsch:
> Thank you @pfetsch ! Can you please also add SCIPmarkDoNotAggrVar() and a basic test to the added methods and variable type?

Yes, will do. This will take a while ...

Joao-Dionisio:
@pfetsch I can do it, no problem. Thank you for the contribution!

PR #1127, opened 2025-12-04
Check Existence of lib when building using custom scip installation

gtjusila:
Hello,

In the new SCIP 10 tarball the lib folder was named lib64 rather than lib. Hence an adjustment was made to check which folder (lib or lib64 exist). 

Cheers,
Gen


Joao-Dionisio:
Thanks Gen!

PR #1140, opened 2025-12-27
Update stubs for latest changes

jonathanberthias:
There have been quite a few new additions since the last update:
- IIS,
- Exact solving,
- A couple new event types,
- Implied int types.

I ran `stubgen` from scratch and filtered out the noise (e.g. it added `-> Any` to all methods, not sure that's useful). The remaining changes are that it now adds private functions and attributes whereas they were ignored before.

The next step I'm working on is to run `stubtest` so that we get an error if we forget to update the stubs, it would avoid having to make these kind of PRs :smile:


Joao-Dionisio:
Awesome! I'll play around with AI a bit to hopefuly make a script that does this automatically. Fingers crossed!

Joao-Dionisio (approved):

Issue #1072, opened 2025-09-15
Distributing type stubs


jonathanberthias:
**Is your feature request related to a problem? Please describe.**

Since this library is packaged as an extension module, type checkers and LSPs are not able to inspect the source code to extract type information. This makes it harder to discover available functions and methods as the IDEs cannot help.

<img width="301" height="111" alt="Image" src="https://github.com/user-attachments/assets/740f03d9-bd1c-4ded-a0c1-c79dc536d713" />

A few months ago, I started working on [type stubs](https://typing.python.org/en/latest/guides/writing_stubs.html) for `pyscipopt`: https://github.com/jonathanberthias/scip-stubs. I haven't caught up with the latest changes yet but it shouldn't be too much work to update. If I install this stub package in the environment, then I get:

<img width="574" height="330" alt="Image" src="https://github.com/user-attachments/assets/724bdcae-d855-489a-a584-fafe0b1e9026" />

**Describe the solution you'd like**

There are 2 ways to distribute the stub files: they can either be part of the package itself as additional files, or they can be distributed in a separate package, conventionally named `pkg-stubs`.

The main difference between the 2 approaches is the development flow. If the stubs are part of the package, then users don't need to install an additional package and they get completion, docstrings in the IDE, type checking, etc. _for free_. However, it puts more burden on the contributors and maintainers as for each change, one has to ensure the stubs reflect the implementation by modifying both the source files and the stub files. It also means a new version of the library must be released for changes to stubs to be available.

If the stubs are packaged separately, there is more freedom to release on a separate schedule. For instance, the type stubs package for `pyscipopt 5.6.0` could be versioned `5.6.0.0`, and if a fix is needed, then `5.6.0.1` can be released. The downside is that the stubs are not in the same repo as the source, increasing the chances that they are out of sync. As an example, to test the stubs, I had to [copy most of the tutorials](https://github.com/jonathanberthias/scip-stubs/tree/main/tests/tutorials) from this repo to have some reference code. If the stubs were in this repo, the code would test the stubs automatically.

I don't have a clear preference for either model. I think it is up to you to decide which model suits your workflow best. In any case, I will gladly share the stubs I have already written and those I will add in the future, either by contributing to this repo or by publishing a separate package.

For reference, `scipy` have a separate repo for stubs: https://github.com/scipy/scipy-stubs, whereas `gurobipy` ship the type stubs with the package directly.

Happy to discuss this further if you need additional information!

[...]

PR #1141, opened 2025-12-28
Run `stubtest` to check stubs are complete

jonathanberthias:
This PR adds a workflow that runs `stubtest` on the bundled type stubs. This means that whenever a class or method is added, the workflow will error until it is added to the `scip.pyi` file.

Part of #1072

# Type errors

A pre-requisite to running `stubtest` is to have `mypy` type check the code without error. There were only a few errors reported by `mypy` which I don't think are controversial

# Allow lists
For now, we rely on an allow list because the stubs are still wrong, they list all methods as taking `*args, **kwargs` parameters which isn't true at runtime. The allow list silences all the known current errors until we manually fix them.
There are 2 allow list files:
- `allowlist` lists errors which are known and expected.
- `todo` lists all the errors which we should aim to fix eventually.

## Entries

Each entry is a regex matching the fully qualified name of the object with an error. Each line can thus hide multiple errors. For instance, for the entry `pyscipopt.LP.addRow`, these are the errors that get silenced:
```
error: pyscipopt.LP.addRow is inconsistent, stub does not have parameter "entries"
error: pyscipopt.LP.addRow is inconsistent, stub does not have parameter "lhs"
error: pyscipopt.LP.addRow is inconsistent, stub does not have parameter "rhs"
error: pyscipopt.LP.addRow is inconsistent, runtime does not have *args parameter "args"
error: pyscipopt.LP.addRow is inconsistent, runtime does not have **kwargs parameter "kwargs"
```

Conversely, some lines are actually duplicates. This is due to the re-exports. For instance, the entries:
```
pyscipopt.LP.addRow
pyscipopt.scip.LP.addRow
```
point to the same underlying object, so both entries could be removed at the same time.

## Maintenance

Finally, the workflow will also fail if an entry in the allow list does not generate any error, ensuring the allow list is always up to date.

## How to fix?

The way forward is to fix all the errors in the `todo` allow list. This means removing a couple of entries from the allow list, running `stubs/test.sh` and fixing any printed error.

Unfortunately I don't know any way to automate this part. `stubtest` detects the argument names but it has no idea what annotation they should have, so some human judgment is required. Or a massive token budget for a LLM :wink:

### DIFF
--- a/.github/workflows/stubs.yml
+++ b/.github/workflows/stubs.yml
@@ -45,6 +45,9 @@ jobs:
           export CFLAGS="-O0 -ggdb -Wall -Wextra -Werror -Wno-error=deprecated-declarations" # Debug mode. More warnings. Warnings as errors, but allow deprecated declarations.
           python -m pip install . -v 2>&1 | tee build.log
 
+      - name: Run MyPy
+        run: python -m mypy --package pyscipopt
+
       - name: Run stubtest
         run: stubs/test.sh
 


stubs/test.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash -e
+
+# Test the stubs for pyscipopt using stubtest
+# This checks that the type hints in the stubs are consistent with the actual implementation
+# Prerequisite: install mypy in same environment as pyscipopt and put stubtest in PATH
```
Joao-Dionisio: What would the error be when `mypy` is not installed? Also, can you be more explicit on what putting stubtest in PATH means?

Joao-Dionisio:
> There were only a few errors reported by mypy which I don't think are controversial

Are the `mypy` errors you ignored the ones you talk about later, or are they different? If so, can you please list (some of) them?

jonathanberthias:
> Are the `mypy` errors you ignored the ones you talk about later, or are they different? If so, can you please list (some of) them?

No the `mypy` errors are the reason there are a few changes to the code outside the `stubs` folder, these are the errors on the current `master` that are now fixed:
```
.venv/lib/python3.14/site-packages/pyscipopt/scip.pyi:382: error: Signature of "__ge__" incompatible with supertype "numpy.ndarray"  [override]
.venv/lib/python3.14/site-packages/pyscipopt/scip.pyi:384: error: Signature of "__le__" incompatible with supertype "numpy.ndarray"  [override]
.venv/lib/python3.14/site-packages/pyscipopt/scip.pyi:397: error: Signature of "__ge__" incompatible with supertype "numpy.ndarray"  [override]
.venv/lib/python3.14/site-packages/pyscipopt/scip.pyi:398: error: Signature of "__le__" incompatible with supertype "numpy.ndarray"  [override]
.venv/lib/python3.14/site-packages/pyscipopt/scip.pyi:1303: error: Incompatible types in assignment (expression has type "None", variable has type "int")  [assignment]
.venv/lib/python3.14/site-packages/pyscipopt/scip.pyi:1304: error: Incompatible types in assignment (expression has type "None", variable has type "int")  [assignment]
.venv/lib/python3.14/site-packages/pyscipopt/scip.pyi:1306: error: Incompatible types in assignment (expression has type "None", variable has type "float")  [assignment]
.venv/lib/python3.14/site-packages/pyscipopt/scip.pyi:1307: error: Incompatible types in assignment (expression has type "None", variable has type "float")  [assignment]
.venv/lib/python3.14/site-packages/pyscipopt/scip.pyi:1308: error: Incompatible types in assignment (expression has type "None", variable has type "float")  [assignment]
.venv/lib/python3.14/site-packages/pyscipopt/scip.pyi:1309: error: Incompatible types in assignment (expression has type "None", variable has type "float")  [assignment]
.venv/lib/python3.14/site-packages/pyscipopt/scip.pyi:1310: error: Incompatible types in assignment (expression has type "None", variable has type "float")  [assignment]
.venv/lib/python3.14/site-packages/pyscipopt/__init__.py:7: error: Item "None" of "str | None" has no attribute "strip"  [union-attr]
.venv/lib/python3.14/site-packages/pyscipopt/recipes/getLocalConss.py:27: error: Need type annotation for "added_conss" (hint: "added_conss: list[<type>] = ...")  [var-annotated]
Found 13 errors in 3 files (checked 10 source files)
```

stubs/test.sh
jonathanberthias: > What would the error be when `mypy` is not installed?

`stubtest` is distributed when you install `mypy`, so installing `mypy` is the only way to install `stubtest`.

> Also, can you be more explicit on what putting stubtest in PATH means?

Yeah I agree that isn't clear, I just meant to activate the environment. Actually, I'll change the command to `python -m stubtest`, I think it's clearer

Joao-Dionisio (approved):
Thank you @jonathanberthias !

PR #1076, opened 2025-10-04
Extend Functionality to Use the Presolver Plugin and Add a Tutorial

fvz185:
To clarify and document the usage of PySCIPOpt's presolver plugin, a working example and a tutorial were added. They required wrappers to be implemented, too. The wrapper functions are supplemented with tests. The example and tutorial replicate the Boundshift presolver.

However, some open issues remain:
1) The test for the var.isActive() wrapper is incomplete since it does not cover all possible settings. Should tests on inactive variables due to probing or aggregation be included?
2) Only SCIP functions that are not easily emulatable within Python and required for the example were wrapped. Should any additional SCIP functions be wrapped to make implementations of custom presolvers more straightforward? This would reduce the efforts needed from PySCIPOpt users to inspect SCIP's source code.

PR #1142, opened 2025-12-28
Automating the generation of type stubs

Joao-Dionisio:
@jonathanberthias when you have the time, can you please take a look? I deleted `scip.pyi` and ran the script, generating the `scip.pyi` that's in this PR.

I also wrapped `SCIPaddClique`, and re-ran the script. It did add the stub to `scip.pyi`.


src/pyscipopt/scip.pyi
```diff
@@ -1307,35 +1309,6 @@ class Statistics:
     primal_bound: float = None
     dual_bound: float = None
     gap: float = None
-    primal_dual_integral: float = None
```
Joao-Dionisio: It has some problems with this statistics class, but this is going to be deprecated in favor of the JSON statistics

scripts/generate_stubs.py
```diff
@@ -0,0 +1,807 @@
...
+        elif '__init__' not in cls_info.methods and not is_numpy_subclass and not cls_info.is_dataclass and cls_info.name not in skip_init_classes:
+            lines.append(f'    {self.SPECIAL_METHODS["__init__"]}')
+
+        # Sort and output regular methods
+        for method in sorted(regular_methods):
+            if method in cls_info.static_methods:
+                lines.append('    @staticmethod')
+                lines.append(f'    def {method}(*args, **kwargs): ...')
+            else:
+                lines.append(f'    def {method}(self, *args, **kwargs): ...')
```
jonathanberthias, line 600: The script hardcodes the arguments to `*args, **kwargs` which won't work anymore once we start putting the exact arguments in the stubs.

Joao-Dionisio: ah yes, this will need to be revisited once we have the arguments, but I'm fairly confident that the AI will manage to incorporate the new types. But, in any case, until we do add the types, I wouldn't be too worried about merging this.

Joao-Dionisio:
removed the clique stuff, the tests weren't wroking and this is more important. The tests should be easier to add once #1076 is merged

Issue #605, opened 2022-07-28
Question: Implement Constraint Handler properly


SanielDous:
I am trying to implement a Branch-and-Price algorithm. Pricing new variables already works fine but I have some issues when it comes to branching. I already came up with a branching rule that adds new constraints to the child nodes and thereby ensures integrality without branching on the variables. My problem is now to actually add these constraints in the implementation:

This is probably the 1000th time this question is asked and there are some useful hints [here](https://github.com/scipopt/PySCIPOpt/issues/395) and [here](https://github.com/scipopt/PySCIPOpt/blob/use-probdata-to-store-model/tests/test_coloring.py) but I still don't understand how to properly implement a branching rule that adds constraints to the child nodes like:

    quicksum(var[i] for i in X) <= C

for one child node and

    quicksum(var[i] for i in X) >= C + 1

for the other child node. (So, really simple constraints)
[This](https://github.com/scipopt/PySCIPOpt/pull/418) PR tried to make life easier by allowing to use this for example:

    self.model.addConsNode(child1, quicksum(var[i] for i in X) >= C + 1)

I don't know if these changes already found its way into master but I still get the following error:

    TypeError: Argument 'cons' has incorrect type (expected pyscipopt.scip.Constraint, got pyscipopt.scip.ExprCons)

So, I understood I have to use a Constraint Handler to add the new constraints. I tried to understand the examples from [this](https://github.com/scipopt/PySCIPOpt/issues/395) issue but I don't get it where the constraints are actually added. In other words where to put the Expression from above. The function `self.model.createCons(self, name, ...)` does not get such an expression. How does the node's model know the new row?

Probably the solution is completely obvious and I just don't see it. Nevertheless, I would appreciate your tips and hints.

Joao-Dionisio:
Hello, @SanielDous! Sorry for leaving you hanging for so long, your problem seemed out of my league, but now I think I may help a bit.

Are you still having problems? What's the status of this?

SanielDous:
Hi, thank you for your response but I'm not working on the corresponding project anymore.

Joao-Dionisio:
I am very sorry for not helping when it was needed. I'm closing the issue then, but if you or anyone else is having similar problems, feel free to reopen the issue.

Issue #580, opened 2022-03-21
Branch and Price Constraint Handler


thksid:
Hi guys, I had a quick question on implementing branch and price using pySCIPOpt. The issue https://github.com/scipopt/PySCIPOpt/issues/395 was very helpful in explaining how to implement branch and price in pySCIPOpt. 

My pricer and branch rule both seem to work properly. My only issue is with the constraint handler. From what I understand, after the branching is done (i.e. create child constraint nodes), the model should enter one of the child nodes and go to the "consactive" method of the constraint handler.  However, this does not happen for me. After my branching, the model goes back to the pricer. If I check "self.model.data['branchingCons']", I notice that the branching constraints are added but none of the constraints are active (I check this using the .isActive() method). 

How can I ensure that the model goes to the constraint handler after the branching is done? 

-Thanks!

SanielDous:
Did you add the constraint created by `self.model.createCons()` in the constraint handler to the node by calling `self.model.addConsNode(child1, cons)`? I forgot this at first and non of the methods of the constraint handler were called. After adding it worked.

Additionally, you have to make sure to add the ConstraintHandler to the model: `model.includeConshdlr()`

Joao-Dionisio:
I'm closing this issue due to inactivity @thksid, but if it's still relevant feel free to reopen it!

Issue #391, opened 2020-06-18
Can I use ExprCons for model.addConsNode()?


purin0604:
Hi,

I would like to add some constraints locally using model.addConsNode().
(Set some variables to be zero. )
From coloring examples and documents, I learned it should be something like:

```
Class myConshdlr(Conshdlr):

    def consprop(self, constraints, nusefulconss, nmarkedconss, proptiming):
         for var in self.model.vars:      
               if condition is True:
                  self.model.chgVarUb(var, 0.0)
        return {"result": SCIP_RESULT.DIDNOTFIND}

    # fundamental callbacks 
    def consenfolp(self, constraints, nusefulconss, solinfeasible):
        return {"result": SCIP_RESULT.FEASIBLE}
    def conscheck(self, constraints, solution, checkintegrality, checklprows, printreason):
        return {"result": SCIP_RESULT.FEASIBLE}
    def conslock(self, constraint, nlockspos, nlocksneg):
        return {"result": SCIP_RESULT.FEASIBLE}

con1 = model.createCons(myConshdlr, "my constraint", stickingatnode=True)
model.addConsNode(node, con1) 

```

1) Is using constraint handler the only way to create constraints locally and use addConsNode()?
Is there any simpler way to add constraint at node, like using constraint expression in addCons()?
Such as
```
         for var in self.model.vars:      
               if condition is True:
                  model.addConsNode(var == 0)
```

2) What does  "stickingatnode" parameter in the model.createCons() do? 

Thank you

fserra:
Hi

you can read about stickingatnode [here](https://www.scipopt.org/doc/html/FAQ.php#pricerandnodedata). 

Regarding your other question... mmhmm. Maybe I am wrong here (@mattmilten !) but it seems to me that one can only use a constraint handler. In SCIP you can create a constraint and then add it. In PySCIPOpt, when you create an ExprCons, the only way to get a `Constraint` object from it is (as far as I can see) using `addCons`, which will create the constraint and add it.
Since, `addConsNode` uses a `Constraint` object, you can't call it with an ExprCons.

Mmm, so I think we have `addCons` like that, because we didn't want the users to have to create a constraint an then add it.
Maybe we can add another function to create a constraint? Or should we extend `addConsNode`, `addConsLocal` to also accept ExprCons?
Or am I missing something?

purin0604:
Thank you for the prompt answer.
Making sure the ExprCons is only used in addCons was enough for me for now :)
[...]

PR #1151, opened 2026-01-14
Change `addConsLocal()`, `addConsNode()` to accept ExprCons

Joao-Dionisio:
There were a bunch of issues and PRs related to `addConsNode()` not behaving as expected (#391, #580, #605), and the current implementation forces users to use an event handler whenever they want to implement a slightly more complicated branching decision. In this event handler, they then use `addConsLocal()`

This is a breaking change, but I feel like it should be done.
A middle ground would be to accept both `ExprCons` and `Constraint`, but it doesn't feel very elegant. `addCons()` doesn't do this. Any thoughts, @mmghannam ?


mmghannam:
So my thoughts are: this is great! I don't even know how it can be used if the input is a `Constraint` so I wouldn't worry too much that it's a breaking change.

PR #1145, opened 2026-01-04
Use real parameter names in stubs

jonathanberthias:
Following #1141, this PR fixes all the errors related to missing or extra function parameters in the stubs! It also fixes some messages related to [`disjoint_base`](https://peps.python.org/pep-0800/) since `stubtest` reports them when missing.

Using a script, all the `*args, **kwargs` parameters introduced by `stubgen` were replaced with the real parameter names. Parameters with default values are also marked with a default of `...`.

This makes for a very large PR, but there shouldn't be much to check other than the output of `stubtest`. To test it out, you can install the version from this branch and any Python Language Server should be able to pick up the names, and provide suggestions when filling in the arguments to a function call:
image

For now, all the parameters and return types are annotated with `Incomplete` which is equivalent to `typing.Any`. We just have to replace all of those with the appropriate types - only 2010 to go!

PR #1152, opened 2026-01-14
Improve stub generating script and add disabled-by-default workflow for regenerating stubs

Joao-Dionisio:
@jonathanberthias how do you feel about this? I'm a bit more concerned since, now that the script would be running automatically, we should be more comfortable that it's actually correct.

Added `isObjIntegral()` to test that the pipelines do what they're supposed to.

### DIFF
--- a/.github/workflows/stubs.yml
+++ b/.github/workflows/stubs.yml
@@ -1,10 +1,6 @@
 name: Check type stubs
 env:
   version: 10.0.0
-  # TODO: Set to 'true' once generate_stubs.py has proven reliable over several PRs.
-  # This will enable automatic stub regeneration and commit when stubtest fails.
-  # For now, use the pre-commit hook (.pre-commit-config.yaml) for local development.
-  AUTO_REGENERATE_STUBS: 'false'
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


src/pyscipopt/scip.pyi
jonathanberthias, line 1: This stub version is a regression from before, we lost all the names of parameters from #1145. We can try to automate the stubs for newly added functions, but we shouldn't overwrite what's already there

Joao-Dionisio: Ah yeah, this is still being adapted. But on the overall approach, assuming things work out, how do you feel? Regarding the script being called automatically by the pipelines whenever the stub tests fails

Joao-Dionisio:
Heck yeah!

src/pyscipopt/scip.pyi
Joao-Dionisio: By the way, @jonathanberthias, this is working now, as far as I can see, minus some other things we might need.

Joao-Dionisio:
Alright, commented out the workflow and added a githook. So, this PR was mostly about improving the script and not much else. Once the script is more bulletproof, I'd still like to enable it.

PR #1161, opened 2026-01-18
Fix free transform segfault

### LINKED ISSUE #604: model.freeTransform() breaks debugging in VsCode
Using a simple example, diet.py when I introduce model.freeTransform() in line 98, the debugging in vscode starts to behave in unexpected ways. Placing a breakpoint between lines 97 and 102 works well for the first iteration of the loop, but for the second iteration the code stops at the breakpoint for a few tenths of a second, and the debugging console says something like "Error [1]: server disconnected unexpectedly". If I remove model.freeTransform(), this issue does not arise. I know that it does not make sense to use it here, but this is a behavior that I have encountered several times within my own code. This is important because it is not unusual that PySCIPOpt terminates without an error message, and so debugging is crucial. I am unsure if this is an issue with my computers (it happens in both), or with vscode itself, but it seems that freeTransform() is the culprit. Below you can find the code to, hopefully, reproduce the problem. Some pictures detail the error as well.

[Here is the code. diet.py.zip](https://github.com/scipopt/PySCIPOpt/files/9142080/diet.py.zip)

First iteration of the loop works fine

image

Second iteration we get this error

image

Joao-Dionisio:
Fix #604 , the oldest issue in the repo. Basically, after `freeTransform()`, VsCode would try to access freed memory, via Python's `__repr__`. This fix caches the names of variables and constraints to prevent the problem.

### DIFF
--- a/src/pyscipopt/scip.pxi
+++ b/src/pyscipopt/scip.pxi
@@ -8214,10 +8214,9 @@ cdef class Model:
             constraint to be deleted
 
         """
+        del self._modelconss[cons.ptr()]
         PY_SCIP_CALL(SCIPdelCons(self._scip, cons.scip_cons))
         # Remove from tracking and invalidate pointer. See issue #604.
-        ptr = cons.ptr()
-        del self._modelconss[ptr]
         cons.scip_cons = NULL
 
     def delConsLocal(self, Constraint cons):


DominikKamp:
This looks like an unnecessary overhead. Should the variable not be captured to avoid dangling pointers? Or maybe just reset the variable pointers in free methods and check this state in `__get__()` and everywhere else.

Joao-Dionisio:
That seems like a lot more work. And at least Claude doesn't seem to enthusiastic. Regarding variable capture and release:

> when Python garbage collects Variables after the Model is freed, SCIPreleaseVar(self._scip, ...) fails because the SCIP instance is already gone. This caused errors during cleanup. Could potentially be solved by storing a reference to the Model (not raw SCIP*) so the Model stays alive while Variables exist, but that creates other complexities (circular references, preventing Model cleanup).

It seems like a lot of work will be needed. How do you feel about having this short-term solution and keeping the proper fix in mind for later? Otherwise, the bug will remain there for a lot longer, I don't have the bandwith.

DominikKamp:
The model destructor could release the variables before freeing the model, but this is a conventional change. Simple fix would be to declare the transformed variables as freed by setting the stored variable pointer to `NULL` in `freeTransform()` before restricting the model variables to the original variables. Then it is just not allowed to call methods on these freed variables (same as currently), which could be safeguarded by additional checks, so the described issue is just that PySCIPOpt uses an invalid memory address.

Joao-Dionisio:
What about now @DominikKamp ?

src/pyscipopt/scip.pxi
```diff
@@ -1573,6 +1573,8 @@ cdef class Variable(Expr):
         return <size_t>(self.scip_var)
 
     def __repr__(self):
+        if self.scip_var == NULL:
+            return "<freed Variable>"
```
DominikKamp: Should be done in the name `__get__()` method above. Maybe better return an empty string to not match an existing name. Same for constraints.

src/pyscipopt/scip.pxi
```diff
@@ -3048,11 +3066,20 @@ cdef class Model:
...
+            if ptr not in original_var_ptrs:
+                (<Variable>var).scip_var = NULL
+        self._modelvars = {ptr: var for ptr, var in self._modelvars.items() if ptr in original_var_ptrs}
+
+        # Invalidate transformed constraints. See issue #604.
+        original_cons_ptrs = {ptr for ptr, cons in self._modelconss.items() if cons.isOriginal()}
+        for ptr, cons in self._modelconss.items():
+            if ptr not in original_cons_ptrs:
+                (<Constraint>cons).scip_cons = NULL
+        self._modelconss = {ptr: cons for ptr, cons in self._modelconss.items() if ptr in original_cons_ptrs}
```
```suggestion
        # Invalidate transformed variables. See issue #604.
        origvars = {ptr: var for ptr, var in self._modelvars.items() if var.isOriginal()}
        for ptr, var in self._modelvars.items():
            if ptr not in origvars:
                (<Variable>var).scip_var = NULL
        self._modelvars = origvars

        # Invalidate transformed constraints. See issue #604.
        origconss = {ptr: cons for ptr, cons in self._modelconss.items() if cons.isOriginal()}
        for ptr, cons in self._modelconss.items():
            if ptr not in origconss:
                (<Constraint>cons).scip_cons = NULL
        self._modelconss = origconss
```

src/pyscipopt/scip.pxi
Joao-Dionisio: Done. How do you feel about Copilot's complaint on other methods? I wouldn't be averse to a `Constraint.create()` wrapper.

src/pyscipopt/scip.pxi
```diff
@@ -2941,6 +2959,18 @@ cdef class Model:
...
+        """Create a Constraint wrapper and track it for proper cleanup."""
+        pyCons = Constraint.create(scip_cons)
+        self._modelconss[pyCons.ptr()] = pyCons
+        return pyCons
+
+    cdef _createVar(self, SCIP_VAR* scip_var):
+        """Create a Variable wrapper and track it for proper cleanup."""
+        pyVar = Variable.create(scip_var)
+        self._modelvars[pyVar.ptr()] = pyVar
+        return pyVar
```
```suggestion
    cdef _createCons(self, SCIP_CONS* scip_cons):
        """Create a Constraint wrapper and track it for proper cleanup."""
        pyCons = Constraint.create(scip_cons)
        assert pyCons.ptr() not in self._modelconss
        self._modelconss[pyCons.ptr()] = pyCons
        return pyCons

    cdef _createVar(self, SCIP_VAR* scip_var):
        """Create a Variable wrapper and track it for proper cleanup."""
        pyVar = Variable.create(scip_var)
        assert pyVar.ptr() not in self._modelvars
        self._modelvars[pyVar.ptr()] = pyVar
        return pyVar
```

src/pyscipopt/scip.pxd
```diff
@@ -2222,12 +2222,17 @@ cdef class Model:
...
+    cdef _modelconss
     # used to keep track of the number of event handlers generated
     cdef int _generated_event_handlers_count
     # store references to Benders subproblem Models for proper cleanup
     cdef _benders_subproblems
     # store iis, if found
     cdef SCIP_IIS* _iis
+    # helper methods for later var and cons cleanup
+    cdef _createCons(self, SCIP_CONS* scip_cons)
+    cdef _createVar(self, SCIP_VAR* scip_var)
```
DominikKamp: Why is this required?

Joao-Dionisio: Doesn't compile otherwise. `_createVar(), _createCons()` are `cdef` functions.

src/pyscipopt/scip.pxi
Joao-Dionisio: Nice catch, but the fix wasn't passing the tests. Changed the methods to first try getting the ptr and only then creating the variable.

Joao-Dionisio:
@DominikKamp can it be merged, then?

src/pyscipopt/scip.pxi
DominikKamp: When is this necessary, so why should something want to create duplicate variables and constraints?

Joao-Dionisio: Say you call `getTransformedVar(var_x)`. It first needs to create the PySCIPOpt variable. If you call `getTransformedVar(var_x)` again, it will see that the associated variable was already created, and return that instead.

This check could be made directly on `getTransformedVar()` and other such methods, but I'd vote in favor of keeping it contained.

DominikKamp: And why do we want to add the transformed counterpart to the model variables already containing the original variable (would this not increase the number of model variables unexpectedly?).

Joao-Dionisio: Yeah, I suppose it shouldn't be named origvars. But afaik, it's not used to determine the number of variables in the model (I'm on mobile now)

In any case, the same would happen by calling getVars twice

Joao-Dionisio: Yeah, `_modelvars` is only accessed via the pointers. For example:

```python
def getVars(self, transformed=False):
        """
        Retrieve all variables.

        Parameters
        ----------
        transformed : bool, optional
            Get transformed variables instead of original (Default value = False)

        Returns
        -------
        list of Variable

        """
        cdef SCIP_VAR** _vars
        cdef int nvars
        cdef int i

        vars = []
        if transformed:
            _vars = SCIPgetVars(self._scip)
            nvars = SCIPgetNVars(self._scip)
        else:
            _vars = SCIPgetOrigVars(self._scip)
            nvars = SCIPgetNOrigVars(self._scip)

        for i in range(nvars):
            ptr = <size_t>(_vars[i])

            # check whether the corresponding variable exists already
            if ptr in self._modelvars:
                vars.append(self._modelvars[ptr])
            else:
                # create a new variable
                vars.append(self._getOrCreateVar(_vars[i]))

        return vars
```

The only time we iterate through `_modelvars` is to `__dealloc__()`.

src/pyscipopt/scip.pyi
```diff
@@ -2181,6 +2182,6 @@ class Variable(Expr):
     def isOriginal(self) -> Incomplete: ...
     def isRelaxationOnly(self) -> Incomplete: ...
     def markRelaxationOnly(self) -> Incomplete: ...
-    def ptr(self) -> Incomplete: ...
+    def ptr(self) -> int: ...
```
DominikKamp: What is this required for?

Joao-Dionisio: No reason, I'll revert.

src/pyscipopt/scip.pxi
```diff
@@ -8214,6 +8241,11 @@ cdef class Model:
 
         """
         PY_SCIP_CALL(SCIPdelCons(self._scip, cons.scip_cons))
+        # Remove from tracking and invalidate pointer. See issue #604.
+        ptr = cons.ptr()
+        if ptr in self._modelconss:
+            del self._modelconss[ptr]
```
```suggestion
        del self._modelconss[ptr]
```
If there is a constraint which we can delete, it should exist in the array.

src/pyscipopt/scip.pxi
```diff
@@ -6574,15 +6607,9 @@ cdef class Model:
...
-            # create a new variable
-            resultant = Variable.create(_resultant)
-            assert resultant.ptr() == ptr
-            self._modelvars[ptr] = resultant
-        else:
-            resultant = self._modelvars[ptr]
-            
-        return resultant
+        if ptr in self._modelvars:
+            return self._modelvars[ptr]
```
DominikKamp: This should not be necessary.

src/pyscipopt/scip.pxi
```diff
@@ -4432,6 +4472,8 @@ cdef class Model:
         if var.ptr() in self._modelvars:
             del self._modelvars[var.ptr()]
```
```suggestion
        del self._modelvars[var.ptr()]
```

src/pyscipopt/scip.pxi
Joao-Dionisio: How come? We can call `getResultantAnd()` multiple times, no?

src/pyscipopt/scip.pxi
```diff
@@ -2193,9 +2195,14 @@ cdef class Constraint:
 
     property name:
         def __get__(self):
+            if self.scip_cons == NULL:
+                return ""
             cname = bytes( SCIPconsGetName(self.scip_cons) )
             return cname.decode('utf-8')
 
+    def ptr(self):
+        return <size_t>(self.scip_cons)
```
DominikKamp, line 2204: Can we remove the empty docstring for variables?

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 - Model.getVal now supports GenExpr type
 - Fixed lotsizing_lazy example
 - Fixed incorrect getVal() result when _bestSol.sol was outdated
+- Fixed segfault when accessing Variable or Constraint objects after freeTransform() or Model destruction
```
```suggestion
- Fixed segmentation fault when using Variable or Constraint objects after freeTransform() or Model destruction
```

src/pyscipopt/scip.pxi
Joao-Dionisio: Done

src/pyscipopt/scip.pxi
DominikKamp: I mean that this done in `self._getOrCreateVar()`.

src/pyscipopt/scip.pxi
```diff
@@ -8214,6 +8240,10 @@ cdef class Model:
 
         """
         PY_SCIP_CALL(SCIPdelCons(self._scip, cons.scip_cons))
+        # Remove from tracking and invalidate pointer. See issue #604.
+        ptr = cons.ptr()
+        del self._modelconss[ptr]
```
```suggestion
        del self._modelconss[cons.ptr()]
        PY_SCIP_CALL(SCIPdelCons(self._scip, cons.scip_cons))
        # Remove from tracking and invalidate pointer. See issue #604.
```

src/pyscipopt/scip.pxi
Joao-Dionisio: True, fixed elsewhere as well.

DominikKamp:
Otherwise, this looks fine, probably `NULL` handlings should be added to more methods but using deleted objects is just not supported, however, the master branch should be fixed before merging anything.

Joao-Dionisio:
> however, the master branch should be fixed before merging anything.

This is a problem with a recent merge not being compatible with numpy 1.16. It's not really relevant (numpy 1.16 is 7 years old), and is actively being worked on in #1164

PR #1174, opened 2026-01-23
Add `getPrimalDualIntegral()` from C API

justinokamoto:
Adds the primal-dual integral statistic from the C API. It's one of the primary metrics my team uses when benchmarking different solving approaches, and I think it would be useful to expose in the Python API as well.

* Expose [SCIPgetPrimalDualIntegral()](https://www.scipopt.org/doc/html/group__PublicSolvingStatsMethods.php#gaf5c76d106165d3946fb860ae5f6eba24)
* Add test
* Update CHANGELOG.md

NOTE: Wasn't sure where to add the test. `tests/test_statistics.py` seemed most appropriate. But lmk if I should move it.


Joao-Dionisio:
Oh man, these are my favorite PRs, where I don't need to do anything :) Thank you so much for the contribution!

justinokamoto:
Thanks for the quick review!

PR #1158, opened 2026-01-17
Add setTracefile() method for structured optimization progress loggingAdd settracefile api

MySweetEden:
## Summary
Adds `setTracefile()` method for structured, machine-readable optimization progress logging.

Closes #1147

## Motivation
As discussed in #1147, structured progress logging is useful for:
- Visualization of optimization progress
- Fair comparison of solver runs
- Analysis of primal/dual bound evolution

This provides a simpler alternative to implementing custom event handlers.

## Design Decisions
- **JSONL format**: Allows streaming writes and partial reads even if optimization crashes
- **Separate from setLogfile()**: Human-readable log and machine-readable trace serve different purposes
- **Append mode by default**: Multiple optimize() calls can accumulate in one file
- **Automatic close**: Trace file is managed internally, no explicit close needed
- **Field names match SCIP log**: `primalbound`, `dualbound`, `time`, `nodes` for consistency

## Changes
- Added `setTracefile(path, mode="a")` method to Model
- Internal `_write_trace_event()` for centralized trace writing
- Event handler for `BESTSOLFOUND` events

## Events Recorded
- `solution_update`: when a new best solution is found
- `solve_finish`: when optimization terminates

## Fields
`type`, `time`, `primalbound`, `dualbound`, `gap`, `nodes`, `nsol`

## Usage
```python
model.setTracefile("trace.jsonl")
model.optimize()
```
trace.jsonl contains JSONL records

## Future Work
- Recipe/tutorial documentation will be added in a follow-up PR if this implementation is accepted

## Open Questions
- Should a `solve_start` event be added to distinguish multiple optimize() calls in append mode?
- Output format: JSONL is chosen for streaming/crash-safety, but CSV/TSV could be discussed

PR #1177, opened 2026-01-24
Extend structured_optimization_trace with real-time JSONL streaming

MySweetEden:
### Motivation
PySCIPOpt already has recipe(s) that store optimization progress in memory. However, in-memory traces are not suitable for **real-time, external observation** (e.g., another process tailing progress, dashboards, log collectors).

This recipe focuses on the missing piece: **a stream-friendly, structured output that can be consumed outside the running Python process during solve**.

### Design Decisions
- **JSONL format**: Designed for streaming writes and partial reads; remains readable even if the run is interrupted or crashes
- **Real-time external output is the primary value**:
  - Records progress updates as one JSON object per line
  - Flushes on key events so downstream consumers can react immediately
- **Schema compatibility with `setTracefile()` (PR #1158)**:
  - Uses the same JSONL field names (`type`, `time`, `primalbound`, `dualbound`, `gap`, `nodes`, `nsol`) for consistency across tracing approaches.
- **In-memory + file output**: Keeps `model.data["trace"]` for convenience/testing, but the recipe is centered on file streaming via `path=...`
- **Robust termination signaling for external monitoring**:
  - Always emits a final `run_end` record on normal termination, interruption, or exception
  - On exceptions, `run_end` includes structured error metadata (`status`, exception type, message)
  - Flushes `run_end` to make completion detection reliable

### Events Recorded
- `bestsol_found`: when a new best solution is found
- `dualbound_improved`: when the dual bound improves
- `run_end`: when optimization terminates (also emitted on interrupt/exception)

### Fields
`type, time, primalbound, dualbound, gap, nodes, nsol` (aligned with the JSONL trace schema introduced in PR #1158)  
(`run_end` may additionally include: `status, exception, message` on failure)


MySweetEden:
I’ll address the comments over the weekend and push updates soon.

MySweetEden:
## Addressed

### 1. `dropEvent()` refcount underflow prevention
SCIP does not provide a dedicated "solve finished" event, so we cannot rely on an event-driven shutdown callback (like `eventexit()`) for cleanup. Therefore, cleanup is performed in `__exit__`, with guards to avoid invalid `dropEvent()` calls.

**Implementation**:
- Added `self._caught_events` (set) to track which events were successfully caught in `eventinit()`
- Modified `__exit__` to drop only the events that were actually caught
- This prevents incorrect `dropEvent()` calls if `eventinit()` partially fails

### 2. Trace data initialization
Changed from `setdefault()` to explicit assignment:
```python
self.model.data["trace"] = []
```
**Rationale**: Ensures each traced run starts with a fresh list, preventing data from previous runs from mixing and keeping in-memory behavior consistent with file output (which always truncates).

### 3. Event handler parameter naming
Changed the first parameter from `s` to `hdlr` to improve readability while avoiding collision with the outer `_TraceRun.self`.  
**Note**: `hdlr` is a conventional abbreviation commonly used in PySCIPOPT/SCIP codebase.

### 4. Exception handling and flush comments
Added explanatory comments to exception handling and flush behavior for clarity.

---

## Not Addressed

### 1. `includeEventhdlr()` multiple invocation issue
The concern about permanent handler registration is valid:
- `includeEventhdlr()` registers handlers permanently with no removal API
- `dropEvent()` only unsubscribes from events, not the handler itself

**Scope**: Refactoring to a handler-reuse pattern would require architectural changes and will be addressed separately. The current implementation assumes single-run usage.

### 2. `dualbound_improved` flush policy
`dualbound_improved` events are intentionally not flushed:

- **Frequency asymmetry**: `dualbound_improved` fires hundreds to thousands of times during optimization, while `bestsol_found` fires only a few dozen times at most; flushing on every dual bound update would accumulate significant I/O overhead
- **OS buffering suffices**: Events naturally flush within seconds via OS buffering, providing adequate real-time visibility
- **Context**: Optimizations typically run for minutes to hours, making second-scale buffering delays negligible

**Discussion**: I'm open to reconsidering the flush policy if there are use cases where immediate flushing of `dualbound_improved` events is valuable (e.g., sub-minute monitoring). Would making it configurable be useful, or is the current approach acceptable?

src/pyscipopt/recipes/realtime_trace_jsonl.py
```diff
@@ -0,0 +1,135 @@
...
+                    snapshot = self._snapshot_now()
+                    self._last_snapshot = snapshot
+                    self._write_event(
+                        "dualbound_improved", fields=snapshot, flush=False
+                    )
+
+        self._handler = _TraceEventhdlr()
+        self.model.includeEventhdlr(
+            self._handler, "realtime_trace_jsonl", "Realtime trace jsonl handler"
+        )
```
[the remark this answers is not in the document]
MySweetEden: ## Not Addressed
### 1. `includeEventhdlr()` multiple invocation issue
The concern about permanent handler registration is valid:
- `includeEventhdlr()` registers handlers permanently with no removal API
- `dropEvent()` only unsubscribes from events, not the handler itself

**Scope**: Refactoring to a handler-reuse pattern would require architectural changes and will be addressed separately. The current implementation assumes single-run usage.

src/pyscipopt/recipes/realtime_trace_jsonl.py
```diff
@@ -0,0 +1,135 @@
...
+
+            if self._handler is not None:
+                for et in (
+                    SCIP_EVENTTYPE.BESTSOLFOUND,
+                    SCIP_EVENTTYPE.DUALBOUNDIMPROVED,
+                ):
+                    try:
+                        self.model.dropEvent(et, self._handler)
+                    except Exception:
+                        pass
```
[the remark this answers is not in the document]
MySweetEden: ## Addressed
### 1. `dropEvent()` refcount underflow prevention
SCIP does not provide a dedicated "solve finished" event, so we cannot rely on an event-driven shutdown callback (like `eventexit()`) for cleanup. Therefore, cleanup is performed in `__exit__`, with guards to avoid invalid `dropEvent()` calls.

**Implementation**:
- Added `self._caught_events` (set) to track which events were successfully caught in `eventinit()`
- Modified `__exit__` to drop only the events that were actually caught
- This prevents incorrect `dropEvent()` calls if `eventinit()` partially fails

src/pyscipopt/recipes/realtime_trace_jsonl.py
```diff
@@ -0,0 +1,135 @@
...
+                if et == SCIP_EVENTTYPE.BESTSOLFOUND:
+                    snapshot = self._snapshot_now()
+                    self._last_snapshot = snapshot
+                    self._write_event("bestsol_found", fields=snapshot, flush=True)
+                elif et == SCIP_EVENTTYPE.DUALBOUNDIMPROVED:
+                    snapshot = self._snapshot_now()
+                    self._last_snapshot = snapshot
+                    self._write_event(
+                        "dualbound_improved", fields=snapshot, flush=False
+                    )
```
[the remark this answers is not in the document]
MySweetEden: ## Not Addressed
### 2. `dualbound_improved` flush policy
`dualbound_improved` events are intentionally not flushed:

- **Frequency asymmetry**: `dualbound_improved` fires hundreds to thousands of times during optimization, while `bestsol_found` fires only a few dozen times at most; flushing on every dual bound update would accumulate significant I/O overhead
- **OS buffering suffices**: Events naturally flush within seconds via OS buffering, providing adequate real-time visibility
- **Context**: Optimizations typically run for minutes to hours, making second-scale buffering delays negligible

**Discussion**: I'm open to reconsidering the flush policy if there are use cases where immediate flushing of `dualbound_improved` events is valuable (e.g., sub-minute monitoring). Would making it configurable be useful, or is the current approach acceptable?

src/pyscipopt/recipes/realtime_trace_jsonl.py
```diff
@@ -0,0 +1,135 @@
...
+        self.path = path
+        self._fh = None
+        self._handler = None
+
+        self._last_snapshot = {}
+
+    def __enter__(self):
+        if not hasattr(self.model, "data") or self.model.data is None:
+            self.model.data = {}
+        self.model.data.setdefault("trace", [])
```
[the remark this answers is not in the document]
MySweetEden: ## Addressed
### 2. Trace data initialization
Changed from `setdefault()` to explicit assignment:
```python
self.model.data["trace"] = []
```
**Rationale**: Ensures each traced run starts with a fresh list, preventing data from previous runs from mixing and keeping in-memory behavior consistent with file output (which always truncates).

src/pyscipopt/recipes/realtime_trace_jsonl.py
```diff
@@ -0,0 +1,135 @@
...
+
+        if self.path is not None:
+            self._fh = open(self.path, "w")
+
+        class _TraceEventhdlr(Eventhdlr):
+            def eventinit(s):
+                self.model.catchEvent(SCIP_EVENTTYPE.BESTSOLFOUND, s)
+                self.model.catchEvent(SCIP_EVENTTYPE.DUALBOUNDIMPROVED, s)
+
+            def eventexec(s, event):
```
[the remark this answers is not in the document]
MySweetEden: ## Addressed
### 3. Event handler parameter naming
Changed the first parameter from `s` to `hdlr` to improve readability while avoiding collision with the outer `_TraceRun.self`.  
**Note**: `hdlr` is a conventional abbreviation commonly used in PySCIPOPT/SCIP codebase.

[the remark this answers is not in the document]
MySweetEden: ## Addressed
### 3. Event handler parameter naming
Changed the first parameter from `s` to `hdlr` to improve readability while avoiding collision with the outer `_TraceRun.self`.  
**Note**: `hdlr` is a conventional abbreviation commonly used in PySCIPOPT/SCIP codebase.

src/pyscipopt/recipes/realtime_trace_jsonl.py
```diff
@@ -0,0 +1,135 @@
...
+                    self._fh = None
+
+            if self._handler is not None:
+                for et in (
+                    SCIP_EVENTTYPE.BESTSOLFOUND,
+                    SCIP_EVENTTYPE.DUALBOUNDIMPROVED,
+                ):
+                    try:
+                        self.model.dropEvent(et, self._handler)
+                    except Exception:
```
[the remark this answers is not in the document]
MySweetEden: ## Addressed
### 4. Exception handling and flush comments
Added explanatory comments to exception handling and flush behavior for clarity.

MySweetEden:
I addressed the actionable review items and all checks are green. A couple of higher-level/trade-off points are intentionally left open for discussion. Could you take another look when you have time?

Joao-Dionisio:
Hey @MySweetEden , yes I will have a look! I will try to lay low for a little bit, for my own sake, but this should get merged, don't worry :)

MySweetEden:
Thanks! Take your time. I appreciate you having a look 🙂

Joao-Dionisio:
Hey @MySweetEden ! Is there a reason for this being a separate recipe, instead of an update on the other one you added?

tests/test_recipe_realtime_trace_jsonl.py
```diff
@@ -0,0 +1,87 @@
...
+    def eventinit(self):
+        self.model.catchEvent(SCIP_EVENTTYPE.BESTSOLFOUND, self)
+
+    def eventexec(self, event):
+        self.model.interruptSolve()
+
+
+def test_optimize_with_trace_records_run_end_on_interrupt(optimize):
+    model = bin_packing_model(
+        sizes=[randint(1, 40) for _ in range(120)],
```
Joao-Dionisio: Please don't use randomness in the tests

src/pyscipopt/recipes/realtime_trace_jsonl.py
```diff
@@ -0,0 +1,139 @@
...
+        finally:
+            if self._fh:
+                try:
+                    self._fh.close()
+                finally:
+                    self._fh = None
+
+            if self._handler is not None:
+                for et in self._caught_events:
+                    try:
```
Joao-Dionisio: Please don't use `try/Except` blocks

tests/test_recipe_realtime_trace_jsonl.py
MySweetEden: Thanks for pointing this out. I replaced the random input with a fixed deterministic one.
I kept the change scoped to this file to keep the PR focused.

src/pyscipopt/recipes/realtime_trace_jsonl.py
MySweetEden: Good point. I removed the broad exception handling during cleanup so failures in dropEvent() are no longer silently ignored. I kept the try/finally structure to ensure resources are still cleaned up even if writing the final event fails.

MySweetEden:
> Hey @MySweetEden ! Is there a reason for this being a separate recipe, instead of an update on the other one you added?

Yes. this version is intended for file-based logging, so it manages tracing through completion of the optimization. As a result, its interface diverged enough from the existing implementation that I kept it separate for now.

I’m still considering the best way to handle the previous attachment-style interface, but if this direction seems preferable, I’d also be open to deprecating the older one.

Joao-Dionisio:
Hey @MySweetEden ! But then, does it not make sense to replace the old version? It seems that this is almost a strictly better recipe? It seems that it'd be confusing for users to have two recipes that are trying to solve very similar problems, and also more annoying to maintain.

The only things that would be needed, would be to decouple the handler from the optimization, like in the old version. Or is there a reason for it to be bundled together, for having a method that optimizes inside it?

Also, Copilot's comment on solving the same problem more than once, and thus adding the event handler more than once, seem to make sense. Perhaps it'd be better to first check whether this handler has already been attached.

MySweetEden:
Hi, @Joao-Dionisio ! I agree that we should avoid keeping two very similar recipes. I think we need to be careful about how we unify them.

I agree that the handler should be separated from the optimization logic. At the same time, I think tracing still needs to be managed through finalization, especially for file-based logging. Based on your comment, I think the right direction is a single unified recipe that supports both one-shot usage for the common case and a context-manager style for cases where the user wants more control. The old attach-only style cannot reliably capture the end of the run, so I do not think it should remain as a separate pattern.

I also agree that multiple handler attachment should be guarded against. The difficulty is that there is no clean API to inspect what has already been attached, so preventing duplicate attachment would require this recipe to manage that state itself. My current view is that we should be careful about adding that much state management into the recipe. A context-manager-style interface feels cleaner to me here, with custom handling left to the user when needed.

MySweetEden:
Hi @Joao-Dionisio ! Sorry for the long delay. I’ve now updated the PR based on your comments.

I folded the realtime_trace_jsonl recipe into structured_optimization_trace and removed the separate recipe/tests, since the two recipes overlapped significantly.

I also decoupled tracing from optimization. The helper no longer calls `model.optimize()` or `model.optimizeNogil()` internally. Instead, the recipe now supports the existing attach-style API for in-memory tracing:
```python
attach_structured_optimization_trace(model)
model.optimize()
```
and a context-manager API for JSONL tracing:
```python
with structured_optimization_trace(model, path="trace.jsonl"):
    model.optimize()
```
For file-based tracing, the context manager is still useful as a clear finalization scope.

I also addressed the repeated-handler concern by reusing the existing event handler on the model to avoid duplicate registration.

src/pyscipopt/recipes/structured_optimization_trace.py
```diff
@@ -1,37 +1,185 @@
...
+
+        self.model.data["trace"].append(event)
+        if self._fh is not None:
+            self._fh.write(json.dumps(event) + "\n")
+            self._fh.flush()
+
+
+def structured_optimization_trace(model: Model, path=None):
+    """
+    Create a context manager for structured optimization progress tracing.
```
Joao-Dionisio: I feel like most of this docstring would work better at the top of the file

CHANGELOG.md
```diff
@@ -6,7 +6,7 @@
 - Include parameter names in type stubs
 - Added pre-commit hook for automatic stub regeneration (see `.pre-commit-config.yaml`)
 - Wrapped `isObjIntegral()` and test
-- Added `structured_optimization_trace` recipe for structured optimization progress tracking
+- Added `structured_optimization_trace` recipe with attach-style in-memory tracing and context-managed JSONL tracing with final `run_end` records
```
Joao-Dionisio: Rather than replacing the old entry, this should be its own entry (and the CHANGELOG is outdated, I believe)

src/pyscipopt/recipes/structured_optimization_trace.py
```diff
@@ -1,37 +1,185 @@
...
+        model: SCIP Model.
+        path: Optional JSONL output path. If None, records are only stored in
+            ``model.data["trace"]``.
+
+    Usage:
+        with structured_optimization_trace(model, path="trace.jsonl"):
+            model.optimize()
+
+        with structured_optimization_trace(model):
+            model.optimizeNogil()
```
Joao-Dionisio: Why are you talking about `optimizeNogil()` here?

src/pyscipopt/recipes/structured_optimization_trace.py
```diff
@@ -1,37 +1,185 @@
...
 from pyscipopt import SCIP_EVENTTYPE, Eventhdlr, Model
 
+_TRACE_HANDLER_KEY = "_structured_optimization_trace_handler"
+
+
+class _TraceEventhdlr(Eventhdlr):
+    def __init__(self):
+        self.trace = None
+        self.write_run_end_active = False
+        self._caught_events = set()
```
Joao-Dionisio: What happens if we attach the handler to a model, solve it, call something like `freeTransform()`, and solve it again? Does this `_caught_events` get cleared inbetween?

Joao-Dionisio:
A couple more things my Claude flagged:

1. **The two APIs share one lifecycle and interfere.** `attach_structured_optimization_trace` never sets `write_run_end_active`, but the context manager only guards on that flag. So `attach_*(model)`
  followed by `with structured_optimization_trace(model):` passes the guard, resets `model.data["trace"] = []`, and on exit sets `handler.trace = None` — silently wiping the attach trace with no error. Root
  cause is `attach_*` calling `__enter__()` directly (and never `__exit__()`), which also leaks the controller. I'd decouple: a small helper that just includes the handler + inits `model.data["trace"]` for
  attach, with the file/`run_end`/guard lifecycle layered on top only for the context manager.

  2. **`run_end` shape is inconsistent.** When events fired, `__exit__` merges the last snapshot (`primalbound`, `dualbound`, gap, nodes…) into `run_end`; when nothing fired it's just `{type, status}`. The 
  docstring only documents `status`/`exception`/`message`, so consumers can't rely on the record shape. Either always include the snapshot fields or never.

  3. **The reuse test is hollow.** `test_..._reuses_handler_for_repeated_contexts` never calls `optimize()` inside the contexts, so it never re-registers events — it can't actually catch the 
  `freeTransform`-then-resolve case. Worth having it solve twice and assert the second run records events.

  Minor: `_write_event` uses `if fields:` rather than `if fields is not None` — an empty dict would be silently dropped.

src/pyscipopt/recipes/structured_optimization_trace.py
MySweetEden: Done. I moved the shared tracing overview and usage examples to the module docstring, and kept the function docstrings focused on each public API's specific behavior, arguments, and return value.

CHANGELOG.md
MySweetEden: Done. I synced with the latest master and updated the CHANGELOG so the existing entry stays concise, while the JSONL/run_end tracing change is listed as its own unreleased entry.

src/pyscipopt/recipes/structured_optimization_trace.py
MySweetEden: Good point. `optimizeNogil()` is not specific to this recipe, so I removed that example and kept the usage focused on the tracing APIs.

src/pyscipopt/recipes/structured_optimization_trace.py
MySweetEden: Thanks, this was a real issue after `freeTransform()`.
The handler now registers events on each `eventinit()`, so reused handlers are registered again for the next solve. I also added a regression test for `attach -> optimize -> freeTransform -> optimize`.

MySweetEden:
Hi @Joao-Dionisio ! Thanks for flagging these. I addressed them in the latest updates.

1. I separated the attach and context-manager lifecycles, since only the context manager has finalization logic. `attach_structured_optimization_trace()` no longer calls `__enter__()`. If both styles are used on the same model at the same time, it now raises explicitly.

2. I simplified `run_end` records so they only describe finalization: status, plus exception metadata when relevant. Progress snapshots stay on progress events.

3. I updated the repeated context test to run two optimizations with `freeTransform()` in between, so it verifies that the reused handler records progress events again.

4. I changed `_write_event()` to check `fields is not None`, so omitted fields and an explicit empty dict are handled separately.

src/pyscipopt/recipes/structured_optimization_trace.py
```diff
@@ -1,37 +1,202 @@
...
+        self.path = path
+        self.write_run_end = write_run_end
+        self._fh = None
+        self._handler = None
+
+    def __enter__(self):
+        self._handler = _attach_trace_handler(self.model, self)
+
+        if self.path is not None:
+            self._fh = open(self.path, "w", encoding="utf-8")
```
Joao-Dionisio: If there is an issue with opening the file, then this will break every subsequent trace attachment, right? Because we will not exit, and so it'l think that we already have the handler attached.

Perhaps the order of the two should be switched. I think that's the last thing, and that it can be merged afterward.

Joao-Dionisio (approved):

src/pyscipopt/recipes/structured_optimization_trace.py
MySweetEden: Good catch. I’ve switched the order so the file is opened before attaching the handler, avoiding a partially initialized state if opening the file fails.

Joao-Dionisio:
Thank you @MySweetEden ! Sorry for taking so long, and thank you for the ruff fix as well!

PR #1178, opened 2026-01-27
improve chgReoptObjective performance

a-schmitt:
For medium sized production problems using reoptimization `chgReoptObjective` leads to quite some bottleneck.

The reason is some O (m * n) (m=#scip variables, n=#variables with non null objective) loop. 
The current code seems to assume that the scip c function needs all m scip variables as input.  
However, only the n variables to change are needed (see https://github.com/scipopt/scip/blob/775ad66cb8edde846e16d913009b30add5c6d07b/src/scip/scip_prob.c#L1294), reducing the complexity to O ( n ).

Changes:
- Improve the loop
- Correct the doc string
- Add some tests

### DIFF
--- a/src/pyscipopt/scip.pxi
+++ b/src/pyscipopt/scip.pxi
@@ -11862,11 +11862,12 @@ cdef class Model:
             the objective sense (Default value = 'minimize')
 
         """
-        cdef SCIP_VAR** _vars
+        cdef SCIP_VAR** vars
         cdef int nvars
         cdef SCIP_Real* _coeffs
         cdef SCIP_OBJSENSE objsense
         cdef int i
+        cdef _VarArray wrapper
 
         if sense == "minimize":
             objsense = SCIP_OBJSENSE_MINIMIZE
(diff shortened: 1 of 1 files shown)


Joao-Dionisio:
Thank you, @a-schmitt ! This does look very useful! Can you please run the before and after on a bigger problem, to archive the performance improvement?

Something I'm wondering is about the removal of the `_varArray()` wrapper, it seems we are losing some type safety and memory management.

a-schmitt:
> Thank you, @a-schmitt ! This does look very useful! Can you please run the before and after on a bigger problem, to archive the performance improvement?

Thanks for your effort! 

I unscientifically tested on a toy problem
```
def benchmark(n, n_obj=10, n_iters=5):
    m = Model()
    m.enableReoptimization()
    x = [m.addVar() for i in range(n)]

    times = []
    for i in range(n_iters):
        m.freeReoptSolve()
        obj = quicksum(x[j] for j in range(i, n))

        t0 = time.perf_counter()
        m.chgReoptObjective(obj)
        t1 = time.perf_counter()

        times.append((t1 - t0) * 1000)

    return times
```
the proposed version
```
       n    mean_ms   stdev_ms
------------------------------
    1000       0.132      0.016
    5000       0.789      0.165
   10000       1.553      0.092
```

the old version

```
       n    mean_ms   stdev_ms
------------------------------
    1000     134.913      0.976
    5000    3432.795     63.303
   10000   14155.320    173.048
```

so something like ~14s vs ~2ms per objective change for the 10k variable problem and the difference scales with the problem size.

> Something I'm wondering is about the removal of the `_varArray()` wrapper, it seems we are losing some type safety and memory management.

I don't like that the _varArray() wrapper approach uses a malloc for each individual variable.
Collecting the vars in a python list then giving the list to _varArray() is a little bit slower compared to the current suggestion. I added an explicit type check. However, I'm also not super against using the _varArray(). Whatever you prefer.

Joao-Dionisio:
> Collecting the vars in a python list then giving the list to _varArray() is a little bit slower compared to the current suggestion. I added an explicit type check.

True, but the major speedup comes from the better complexity. It just itches a bit not to have a centralized method for memory handling. Plus, it's safer if `SCIPchgReoptObjective()` ever fails.

Can you run another benchmark with this change? If it's significant enough, we can discard it.

a-schmitt:
I added the wrapper within the loop similar how it is done in `_createConsLinear`. The timeloss is not significant. 

I also removed the `assert len(term) == 1`, since there is already a check for the degree of the expression.

Joao-Dionisio:
Thank you, @a-schmitt , and very cool that your first Github contribution was so mature!

PR #1186, opened 2026-02-05
Skip flaky performance tests in CI

Joao-Dionisio:


Joao-Dionisio:
What do you think, @DominikKamp @mmghannam @Opt-Mucca ? This comes after #1168 . Would you be okay with ignoring performance tests in the pipelines and relying on local performance runs? I can also add a github action that would write to the PR whenever these tests were failing, and we could decide if it was relevant.

DominikKamp:
I do not really know what is the point of these one-sided performance tests, but if they matter, increasing the threshold factor to 1.5 might be better than just dropping them completely.

Opt-Mucca:
I don't think these pipelines are intended to have performance tests in them, so I don't see a problem with disabling them.

Joao-Dionisio:
Welp, that's good enough for me! Sorry for the annoyance, @Zeroto521 , and thanks for the input, guys!

Issue #26, opened 2016-07-23
release variable, constraints, rows, etc


fserra:
this is the follow up of issue 55 on gitlab.

In `addVar` we create a SCIP var and then release it. The same happens in `addCons`.  
So the idea is then to have a `__dealloc__`, as in `class Model`, for variables, constraints, rows, etc, right?

fserra:
btw, we have in `class Model`: `def releaseVar`. We should probably organize a cleaning day

fserra:
So one issue is that to call `releaseVar` or `releaseRow` or `releaseCons` you need a `scip` pointer. However, the classes `Variable`, `Constraint`, `Row` don't have a `scip` pointer.  
Should we just have a weakref to Model in those classes?

leethargo:
So your point is that we release the variables too early and might risk that the `Variable` objects becomes broken when `Model` is gc'd?

For cleaning, I suggest removing methods you dislike as long as the tests pass ;-)

I don't understand why one would need a SCIP to release a var, but with a weakref we also risk a segfault, right?

fserra:
I think the problem is that if SCIP decides to remove the variable, then we don't have access to it any more and the object is broken. Benny's original comment was:

>I don't know if this is related to the problem but the SCIPreleaseVar call in model.addVar() seems to be strange (wrong?). How could you implement a separator if the variables will be directly released after their creation?

>Edit: I meant for an algorithm that uses the knowledge of the problem structure, e.g., where you need to keep track of some important variables
  
The SCIP is needed because SCIP knows the memory where the object was allocated. And yes, one could risk a segfault because SCIP could get free'd first... so now I am unsure about the `__dealloc__` solution... @mattmilten

fserra:
So what are we going to do about this?

It doesn't really make sense that variables don't know the scip pointer/model, since actually, you can only create a variable in a model:
```
scip = Model()
x = scip.addVar(...)
```
Also, it seems one can give callbacks to weakreferences so that they are called when the wearrefereed object is freed... this way we can ensure that variables get freed when freeing the main scip...

mattmilten:
Yes, it seems to make a lot of sense to give the Constraint and Variable classes a scip pointer. What could be the downside?

PR #1190, opened 2026-02-19
Remove incorrect Py_INCREF/DECREF on Model in catchEvent/dropEvent

Joao-Dionisio:


DominikKamp:
And what prevents event handlers from calling `dropEvent()`?

Joao-Dionisio:
@DominikKamp Nothing, but I don't think anything bad will happen if they do.

If the question is why wouldn't a user just do it, I don't think we should expect Python users to worry about these things.

tests/test_event.py
```diff
@@ -189,4 +193,37 @@ def eventexec(self, event):
...
+    self.model is already dead), this caused the Model to leak — its refcount
+    never returned to zero, preventing garbage collection.
+    """
+
+    class SimpleEvent(Eventhdlr):
+        def eventinit(self):
+            self.model.catchEvent(SCIP_EVENTTYPE.NODEFOCUSED, self)
+
+        def eventexit(self):
+            pass  # intentionally no dropEvent, as is common in practice
```
DominikKamp:
```suggestion
            pass  # intentionally no dropEvent, which is bad practice
```

tests/test_event.py
```diff
@@ -189,4 +193,37 @@ def eventexec(self, event):
...
+    ev = SimpleEvent()
+    m.includeEventhdlr(ev, "simple", "test event handler")
+    m.addVar("x", obj=1, vtype="I")
+    m.optimize()
+
+    ref = weakref.ref(m)
+    del ev
+    del m
+    gc.collect()
+
```
DominikKamp, line 227:
```suggestion

    del m
    gc.collect()
    assert ref() is not None, "Model was accidentally garbage collected — event handler active"

    del ev
    gc.collect()
```
To ensure the test is working.

tests/test_event.py
```diff
@@ -1,3 +1,7 @@
+import gc
+import sys
+import weakref
+
 import pytest, random
```
DominikKamp:
```suggestion
import pytest, weakref, gc, random
```

tests/test_event.py
```diff
@@ -189,4 +189,43 @@ def eventexec(self, event):
...
+    del ev
+    del m
+    gc.collect()
+
+    del m
+    gc.collect()
+    assert ref() is not None, "Model was accidentally garbage collected — event handler active"
+
+    del ev
+    gc.collect()
```
DominikKamp:
```suggestion

    del m
    gc.collect()
    assert ref() is not None, "Model was accidentally garbage collected — event handler active"

    del ev
    gc.collect()
```
GitHub messed up the suggestion.

Joao-Dionisio: Wait, why would you want the event handler to remain active after deleting the model?

Joao-Dionisio: Also, we're using weakref proxies, so event handlers aren't directly tying Models down.

I have a branch somewhere that reverts the usage of these weakrefs, but never got around to creating a PR.

DominikKamp: The event handler references the model, so it should not be garbage collected before the event handler is deleted. If del kills the model, why do we expect reference counters to prevent this?

Joao-Dionisio: Hang on, the weak ref branch fixes this more cleanly, I'll create another PR in a bit

Joao-Dionisio: ok not more clean, as users would need to explicitly free their models.
But you can take a look at the branch remove-weakref-proxy for this alternative.

DominikKamp: Sounds very clean. But I just do not understand the test. Did it fail before the fix?

Joao-Dionisio: Regarding your point, the event handler does not reference the model. It only has a weakly-referenced proxy. This was first discussed in #26

DominikKamp: > It only has a weakly-referenced proxy.

And why does this make sense or is required?

Joao-Dionisio: Felipe was saying this:

image


I understand that removing the weakref alternative is attractive for someone coming from the C world, but from a Python perspective, it's not intuitive that you start having to worry about memory.

tests/test_event.py
```diff
@@ -189,4 +189,40 @@ def eventexec(self, event):
...
+    m.addVar("x", obj=1, vtype="I")
+    m.optimize()
+
+    ref = weakref.ref(m)
+
+    del m
+    gc.collect()
+    assert ref() is not None, "Model was accidentally garbage collected — event handler active"
+
+    del ev
```
DominikKamp:
```suggestion
    del ev
    gc.collect()
    assert ref() is not None, "Model was garbage collected — event handler absorbed a reference"

    del m
```

tests/test_event.py
DominikKamp: You probably mean that we can not avoid the weak referencing because we did not stop worrying about memory. So the model imposes a manual reference on the eventhandler and the eventhandler gets a weak reference of the model. Deleting the eventhandler will not take any effect on the model and deleting the model will release the eventhandler, which however is not triggered due to this cyclic referencing of catch and drop, but this is what Copilot should have said.

tests/test_event.py
```diff
@@ -189,4 +189,40 @@ def eventexec(self, event):
...
\ No newline at end of file
+        m.optimize()
+
+def test_catchEvent_does_not_leak_model():
+    """catchEvent should not artificially increment the Model's reference count.
+
+    Previously, catchEvent called Py_INCREF(self) on the Model, and dropEvent
+    called Py_DECREF(self). Since many event handlers skip dropEvent (e.g. when
+    self.model is already dead), this caused the Model to leak — its refcount
+    never returned to zero, preventing garbage collection.
```
DominikKamp: This sentence feels quite incomplete, but well.

DominikKamp:
A changelog entry is necessary.

CHANGELOG.md
```diff
@@ -2,7 +2,9 @@
 
 ## Unreleased
 ### Added
+- Added `getMemUsed()`, `getMemTotal()`, and `getMemExternEstim()` methods
```
DominikKamp, line 5: Probably that we do not have to worry about memory anymore because we know how to measure it.

CHANGELOG.md
```diff
@@ -2,7 +2,9 @@
...
 ### Fixed
+- Removed incorrect `Py_INCREF`/`Py_DECREF` on `Model` in `catchEvent`/`dropEvent` that caused reference count imbalance
```
```suggestion
- Removed `Py_INCREF`/`Py_DECREF` on `Model` in `catchEvent`/`dropEvent` that caused memory leak for imbalanced usage
```

tests/test_event.py
Joao-Dionisio: Changed.

PR #1194, opened 2026-03-03
Replace weakref.proxy with strong references for plugin self.model (#…

### LINKED ISSUE #1193: Potential conshdlr conslock life cycle problem
**Describe the bug**
Cleaning up the SCIP model results in a ReferenceError in conslock of a constraint handler.

**To Reproduce**
Construct a single file with the following contents and execute it:
```
import pyscipopt as ps


class DummyConshdlr(ps.Conshdlr):

    def createCons(self, vars, model):
        cons = self.model.createCons(self, "dummy")
        # cons.data = {"vars": vars, "model": model} # alt 1
        cons.data = {"vars": vars} # alt 2
        return cons

    def conscheck(self, constraints, solution, check_integrality, check_lp_rows, print_reason, completely, **results):
        return {"result": ps.SCIP_RESULT.FEASIBLE}

    def consenfolp(self, constraints, n_useful_conss, sol_infeasible):
        return {"result": ps.SCIP_RESULT.FEASIBLE}

    def conslock(self, constraint, locktype, nlockspos, nlocksneg):
        model = self.model
        for var in constraint.data["vars"]:
            model.addVarLocks(var, nlockspos + nlocksneg, nlockspos + nlocksneg)


def test():

    model = ps.Model()
    vars = [ model.addVar() ]

    conshdlr = DummyConshdlr()
    model.includeConshdlr(conshdlr, "dummy", "dummy")
    model.addPyCons(conshdlr.createCons(vars, model))

    model.optimize()
    

test()
```

This then first successfully solves the problem, but also results in an unexpected ReferenceError. See the following log:

```
feasible solution found by trivial heuristic after 0.0 seconds, objective value 0.000000e+00
presolving:
presolving (1 rounds: 1 fast, 0 medium, 0 exhaustive):
 0 deleted vars, 0 deleted constraints, 0 added constraints, 0 tightened bounds, 0 added holes, 0 changed sides, 0 changed coefficients
 0 implications, 0 cliques, 0 implied integral variables (0 bin, 0 int, 0 cont)
transformed 3/3 original solutions to the transformed problem space
presolving solved problem
Presolving Time: 0.00

SCIP Status        : problem is solved [optimal solution found]
Solving Time (sec) : 0.00
Solving Nodes      : 0
Primal Bound       : +0.00000000000000e+00 (3 solutions)
Dual Bound         : +0.00000000000000e+00
Gap                : 0.00 %
Traceback (most recent call last):
  File "XXXXXXXXX", line 21, in conslock
    model.addVarLocks(var, nlockspos + nlocksneg, nlockspos + nlocksneg)
    ^^^^^^^^^^^^^^^^^
ReferenceError: weakly-referenced object no longer exists
Exception ignored in: 'pyscipopt.scip.PyConsLock'
Traceback (most recent call last):
  File "XXXXXXXXX.py", line 21, in conslock
    model.addVarLocks(var, nlockspos + nlocksneg, nlockspos + nlocksneg)
ReferenceError: weakly-referenced object no longer exists
[cons.c:7592] ERROR: Error <0> in function call
[prob.c:1602] ERROR: Error <0> in function call
[prob.c:458] ERROR: Error <0> in function call
[scip_solve.c:2033] ERROR: Error <0> in function call
[scip_solve.c:3511] ERROR: Error <0> in function call
[scip_prob.c:847] ERROR: Error <0> in function call
[scip_general.c:412] ERROR: Error <0> in function call
Traceback (most recent call last):
  File "src/pyscipopt/scip.pxi", line 319, in pyscipopt.scip.PY_SCIP_CALL
Exception: SCIP: unspecified error!
Exception ignored in: 'pyscipopt.scip.Model.__dealloc__'
Traceback (most recent call last):
  File "src/pyscipopt/scip.pxi", line 319, in pyscipopt.scip.PY_SCIP_CALL
Exception: SCIP: unspecified error!
```

**Expected behavior**
I expect the code to run without such error and produce a log identical to the situation where _alt 1_ is uncommented and _alt 2_ is commented, which yields the same log, except for the part after _Gap:_, which is not present with _alt 1_ uncommented. Additionally, I do not expect a different result from the code depending on whether _alt 1_ or _alt 2_ is commented.

**System**
 - OS: Windows
 - Version 11 Enterprise
 - SCIP version 10.0.0
 - Installed pyscipopt version 6.0.0 through pip

Joao-Dionisio:
Fix #1193

Picking up something I had been playing with. We keep the references to the plugins, and allow Python to garbage collect them itself. This is the proper fix to the problem fixed by #1190.  I was just hesitant to use it because, at the time, this would require users to explicitly call `dropEvent()`. This is no longer the case.

Can you please take a look @DominikKamp ?


DominikKamp:
And when is `__del__()` called?

Joao-Dionisio:
> And when is `__del__()` called?

`__del__()` is one of those Python keywords, and is called just before the object is deleted, I believe.

DominikKamp:
> and is called just before the object is deleted

Triggered by the garbage collector, or who deletes this?

Joao-Dionisio:
> Triggered by the garbage collector, or who deletes this?

The garbage collector at the end, or by the garbage collector when you explicitly call `del`.

DominikKamp:
So this makes sure that the model is not freed before the plugins, which makes sense, but still someone has to trigger the cycle breaking, or what was the memory reference issue that justified using weak references?

Joao-Dionisio:
> someone has to trigger the cycle breaking, or what was the memory reference issue that justified using weak references?

@DominikKamp A Claude + Google search led me to [PEP 442](https://peps.python.org/pep-0442/) that introduced proper cleanup with `__del__`. It doesn't try to break the cycle, but cleans up things in an orderly manner. I'm not sure why this option wasn't chosen originally, though, I think it was available.

Joao-Dionisio:
@DominikKamp while the cluster is down, can you please take another look at this? I think this fixes what's blocking a new release.

src/pyscipopt/event.pxi
```diff
@@ -3,6 +3,10 @@
 cdef class Eventhdlr:
     cdef public Model model
     cdef public str name
+    cdef public list _caught_events
```
DominikKamp, line 6: Why is this required now? The previous claim was that everyone can ignore it and events are magically dropped. Why now maintain a list for what the eventhandlers should do?

tests/test_event.py
```diff
@@ -200,6 +200,9 @@ def eventexec(self, event):
     with pytest.raises(Exception):
         m.optimize()
 
+    del m, ev, v
+    gc.collect()
```
DominikKamp, line 204: What is this for?

src/pyscipopt/scip.pxd
```diff
@@ -2251,7 +2251,7 @@ cdef class Model:
     cdef int _generated_event_handlers_count
     # store references to Benders subproblem Models for proper cleanup
     cdef _benders_subproblems
-    # store references to plugins to break circular references in __dealloc__
+    # store references to plugins for the Model <-> Plugin reference cycle
```
DominikKamp, line 2254: So how does this work now?

src/pyscipopt/scip.pxi
```diff
@@ -2918,20 +2933,17 @@ cdef class Model:
...
             if self._modelvars:
                 for var in self._modelvars.values():
                     (<Variable>var).scip_var = NULL
-                self._modelvars = {}
             if self._modelconss:
                 for cons in self._modelconss.values():
                     (<Constraint>cons).scip_cons = NULL
-                self._modelconss = {}
-
-            PY_SCIP_CALL( SCIPfree(&self._scip) )
```
DominikKamp, line 2934: So why do we still need to clear the pointers if it does no longer matter when to do it?

CHANGELOG.md
```diff
@@ -6,7 +6,8 @@
 - Added `getMemUsed()`, `getMemTotal()`, and `getMemExternEstim()` methods
 ### Fixed
 - Removed `Py_INCREF`/`Py_DECREF` on `Model` in `catchEvent`/`dropEvent` that caused memory leak for imbalanced usage
-- Used `getIndex()` instead of `ptr()` for sorting nonlinear expression terms to avoid nondeterministic behavior 
+- Used `getIndex()` instead of `ptr()` for sorting nonlinear expression terms to avoid nondeterministic behavior
+- Replaced `weakref.proxy` with strong references for plugin `self.model`, fixing `ReferenceError` during cleanup callbacks (#1193)
```
DominikKamp, line 10: Should this not be "to the plugins from `self.model`"?

DominikKamp:
Fine if it works, but a comprehensive description of what the current approach is now finally with a fixed policy how plugin types need to be designed is in need.

src/pyscipopt/event.pxi
Joao-Dionisio: No need to be public, you're right. But the magic is not as magic as that, we're calling `dropEvent()` ourselves, so we must iterate through these events.

tests/test_event.py
Joao-Dionisio: Old stuff

src/pyscipopt/scip.pxd
Joao-Dionisio: Added comment explaining

src/pyscipopt/scip.pxi
Joao-Dionisio: This is for the case where users are holding PySCIPOpt variables after the corresponding SCIP variables were freed.

CHANGELOG.md
Joao-Dionisio: changed

Joao-Dionisio:
Hopefully it's clearer now. Will merge, assuming everything passes.

Joao-Dionisio:
Ok, all good. Will merge. Hopefuly, no one will come back to this PR in 5 years time to understand what the heck people were thinking... 👁️ 👁️

PR #1196, opened 2026-03-06
Avoid random order dependence of nonlinear expressions

pfetsch:
Parsing the terms in a nonlinear expression sorts the terms according to the pointer of the SCIP variable. This causes random effects, since the pointer varies from run to run. In particular, the solving times vary significantly from instance to instance.

This MR uses `getIndex()` instead of `ptr()`. This seems to fix the issue.

However, I am not sure whether there are other cases in which the order depends on such effects.

### DIFF
--- a/src/pyscipopt/expr.pxi
+++ b/src/pyscipopt/expr.pxi
@@ -108,7 +108,7 @@ cdef class Term:
     cdef Py_ssize_t hashval
 
     def __init__(self, *vartuple: Variable):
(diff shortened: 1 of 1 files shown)


Joao-Dionisio:
Thanks, Marc, makes perfect sense!
I've looked around at the code, and I'd say the other uses of `ptr` wouldn't impact determinism, so I'll just merge this.

pfetsch:
Thanks for looking at it and merging. 
The only open question is whether this should be noted somewhere, e.g., a changlog.

PR #1210, opened 2026-04-22
Add addMatrixConsDisjunction for matrix constraints

### LINKED ISSUE #1084: Cannot create disjunction of matrix constraints
I'm not sure if this is a documentation bug (i.e. it might just not be possible), or a bug in pyscipopt, but trying to add a disjunction constraint of two matrix constraints reports 

```
AttributeError: 'MatrixConstraint' object has no attribute '_lhs'
```

I guess even if it is not intended to be supported, letting it fall through to a Python error is not ideal.
 

**To Reproduce**

Run python on the following script
```
from pyscipopt import Model, quicksum
import numpy as np
scip = Model()

n=5
shape=(n,1)
x = scip.addMatrixVar(shape, vtype='C', name='x', ub=8)

A1 = np.random.rand(2,5)
A2 = np.random.rand(2,5)

z2 = np.c_[[0,0]]

c1=scip.addMatrixCons(A1 @ x == z2, enforce=False)
c2=scip.addMatrixCons(A2 @ x == z2, enforce=False)

scip.addConsDisjunction([c1,c2])
```

**Expected behavior**

Disjunctive Constraint added to model.
**System**
 - OS: linux
 - Version 5.6.0
 - SCIP version 9.2.1
 - How did you install `pyscipopt`? from A debian package that I made.

Joao-Dionisio:
Fixes #1084.

@Zeroto521 you're our matrix guy, would you mind taking a look? :)


src/pyscipopt/scip.pxi
```diff
@@ -6641,6 +6641,122 @@ cdef class Model:
...
+            Is the constraint subject to aging? (Default value = False)
+
+        Returns
+        -------
+        MatrixConstraint or Constraint
+            A ``MatrixConstraint`` of the common shape when any entry of
+            ``conss`` is a ``MatrixExprCons``; otherwise a single
+            ``Constraint`` from the scalar fallback.
+        """
+        assert isinstance(conss, Iterable), "Given constraint list is not iterable"
```
Zeroto521: It would be better to use `raise` instead of `assert` in the main code.

```suggestion
        if not isinstance(conss, Iterable):
            raise TypeError(f"given constraint list is not iterable, but got {type(conss).__name__}")
```

src/pyscipopt/scip.pxi
```diff
@@ -6641,6 +6641,122 @@ cdef class Model:
...
+                elem_conss,
+                name=matrix_names[idx],
+                initial=matrix_initial[idx],
+                relaxcons=relaxcons,
+                enforce=matrix_enforce[idx],
+                check=matrix_check[idx],
+                local=matrix_local[idx],
+                modifiable=matrix_modifiable[idx],
+                dynamic=matrix_dynamic[idx],
+            )
```
Zeroto521, line 6759: It's too slow to use Python loops to build matrices when using `addMatrixCons` and `addMatrixConsIndicator` for large variables. I'm planning to port this to Cython later for a speed boost, but it works fine as-is for now.

Zeroto521 (commented):
I haven't used the `addConsDisjunction` API before, so I cannot provide much useful advice here.

PR #1214, opened 2026-05-05
release SOS1/SOS2 constraints after addition

pfetsch:
For all other constraint types, the constraint is released after addition.

### DIFF
--- a/src/pyscipopt/scip.pxi
+++ b/src/pyscipopt/scip.pxi
@@ -7151,6 +7151,7 @@ cdef class Model:
                 PY_SCIP_CALL(SCIPaddVarSOS1(self._scip, scip_cons, wrapper.ptr[i], weights[i]))
 
         PY_SCIP_CALL(SCIPaddCons(self._scip, scip_cons))
+        PY_SCIP_CALL(SCIPreleaseCons(self._scip, &scip_cons))
 
         return self._getOrCreateCons(scip_cons)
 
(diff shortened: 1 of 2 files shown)


src/pyscipopt/scip.pxi
```diff
@@ -7151,6 +7151,7 @@ cdef class Model:
                 PY_SCIP_CALL(SCIPaddVarSOS1(self._scip, scip_cons, wrapper.ptr[i], weights[i]))
 
         PY_SCIP_CALL(SCIPaddCons(self._scip, scip_cons))
+        PY_SCIP_CALL(SCIPreleaseCons(self._scip, &scip_cons))
 
         return self._getOrCreateCons(scip_cons)
```
Joao-Dionisio: I think you need to `getOrCreateCons()` before releasing the constraint, as in the other methods. Example from knapsack constraints:

```python
        PY_SCIP_CALL(SCIPaddCons(self._scip, scip_cons))

        pyCons = self._getOrCreateCons(scip_cons)
        PY_SCIP_CALL(SCIPreleaseCons(self._scip, &scip_cons))

        return pyCons
```

pfetsch: Yes, indeed. Thanks for finding this. Should be fixed now.

Joao-Dionisio:
Can you please add a changelog entry under 6.2.0? In the middle of a release, will focus on it in the coming days.

pfetsch:
> Can you please add a changelog entry under 6.2.0? In the middle of a release, will focus on it in the coming days.

Good point. I have added this.

Joao-Dionisio (approved):
Thank you @pfetsch !

PR #1220, opened 2026-05-14
Annotate several Model methods in the stubs

jonathanberthias:
This is a first batch of annotations and default parameter values. I believe these are rather straightforward.

I ignored some annotations on purpose: anything related to `ExprCons` because the annotations for `Expr` operations aren't correct yet, and anything related to the `PY_SCIP_*` enums. These can be revisited at a later stage.

Part of #1072


Joao-Dionisio:
Awesome @jonathanberthias ! Can you also adapt the stub checker for this, to force new contributions to annotate as well?

jonathanberthias:
> Awesome @jonathanberthias ! Can you also adapt the stub checker for this, to force new contributions to annotate as well?

I'm not sure that's possible yet, I only annotated a tiny part of the API so we would need to "finish" everything before we can forbid broad annotations like ˋIncompleteˋ. 

BTW the automated script to generate stubs can't really be used anymore since it would remove all of these new annotations. We either need to make it smarter and not remove existing annotations, or rely on manual work for now.

Joao-Dionisio:
>  so we would need to "finish" everything before we can forbid broad annotations like ˋIncompleteˋ.

That is true, but perhaps we could have an intermediate script that just checks that no new stub errors are introduced. But maybe it's not worth the effort, yes.

On the automated script: yeah, I should do another pass on it someday. Not this weekend, but perhaps the next.

jonathanberthias:
@Joao-Dionisio I think this is ready to merge in the current state, I prefer to tackle other topics in other PRs to avoid having too many merge conflicts. Let me know if you see anything wrong!

Joao-Dionisio:
@jonathanberthias alright, thank you very much! This week might be a bit busy, but I'll try to take a look over the weekend.

Joao-Dionisio:
Hey @jonathanberthias ! My claude flagged the following:

  Found 5 issues:
  
  1. getLhs is annotated -> None but returns the constraint's left-hand side value (a float). Should be -> float.

  https://github.com/scipopt/PySCIPOpt/blob/48148228a95840ceff0f7d0fe692d530ba80d801/src/pyscipopt/scip.pyi#L1127-L1129

  2. addConsLocal takes cons: Constraint and returns -> None, but since #1151 the implementation declares ExprCons cons and ends with return pycons (a Constraint). Should be cons: ExprCons and -> Constraint.
  (ExprCons and Constraint are unrelated classes, so passing an expression like x + y <= 5 would be wrongly flagged.)

  https://github.com/scipopt/PySCIPOpt/blob/48148228a95840ceff0f7d0fe692d530ba80d801/src/pyscipopt/scip.pyi#L694-L708

  3. Same issue in addConsNode: cons: Constraint should be ExprCons, and -> None should be -> Constraint.

  https://github.com/scipopt/PySCIPOpt/blob/48148228a95840ceff0f7d0fe692d530ba80d801/src/pyscipopt/scip.pyi#L709-L724

  4. setBoolParam types value: float, but it sets a boolean parameter (docstring says bool, and SCIPsetBoolParam takes a bool). Should be value: bool — siblings setIntParam/setRealParam are correctly typed.

  https://github.com/scipopt/PySCIPOpt/blob/48148228a95840ceff0f7d0fe692d530ba80d801/src/pyscipopt/scip.pyi#L1492-L1494

  5. calcNodeselPriority is annotated -> int, but SCIPcalcNodeselPriority returns SCIP_Real. Should be -> float. Relatedly, createChild's nodeselprio (L967) is typed int but SCIPcreateChild takes SCIP_Real 
  nodeselprio, so it should be float too.

  https://github.com/scipopt/PySCIPOpt/blob/48148228a95840ceff0f7d0fe692d530ba80d801/src/pyscipopt/scip.pyi#L913-L916

jonathanberthias:
Thanks a lot for the review @Joao-Dionisio! I've fixed the findings, indeed I didn't integrate recent changes to the repo correctly, my bad :sweat_smile:

Joao-Dionisio:
Okay, I think this should be the last of them:

  **Over-permissive parameters (`None` would crash at runtime):**

  1. `addVar` — `obj: float | None`, but only `lb`/`ub` are None-handled; `obj=None` would be passed straight to `SCIPcreateVarBasic`. Should be `obj: float`.

  https://github.com/scipopt/PySCIPOpt/blob/0486e4f4d41bd43bbf9ebcdc465ff9ec052915d8/src/pyscipopt/scip.pyi#L877-L887

  2. `addMatrixVar` — `obj` and `pricedVarScore` are typed `... | None`, but the implementation is `Union[int, float, np.ndarray]` with no None guard (`np.full(..., dtype=float)` crashes on `None`). Drop the
  `| None`.

  https://github.com/scipopt/PySCIPOpt/blob/0486e4f4d41bd43bbf9ebcdc465ff9ec052915d8/src/pyscipopt/scip.pyi#L860-L870

  3. `getRowParallelism` — `orthofunc: Literal["d", "e", 100, 101]`, but the value is passed to a C `char` via integer conversion, so a string raises `TypeError`; only the ASCII codes work. Should be
  `Literal[100, 101] = 101`.

  https://github.com/scipopt/PySCIPOpt/blob/0486e4f4d41bd43bbf9ebcdc465ff9ec052915d8/src/pyscipopt/scip.pyi#L1216-L1218

  **Return type missing `| None` (the implementation has a `return None` / NULL path):**

  4. `getConsVals` and `getConsVars` return `None` when the constraint can't be expressed as a single linear constraint (`if not success: return None`). Should be `list[float] | None` and `list[Variable] |
  None`.

  https://github.com/scipopt/PySCIPOpt/blob/0486e4f4d41bd43bbf9ebcdc465ff9ec052915d8/src/pyscipopt/scip.pyi#L1107-L1108

  5. `getCurrentNode` — `Node.create(SCIPgetCurrentNode(...))` returns `None` when not in the solving stage (sibling getters like `getBestChild` are already `Node | None`). Should be `Node | None`.

  https://github.com/scipopt/PySCIPOpt/blob/0486e4f4d41bd43bbf9ebcdc465ff9ec052915d8/src/pyscipopt/scip.pyi#L1110

  **Parameter missing `| None`:**

  6. `getSolVal` (both overloads) — `sol: Solution`, but the body is `(sol or Solution.create(...))[expr]` and the docstring says "If None, the current LP/pseudo solution is used" (the sibling `getSolObjVal`
  is already `Solution | None`). Should be `sol: Solution | None`.

  https://github.com/scipopt/PySCIPOpt/blob/0486e4f4d41bd43bbf9ebcdc465ff9ec052915d8/src/pyscipopt/scip.pyi#L1229-L1232

  **Wrong numeric return type (`SCIP_Longint` maps to Python `int`, not `float`):**

  7. `getCapacityKnapsack` returns `SCIP_Longint`. Should be `-> int`.

  https://github.com/scipopt/PySCIPOpt/blob/0486e4f4d41bd43bbf9ebcdc465ff9ec052915d8/src/pyscipopt/scip.pyi#L1102

  8. `getWeightsKnapsack` values come from `SCIP_Longint*`. Should be `-> dict[str, int]`.

  https://github.com/scipopt/PySCIPOpt/blob/0486e4f4d41bd43bbf9ebcdc465ff9ec052915d8/src/pyscipopt/scip.pyi#L1294

  9. `getParam` / `getParams` — the unions omit `int`, but `SCIP_PARAMTYPE_INT`/`LONGINT` params return Python `int`. Add `int` to both.

  https://github.com/scipopt/PySCIPOpt/blob/0486e4f4d41bd43bbf9ebcdc465ff9ec052915d8/src/pyscipopt/scip.pyi#L1195-L1196

  **Return class too narrow:**
  
  10. `addMatrixCons` is `-> MatrixConstraint`, but the `isinstance(cons, ExprCons)` fast-path returns `self.addCons(...)`, a plain `Constraint` (docstring: "Constraint or MatrixConstraint"). Should be
  `MatrixConstraint | Constraint`.

  https://github.com/scipopt/PySCIPOpt/blob/0486e4f4d41bd43bbf9ebcdc465ff9ec052915d8/src/pyscipopt/scip.pyi#L817-L831

  11. `addMatrixConsIndicator` — same pattern; the `ExprCons` path returns `self.addConsIndicator(...)` → `Constraint`. Should be `MatrixConstraint | Constraint`.

  https://github.com/scipopt/PySCIPOpt/blob/0486e4f4d41bd43bbf9ebcdc465ff9ec052915d8/src/pyscipopt/scip.pyi#L844-L859

  **Spurious `| None` in return:**
  
  12. `solveBendersSubproblem` is `-> tuple[bool, float | None]`, but `objective` is a `cdef SCIP_Real` always populated by the C call — never `None`. Should be `tuple[bool, float]`.

  https://github.com/scipopt/PySCIPOpt/blob/0486e4f4d41bd43bbf9ebcdc465ff9ec052915d8/src/pyscipopt/scip.pyi#L1544-L1550

jonathanberthias:
Hi @Joao-Dionisio, I forgot to ping you but I've addressed the previous findings. I tried to scan a bit more myself but there a way too many functions to be exhaustive manually :sweat_smile: 
Don't hesitate to push to the branch if you find more issues!

Joao-Dionisio:
Oops, sorry @jonathanberthias , I missed your update! But yeah, things are looking good now, thank you for your work!!

Joao-Dionisio (approved):

PR #1221, opened 2026-05-20
Wrap `captureCons` and `releaseCons`

Joao-Dionisio:


src/pyscipopt/scip.pxi
```diff
@@ -8568,7 +8568,31 @@ cdef class Model:
...
+        PY_SCIP_CALL(SCIPcaptureCons(self._scip, cons.scip_cons))
+
+    def releaseCons(self, Constraint cons):
+        """
+        Decrease the usage counter of a constraint; frees it when the counter reaches zero.
+
+        Parameters
+        ----------
+        cons : Constraint
+            constraint to release
```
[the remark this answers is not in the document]
DominikKamp: Well, `frees it when the counter reaches zero` is quite explicit, I rather wonder whether this manual handling complies with the automatic Python referencing.

src/pyscipopt/scip.pxi
```diff
@@ -8568,7 +8568,31 @@ cdef class Model:
...
+
+        """
+        PY_SCIP_CALL(SCIPreleaseCons(self._scip, &cons.scip_cons))
```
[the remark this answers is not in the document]
DominikKamp: Should be done since the same Python constraint object can be used multiple times and the cleanup is handled by Python.

src/pyscipopt/scip.pxi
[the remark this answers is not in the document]
Joao-Dionisio: I believe it does not, nice catch. What should be the fix? Would a docstring warning be enough?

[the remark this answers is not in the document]
DominikKamp: Alternatively, throw an exception if anything would be released to zero by checking the uses in advance, the rule would be to not release more than captured.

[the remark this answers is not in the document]
Joao-Dionisio: Added. The only concern is with people wanting to release a constraint created via `createConsFromExpr()`, but they shouldn't be using this method in any case.

DominikKamp:
Could this also be done for variables to avoid asymmetry?

Joao-Dionisio:
@DominikKamp done

src/pyscipopt/scip.pxi
```diff
@@ -4636,6 +4636,62 @@ cdef class Model:
...
+        """
+        PY_SCIP_CALL(SCIPcaptureVar(self._scip, var.scip_var))
+
+    def releaseVar(self, Variable var):
+        """
+        Decrease the usage counter of a variable.
+
+        Unlike the underlying ``SCIPreleaseVar``, this wrapper refuses to
+        release the last reference: it must be paired with a prior
+        captureVar call. This guarantees the variable is never freed via
```
```suggestion
        release the last reference. It must be paired with a prior
        captureVar call. This guarantees the variable is never freed via
```

src/pyscipopt/scip.pxi
```diff
@@ -8568,7 +8624,63 @@ cdef class Model:
...
+        """
+        PY_SCIP_CALL(SCIPcaptureCons(self._scip, cons.scip_cons))
+
+    def releaseCons(self, Constraint cons):
+        """
+        Decrease the usage counter of a constraint.
+
+        Unlike the underlying ``SCIPreleaseCons``, this wrapper refuses to
+        release the last reference: it must be paired with a prior
+        captureCons call. This guarantees the constraint is never freed via
```
```suggestion
        release the last reference. It must be paired with a prior
        captureCons call. This guarantees the constraint is never freed via
```

src/pyscipopt/scip.pxd
```diff
@@ -820,6 +820,7 @@ cdef extern from "scip/scip.h":
     SCIP_RETCODE SCIPcaptureVar(SCIP* scip, SCIP_VAR* var)
     SCIP_RETCODE SCIPaddPricedVar(SCIP* scip, SCIP_VAR* var, SCIP_Real score)
```
DominikKamp: Why sandwich this?

src/pyscipopt/scip.pxi
```diff
@@ -4636,6 +4636,62 @@ cdef class Model:
...
+        Get the number of times the variable is currently captured.
+
+        Parameters
+        ----------
+        var : Variable
+            variable to query
+
+        Returns
+        -------
+        int
```
DominikKamp, line 4690: This seems incomplete.

src/pyscipopt/scip.pxi
```diff
@@ -8568,7 +8624,63 @@ cdef class Model:
...
+        Get the number of times the constraint is currently captured.
+
+        Parameters
+        ----------
+        cons : Constraint
+            constraint to query
+
+        Returns
+        -------
+        int
```
DominikKamp, line 8680: This seems incomplete.

src/pyscipopt/scip.pxi
```diff
@@ -4636,6 +4636,62 @@ cdef class Model:
         var.scip_var = NULL
         return deleted
 
+    def captureVar(self, Variable var):
+        """
+        Increase the usage counter of a variable. Must be matched by a releaseVar.
```
```suggestion
        Increase the usage counter of a variable. Must be paired with a later releaseVar.
```

src/pyscipopt/scip.pxi
```diff
@@ -8568,7 +8624,63 @@ cdef class Model:
 
         """
         PY_SCIP_CALL(SCIPdelConsLocal(self._scip, cons.scip_cons))
-    
+
+    def captureCons(self, Constraint cons):
+        """
+        Increase the usage counter of a constraint. Must be matched by a releaseCons.
```
```suggestion
        Increase the usage counter of a constraint. Must be paired with a later releaseCons.
```

tests/test_cons.py
```diff
@@ -368,3 +368,27 @@ def test_getValsLinear():
...
+    # Model continues to function — problem's capture survived.
+    m.optimize()
+    assert m.getStatus() == "optimal"
+
+    with pytest.raises(TypeError):
+        m.captureCons("not a constraint")
+    with pytest.raises(TypeError):
+        m.releaseCons("not a constraint")
+    with pytest.raises(TypeError):
+        m.getConsNUses("not a constraint")
```
DominikKamp: What is this for?

tests/test_vars.py
```diff
@@ -223,4 +225,26 @@ def test_adjustedVarUb():
...
+
+    m.optimize()
+    assert m.getStatus() == "optimal"
+
+    with pytest.raises(TypeError):
+        m.captureVar("not a variable")
+    with pytest.raises(TypeError):
+        m.releaseVar("not a variable")
+    with pytest.raises(TypeError):
+        m.getVarNUses("not a variable")
```
DominikKamp: What is this for?

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 ### Added
+- Added `Model.captureCons()`, `Model.releaseCons()`, `Model.getConsNUses()`, `Model.captureVar()`, `Model.releaseVar()`, and `Model.getVarNUses()`
```
```suggestion
- Added `Model.captureVar()`, `Model.releaseVar()`, `Model.getVarNUses()`, `Model.captureCons()`, `Model.releaseCons()`, and `Model.getConsNUses()`
```

DominikKamp (commented):
I do not understand, what the changes in scip.pyi have to do with captures, otherwise this looks okay.

src/pyscipopt/scip.pxd
Joao-Dionisio: I like sandwiches, they can be pretty tasty

DominikKamp: Depends on what is inside the sandwich, this one looks more expensive than tasty.

Joao-Dionisio: Actually, and I'm not even kidding, the best sandwich I've ever had was on a Subway in Montreal, when I was there for ISMP. So it's thanks to you, really!

mmghannam:
What is the use case that this enables? are you sure you want to expose memory handling? misuse of these can expose memory unsafety and memory leaks.

DominikKamp (approved):
This looks okay, but maybe wait until the usecase is reliable. So far this only introduces user unsafety. The memory should be fine since proper programs do not leave captures unreleased, same for events.

Joao-Dionisio:
@mmghannam the usecase is [scip#201](https://github.com/scipopt/scip/issues/201), but apparently it's still being discussed.

@mmghannam @DominikKamp , in the in-person meeting Marc (Pfetsch) wished for an `I_KNOW_WHAT_I_AM_DOING` flag in SCIP, but then what if we add something similar to PySCIPOpt? Like, have a secondary model class that inherits from the main model, but also has these more scary methods. We can even have it in a separate file, so that people really need to do `from pyscipopt.scary_methods import crazyModel`.

Perhaps such a separation would make sense also for more advanced use cases. For example, we've add users thinking they had to use `addConsKnapsack()`, because they didn't know that SCIP upgrades constraints. Perhaps if such methods were only declared in an "advanced user model", these errors would diminish.

Does this make any sense or is it time for me to go to bed?

DominikKamp:
Maybe there could be a base class with elementary functionality, but for these functions a warning label should be sufficient, so better go to bed at first.

Joao-Dionisio:
Closing to clear the open PRs, it seems like this will end up being fixed on the SCIP side.

But I did like the idea of having more than one model class.

Issue #1222, opened 2026-05-27
addConsCumulative and AddConsCumulativeOpt?


tias:
I just learned at CPAIOR that SCIP has a cumulative (and disjunctive/no-overlap?) constraint.

Indeed, I can find 'cons_cumulative.c' on: https://www.scipopt.org/doc/html/group__DEFPLUGINS__CONS.php
and I can even find a 'cons_optcumulative.c' on: https://www.scipopt.org/doc/html/group__CONSHDLRS.php
(the difference between the two doc files escapes me)

I do not immediately see a Disjunctive/NoOverlap constraint (but I am well aware we can use Cumulative for that).

We would love to interface to these constraint handlers from our CPMpy library. Would it be possible to add them to PySCIPOpt? Or is there a reason not to?

We could help out if you think its a good idea to add it, and you have some pointers of where to start.

Kind regards,
Tias

Joao-Dionisio:
Hey @tias , thank for the issue! Always happy when SCIP/PySCIPOpt comes up in the wild 😁 

PySCIPOpt's initial focus was more on modeling (and some plugin work as well), but we've seen people interested in more advanced use cases. SCIP internally upgrades constraints into more specialized constraints (like cumulative constraints), but advanced users still want more control (for example #941 with knapsack constraints).

For this reason, in #977 I'm tracking the constraints that have/have not yet been ported into PySCIPOpt, and indeed cumulative constraints are there, there hasn't been much progress. But the process of adding a new constraint handler is simple, once it's done the first time, it's very mechanical. So, here are the steps:

1 - Edit `scip.pxd` with the SCIP methods you want PySCIPOpt to call. For example, `SCIPcreateConsCumulative()`:
```python
cdef extern from "scip/cons_cumulative.h":
    SCIP_RETCODE SCIPcreateConsCumulative(SCIP* scip,
                                         SCIP_CONS** cons,
                                         etc etc)
```
2 - Edit `scip.pxi` with the interface to these methods. Something like (simplified):
```python
def addConsCumulative(args):
    cdef SCIP_CONS* scip_cons
    ...
    PY_SCIP_CALL(SCIPcreateConsCumulative(self.scip, &scip_cons, args))
    PY_SCIP_CALL(SCIPaddCons(self._scip, scip_cons))
    pyCons = self._getOrCreateCons(scip_cons)
    return pyCons
```
[...]

PR #1223, opened 2026-05-27
addConsCumulative

tias:
Hi Joao,

Thanks for the very clear instructurions. With the help of a well-supervised digital friend I created a test-passing PR.

Does that look OK?

Note that I see tons of changes to 'src/pyscipopt/scip.pyi' but I think that is an artefact of code generation, I did not edit this file.

Even if good, don't merge it yet. We will first connect it to CPMpy so we can run our larger test-suite to make sure everything behaves appropriately.

Forgot to write that this is a follow-up to #1222


Joao-Dionisio:
Thank you for the quick follow-up @tias !

On the `.pyi` file, I think the issue is with our common digital friend :) This year we started adding stubs so that code editors could auto-complete (see #1072), meaning that every function addiction must also come with a 1 or 2 line addition to the `pyi` file describing the input/output (forgot this step, sorry 😬 ). This is tedious, so I decided to create an automatic stub generator, that unfortunately is still a WIP, as you can see. The way forward is just telling the digital friend to roll back these `pyi` changes, and just add directly the `consCumulative()` ones, not using the stub generating script.

EDIT: But other than this it does look okay!

Joao-Dionisio (approved):
Hey @tias , took the liberty to update your fork, all should be okay now. Thank you for your contribution, let me know if CPMpy needs any more help! :)

PR #1228, opened 2026-06-16
Fix lotsizing example issue

abb-omidi:


Joao-Dionisio:
Thank you, @abb-omidi ! As this is a fix to an existing example, can you please replace the old file, instead of adding a new one? And also, please add a description of what the problem was and also what you did to fix it, so that people in the future can take a look.

Again, thank you so much!

abb-omidi:
@Joao-Dionisio,

I just updated the file to the original one in the repository. Let me explain the issues:
* The problem described in the repository example is a standard lot-sizing problem known as $\text{(LS-U)}$ formulation.
* The problem contains the initial inventory variable that should be equal to the zero in the end of the last period.
* The mentioned $\text{(L,S)}$ cut in the original problem is actually defined on the space of variables $x \in \text{R}^+$ and $y \in \text{B}$, while instead it should be defined on the space of $x \in \text{R}^+$, $y \in \text{B}$, and $s\in \text{R}^+$ variables. 
* Actually, based on the original paper published by Prof. Wolsey, to separate correctly the mentioned cuts from the solution space we need summing over the flow conservation constraints (involving $x$ and $s$ variables) $\forall j \in L$ in the standard $\text{(LS-U)}$ formulation, and adding them to the example mentioned inequality, gives the correct form. (Actually, the second is a lifted form of the first).
* Finally, defining the matrix to shape the all of cuts needed and using this matrix as a separation parameter to make desired results. 

Please, let me know if you need any further information.
All the best

Joao-Dionisio:
Hey @abb-omidi , thank you for the changes! Just wanted to let you know that I'll likely be swamped until the end of August. By then, things should clear a bunch, and I'll do a proper pass through the repo.

Joao-Dionisio:
I'm closing this PR in favor of #1252.

PR #1236, opened 2026-07-19
Interface SCIP methods missing from PySCIPOpt

adamj34:
Wrap SCIPgetNNodesLeft  and append a check to an existing test


Joao-Dionisio:
Thank you @adamj34 , this looks good! How many other methods did you interface? It may be good to add a bunch at a time, to avoid merging too many PRs and spamming our followers.

adamj34:
@Joao-Dionisio I could interface a few more. I wanted to keep it simple for the first PR but I can add the rest in bulk, as you suggested.

I see that there's an error in the CI pipeline:
error: pyscipopt.scip.Model.getNNodesLeft is not present in stub

Should new methods be added there by hand, or it's supposed to be automated in some way (by the .pre-commit-config.yaml hook, possibly)?

Joao-Dionisio:
> I wanted to keep it simple for the first PR

That's fair, we can merge this one and do the bulk one later.

> Should new methods be added there by hand, or it's supposed to be automated in some way (by the .pre-commit-config.yaml hook, possibly)?

We've started working on adding stubs this year, so for now this is still something that needs to be done manually, but hopefuly not for much longer. There is a stub generator, but it's not working yet.

adamj34:
ok, I've added it manually

src/pyscipopt/scip.pxd
```diff
@@ -1487,6 +1487,9 @@ cdef extern from "scip/scip.h":
     SCIP_Longint SCIPgetNNodeLPIterations(SCIP* scip)
     SCIP_Longint SCIPgetNStrongbranchLPIterations(SCIP* scip)
     SCIP_Real SCIPgetPrimalDualIntegral(SCIP* scip)
+    int SCIPgetNRuns(SCIP* scip)
```
Joao-Dionisio: Can you please order this roughly in the order in which they appear in the [SCIP documentation](https://www.scipopt.org/doc-10.0.0/html/group__PublicSolvingStatsMethods.php)?

src/pyscipopt/scip.pxd
```diff
@@ -1487,6 +1487,9 @@ cdef extern from "scip/scip.h":
...
+    int SCIPgetNReoptRuns(SCIP* scip)
+    void SCIPaddNNodes(SCIP* scip, SCIP_Longint nnodes)
```
Joao-Dionisio: Can you let me know what's the use of this method? Is it to let SCIP know that there were more nodes that were hidden from it somehow?

src/pyscipopt/scip.pxd
```diff
@@ -2134,6 +2137,7 @@ cdef extern from "scip/scip_tree.h":
     SCIP_RETCODE SCIPgetNSiblings(SCIP* scip)
     SCIP_RETCODE SCIPgetLeaves(SCIP* scip, SCIP_NODE*** leaves, int* nleaves)
     SCIP_Longint SCIPgetNLeaves(SCIP* scip)
+    SCIP_Longint SCIPgetNNodesLeft(SCIP* scip)
```
Joao-Dionisio:
```suggestion
    int SCIPgetNNodesLeft(SCIP* scip)
```

src/pyscipopt/scip.pxd
Joao-Dionisio: Ah, I'm guessing it's because of sub-SCIPs, etc. In any case, perhaps adding a line to the PySCIPOpt docstring explaining this would be nice.

adamj34: from the functional standpoint, it increments scip->stat->nnodes and scip->stat->ntotalnodes

as far as concrete use cases are concerned, I'm not sure about them but your guess looks reasonable

src/pyscipopt/scip.pxd
adamj34: Oh, sorry for a careless mistake. Fixed in: 7e4cdc7

src/pyscipopt/scip.pxd
adamj34: done in 7e4cdc7

Joao-Dionisio (approved):
Looks good @adamj34 , thank you! If there are other methods you wish to interface, we'd be very appreciative!

adamj34:
Sure, I'll probably find some time this week to interface a few more :)

PR #1237, opened 2026-07-25
Interface SCIP methods

adamj34:
Adds the following methods: 
`getNNodesLeft()`, `getNRuns()`, `getNReoptRuns()`, `addNNodes()`, `getDeterministicTime()`, `getAvgDualbound()`, `getMaxTotalDepth()`, `getNBacktracks()`, `getFocusNode()`, `getAvgLowerbound()`, `getFirstPrimalBound()`, `getLowerboundRoot()`, `getUpperbound()`, `getNObjlimLeaves()`

Adds tests in `tests/test_statistics.py` and `tests/test_node.py` for migrated methods

Adds function signatures to `scip.pyi` for type hints

Updates `CHANGELOG.md`


src/pyscipopt/scip.pyi
adamj34: @Joao-Dionisio were there some changes in the ruff config? I didn't even touch these lines in my commits.

If it's ok I can refactor it to use the more modern X | Y syntax as suggested in the error message

Joao-Dionisio: This is likely just a new ruff version flagging some errors. I'd just merge without caring much about this, but if you're offering to fix this, I'm more than happy to accept :)

I'll take a little bit to review, but everything seems to be in order. Can you please make sure that the `optimized_model()` manages to find primal solutions?

adamj34: I checked that `optmized_model()` wasn't able to find any primal solution, so I 've increased the node limit parameter and now it finds 1.

Regarding the `X | Y` syntax, I only changed it in `scip.pyi` to pass the pipeline checks and because this syntax is already used there.  

I was thinking about replacing `typing.Union` in the entire repo, but in general such fixes are considered unsafe for Python versions prior to 3.10 (see for reference: https://docs.astral.sh/ruff/rules/non-pep604-annotation-union/) and `pyproject.toml` only requires `python>=3.8`. Maybe there's no point in touching it until the project moves to `python>=3.10`?

Joao-Dionisio: Yes, we tend to prefer supporting as many versions as possible, and 3.10 is not thaaat old yet.

src/pyscipopt/scip.pxi
```diff
@@ -3462,6 +3507,28 @@ cdef class Model:
...
+        Returns
+        -------
+        int
+
+        """
+        return SCIPgetMaxTotalDepth(self._scip)
+
+    def getNBacktracks(self):
+        """
+        Gets total number of backtracks, i.e. number of times, the new node was selected from the leaves queue.
```
```suggestion
        Gets total number of backtracks, i.e., number of times the new node was selected from the leaves queue.
```

tests/test_statistics.py
```diff
@@ -41,9 +48,84 @@ def test_getNReoptRuns(optimized_model):
...
+    leaves, children, siblings = optimized_model.getOpenNodes()
+    open_nodes = leaves + children + siblings
+    manual_avg_lowerbound = 0.0
+    if len(open_nodes) > 0:
+        manual_avg_lowerbound = np.mean(
+            [node.getLowerbound() for node in open_nodes] + [optimized_model.getFocusNode().getLowerbound()]
+        )
+
+    assert isinstance(avg_lowerbound, float)
+    assert manual_avg_lowerbound == pytest.approx(avg_lowerbound)
```
Joao-Dionisio:
```suggestion
    assert optimized_model.isEQ( manual_avg_lowerbound, avg_lowerbound)
```

tests/test_statistics.py
```diff
@@ -41,9 +48,84 @@ def test_getNReoptRuns(optimized_model):
...
+
+
+def test_getAvgDualbound(optimized_model):
+    avg_dualbound = optimized_model.getAvgDualbound()
+    avg_lowerbound = optimized_model.getAvgLowerbound()
+
+    assert isinstance(avg_dualbound, float)
+    assert avg_dualbound == pytest.approx(avg_lowerbound) or avg_dualbound == pytest.approx(-avg_lowerbound)
```
Joao-Dionisio:
```suggestion
    assert optimized_model.isEQ( avg_dualbound, avg_lowerbound) or optimized_model.isEQ( avg_dualbound, -avg_lowerbound)
```

src/pyscipopt/scip.pxi
```diff
@@ -3285,6 +3285,28 @@ cdef class Model:
         """
         return SCIPgetPresolvingTime(self._scip)
 
+    def getDeterministicTime(self):
```
Joao-Dionisio, line 3288: I had absolutely no clue this existed, and it would have been useful at some point 😁

tests/test_statistics.py
```diff
@@ -41,9 +48,84 @@ def test_getNReoptRuns(optimized_model):
...
+    assert isinstance(det_time, float)
+    assert det_time >= 0.0
+
+
+def test_getUpperbound(optimized_model):
+    upperbound = optimized_model.getUpperbound()
+    lowerbound = optimized_model.getLowerbound()
+
+    assert isinstance(upperbound, float)
+    assert upperbound >= lowerbound
```
Joao-Dionisio:
```suggestion
    assert optimized_model.isGE( upperbound, lowerbound )
```

tests/test_statistics.py
```diff
@@ -41,9 +48,84 @@ def test_getNReoptRuns(optimized_model):
...
+
+
+def test_getFirstPrimalBound(optimized_model):
+    first_primal = optimized_model.getFirstPrimalBound()
+    upperbound = optimized_model.getUpperbound()
+
+    assert isinstance(first_primal, float)
+    assert first_primal >= upperbound
```
Joao-Dionisio:
```suggestion
    assert optimized_model.isGE( first_primal, upperbound )
```

I suppose that, in order for this to make sense, we'd need to ensure that the optimized model finds more than one solution.

tests/test_statistics.py
```diff
@@ -41,9 +48,84 @@ def test_getNReoptRuns(optimized_model):
...
+
+
+def test_getLowerboundRoot(optimized_model):
+    lowerbound_root = optimized_model.getLowerboundRoot()
+    lowerbound = optimized_model.getLowerbound()
+
+    assert isinstance(lowerbound_root, float)
+    assert lowerbound_root <= lowerbound
```
Joao-Dionisio:
```suggestion
    assert optimized_model.isLE( lowerbound_root, lowerbound )
```

CHANGELOG.md
```diff
@@ -2,7 +2,8 @@
 
 ## Unreleased
 ### Added
-- Added methods: `getNNodesLeft()`, `getNRuns()`, `getNReoptRuns()`, `addNNodes()` with tests
+- Added methods: `getNNodesLeft()`, `getNRuns()`, `getNReoptRuns()`, `addNNodes()`, `getDeterministicTime()`, `getAvgDualbound()`, `getMaxTotalDepth()`, `getNBacktracks()`,\
+`getFocusNode()`, `getAvgLowerbound()`, `getFirstPrimalBound()`, `getLowerboundRoot()`, `getUpperbound()`, `getNObjlimLeaves()` with tests
```
Joao-Dionisio:
```suggestion
- Added the following methods with tests: `getNNodesLeft()`, `getNRuns()`, `getNReoptRuns()`, `addNNodes()`, `getDeterministicTime()`, `getAvgDualbound()`, `getMaxTotalDepth()`, `getNBacktracks()`,\
`getFocusNode()`, `getAvgLowerbound()`, `getFirstPrimalBound()`, `getLowerboundRoot()`, `getUpperbound()`, `getNObjlimLeaves()`
```

Joao-Dionisio:
Sorry for the delay @adamj34 ! I'm not as free in September as I was expecting to be, but I'll still manage to take a closer look.

I'm not sure what the code coverage bot is complaining about (am AFK), but it'd be nice to ensure that everything is tested.

src/pyscipopt/scip.pxi
adamj34: ;)

tests/test_statistics.py
adamj34: Good point. I changed it to a strict inequality (m.GT instead of m.GE) to make sure it makes sense.

tests/test_statistics.py
```diff
@@ -7,11 +8,34 @@
...
+    # finds 1 primal solution
+    model = random_mip_1(small=True, node_lim=2400)
     model.optimize()
     return model
 
 
+# very simple model with 2 primal solutions
+# created because getting more than 1 primal solution from random_mip_1 requres setting a large node limit, which slows down the tests
+@pytest.fixture
+def optimized_model_with_primal_solutions():
```
Joao-Dionisio: There's no guarantee that this will remain working for future SCIP versions, but I suppose the same would be true for every model.

tests/test_statistics.py
```diff
@@ -41,9 +76,88 @@ def test_getNReoptRuns(optimized_model):
...
+    assert isinstance(upperbound, float)
+    assert optimized_model.isGE(upperbound, lowerbound)
+
+
+def test_getFirstPrimalBound(optimized_model_with_primal_solutions):
+    first_primal = optimized_model_with_primal_solutions.getFirstPrimalBound()
+    upperbound = optimized_model_with_primal_solutions.getUpperbound()
+
+    assert isinstance(first_primal, float)
+    assert optimized_model_with_primal_solutions.isGT(first_primal, upperbound)
```
Joao-Dionisio: One is in the original space and the other in the transformed space, right?

tests/test_statistics.py
```diff
@@ -1,16 +1,41 @@
...
-    model = random_mip_1(small=True)  # Using small=True for speed across tests
+    # Using small=True for speed across tests
+    # finds 1 primal solution
+    model = random_mip_1(small=True, node_lim=2400)
     model.optimize()
     return model
 
 
+# very simple model with 2 primal solutions
+# created because getting more than 1 primal solution from random_mip_1 requres setting a large node limit, which slows down the tests
```
Joao-Dionisio:
```suggestion
# created because getting more than 1 primal solution from random_mip_1 requires setting a large node limit, which slows down the tests
```

src/pyscipopt/scip.pxi
```diff
@@ -3285,6 +3285,28 @@ cdef class Model:
...
+        Returns
+        -------
+        float
+
+        """
+        return SCIPgetDeterministicTime(self._scip)
+
+    def getFirstPrimalBound(self):
+        """
+        Gets the primal bound of the very first solution.
```
```suggestion
        Gets the primal bound of the very first solution in the original space.
```

tests/test_statistics.py
```diff
@@ -41,9 +76,88 @@ def test_getNReoptRuns(optimized_model):
...
+    assert isinstance(n_backtracks, int)
+    assert n_backtracks >= 0
+
+
+def test_getAvgLowerbound(optimized_model):
+    avg_lowerbound = optimized_model.getAvgLowerbound()
+    leaves, children, siblings = optimized_model.getOpenNodes()
+    open_nodes = leaves + children + siblings
+    manual_avg_lowerbound = 0.0
+    if len(open_nodes) > 0:
```
Joao-Dionisio: This is 0 if the model is solved to optimality, so I suppose we should assert that it isn't.

tests/test_statistics.py
```diff
@@ -41,9 +76,88 @@ def test_getNReoptRuns(optimized_model):
...
+            [node.getLowerbound() for node in open_nodes]
+            + [optimized_model.getFocusNode().getLowerbound()]
+        )
+
+    assert isinstance(avg_lowerbound, float)
+    assert optimized_model.isEQ(manual_avg_lowerbound, avg_lowerbound)
+
+
+def test_getAvgDualbound(optimized_model):
+    avg_dualbound = optimized_model.getAvgDualbound()
```
Joao-Dionisio, line 155: This has the same original vs transformed space fragility. But, I suppose it's not a big deal, it's a minimization problem, and any SCIP changes to the way it handles this will error out very loudly everywhere.

tests/test_statistics.py
adamj34: Yes, `getUpperbound()` is in the transformed space and `getFirstPrimalBound()` is in the original space. However, I've refactored this test a bit and no longer use `getUpperbound()`.

tests/test_statistics.py
adamj34: I deleted the if statement.

tests/test_statistics.py
adamj34: While I'm not sure how likely it was to break in the future, I think it would be better to have something more reliable than assuming that an _x_ number of node limit will produce _y_ number of solutions.

I came up with choosing a simple model and providing solutions by hand, so that it's easy to reason about all possible solutions. I also added a few comments in the tests to make them clearer.

Please let me know if this approach is better.

tests/test_statistics.py
adamj34: I think that for a minimization problem `getAvgDualbound()` and `getAvgLowerbound()` will be exactly the same.
For a maximization problem, one will be the negative of the other. That's why I included `or` in the assert statement.

adamj34: We could test it similarly to `test_getAvgLowerbound` but there's no `getDualbound()` method defined on Node. This is the reason behind using a comparison to `getAvgLowerbound` for testing function.

adamj34:
Thank you for your suggestions and comments @Joao-Dionisio!

I believe I've addressed all of them. If there's anything that requires further improvement, then let me know.

Regarding the Codecov complaints, I double checked if all methods that I added were tested. It seems that `getFocusNode` was only indirectly tested in `test_getAvgLowerbound`, so I added a dedicated test for it in [e1caacd](https://github.com/scipopt/PySCIPOpt/pull/1237/commits/e1caacde2c29d599aa8b495180aa0d1ecbb3bb4a). Beyond that, I think that all methods have an accompanying test.

PS I hope that these reviews aren't taking too much time away from your PhD 😊

tests/test_statistics.py
Joao-Dionisio: If the model is solved to optimality, `getFocusNode()` will return `None` and the test will raise an error. I think a pre-condition that the model is not solved to optimality would be better

Joao-Dionisio:
Thank you! We're really in the final stages, less than a month to go, thankfully 🙏

tests/test_statistics.py
Joao-Dionisio: > I think that for a minimization problem getAvgDualbound() and getAvgLowerbound() will be exactly the same.

The sign will be the same, but SCIP may introduce an objective offset and objective scale, for example. But, without presolving, I suppose that's not the case.

tests/test_statistics.py
```diff
@@ -1,16 +1,49 @@
+from pyscipopt.scip import Model
```
Joao-Dionisio, line 1: Why not import from `pyscipopt` directly?

adamj34: You're right, I somehow overlooked that. Fixed in [be8cb16](https://github.com/scipopt/PySCIPOpt/pull/1237/commits/be8cb165e1aba9d81a5c5e809f4be8649daf4759)

tests/test_statistics.py
adamj34: Yes, presolving is off for `random_mip_1` which is used for this test

tests/test_statistics.py
adamj34: I refactored this test so it now handles both models solved to optimality and those not solved to optimality.

Joao-Dionisio (approved):
Thank you @adamj34 , and sorry for the long review process! Keep 'em coming 😁

PR #1241, opened 2026-08-07
Fix type stub lint for Ruff 0.16

MySweetEden:
## Summary

- Replace the remaining `Union[Expr, GenExpr]` annotations in `scip.pyi` with `Expr | GenExpr`
- Remove the now-unused `Union` import

## Why

The type-stub lint workflow installs the latest Ruff release. Since Ruff 0.16, the existing annotations are reported as `UP007` violations, causing the lint check to fail on unrelated pull requests.

This is a typing-only syntax update and does not change runtime behavior.


Joao-Dionisio (approved):
Thank you!

PR #1246, opened 2026-08-10
Fix failing structured trace handler reuse test on master

MySweetEden:
**Summary**
Fix `test_structured_optimization_trace_reuses_handler_for_repeated_contexts`, which is currently failing on `master`.

On the Windows runner, the one-second time limit could be exhausted during presolving, before any trace event was emitted. This left the trace with only a `run_end` record and caused the test to fail.

This PR:
- disables presolving in the structured trace handler reuse test
- retains the five-second optimization limit configured by `_model()`

**Why this was not caught before merge**
The failure was detected only after merging because the `Integration test` workflow runs on pushes to `master`, but not on pull requests.

**Follow-up**
I propose tracking the pre-merge CI coverage gap in a separate issue, including whether the `Integration test` workflow should run before merge. This PR leaves the CI configuration unchanged and is limited to fixing the failing test.

### DIFF
--- a/tests/test_recipe_structured_optimization_trace.py
+++ b/tests/test_recipe_structured_optimization_trace.py
@@ -3,6 +3,7 @@ import json
 import pytest
 from helpers.utils import bin_packing_model
 
+from pyscipopt import SCIP_PARAMSETTING
 from pyscipopt.recipes.structured_optimization_trace import (
     attach_structured_optimization_trace,
     structured_optimization_trace,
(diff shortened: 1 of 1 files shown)


Joao-Dionisio:
Hey @MySweetEden , thank you!

> I propose tracking the pre-merge CI coverage gap in a separate issue, including whether the Integration test workflow should run before merge.

On this, it's a good idea. One requirement would be that it doesn't run every push, to avoid wasting resources, but only at merge time.

Joao-Dionisio (approved):

PR #1251, opened 2026-09-06
Implement missing test cases for `isActive` and `getRowLinear` methods

adamj34:
Resolve TODOs for implementing or extending testing for `isActive` and `getRowLinear` methods.

Update `CHANGELOG.md`.

### DIFF
--- a/tests/test_cons.py
+++ b/tests/test_cons.py
@@ -404,7 +404,7 @@ def test_getRowLinear_for_linear_constraints():
     class RowAddedEvent(Eventhdlr):
         def __init__(self, cons):
             super().__init__()
-            self.cons = cons
+            self.cons_by_name = {con.name: con for con in cons}
             self.matched = {con: False for con in cons}
 
         def eventinit(self):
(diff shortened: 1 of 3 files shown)


Joao-Dionisio:
Hey @adamj34 , thank you so much! Just a small comment, but this looks good :)

tests/test_cons.py
```diff
@@ -398,6 +399,74 @@ def test_getValsLinear():
...
+            self.model.catchEvent(SCIP_EVENTTYPE.ROWADDEDLP, self)
+
+        def eventexit(self):
+            self.model.dropEvent(SCIP_EVENTTYPE.ROWADDEDLP, self)
+
+        def eventexec(self, event):
+            row = event.getRow()
+
+            for con in self.cons:
+                try:
```
Joao-Dionisio: I'd rather not have a try/except here. Not a big fan of these, in general. I believe that a row's name shares the corresponding constraint's name, so at `__init__()` you could record something like `self.cons_by_name = {con.name: con for con in cons}`, and then in `eventexec()` do `trans_cons = self.model.getTransformedCons(self.cons_by_name[row.name])`.

Joao-Dionisio: If you're fine with this, and it ends up working, then I'm all for merging!

Joao-Dionisio (changes requested):

tests/test_cons.py
adamj34: Thanks. I've refactored it. Is that what you had in mind?

Joao-Dionisio (approved):

tests/test_cons.py
Joao-Dionisio: Yep, thank you so much!!

PR #1252, opened 2026-09-08
Fix lotsizing_lazy example: lock variables in conslock

Joao-Dionisio:
Supersedes #1228. @abb-omidi , do you mind taking a look?


abb-omidi:
@Joao-Dionisio, do I have to check something?

Joao-Dionisio:
@abb-omidi you don't have to, of course, just if you have the time :)

You can see [here](https://github.com/scipopt/PySCIPOpt/pull/1252/changes) the changes I made, essentially adding variable locks.

PR #1245, opened 2026-08-09
Extension on examples

abb-omidi:
1) An extension on the heuristic callback
```
from pyscipopt import Model, quicksum, SCIP_PARAMSETTING, Heur, SCIP_RESULT, SCIP_HEURTIMING
import matplotlib.pyplot as plt
import matplotlib.patches as patches
import seaborn as sns

# =============================================================================
# Model data
# =============================================================================

def data_():
    p = [4, 9, 7, 5, 5, 4, 
         5, 9, 3, 7, 10, 7, 
         10, 9, 4, 8, 4, 5, 
         8, 6, 5, 5, 4, 4, 7, 
         9, 4, 8, 9, 5, 3, 7, 
        #  4, 9, 5, 
        #  5, 7, 8, 8, 6, 6, 3, 
        #  5, 3, 5, 4, 8, 7, 9, 5
         ]

    r = [61, 59, 0, 21, 6, 0, 
         57, 41, 0, 67, 0, 9, 
         40, 58, 0, 0, 0, 0, 
         0, 25, 0, 0, 5, 0, 22, 
         60, 0, 0, 61, 4, 0, 0, 
        #  24, 39, 0, 
        #  0, 24, 0, 
        #  0, 0, 0, 33, 28, 64, 
        #  31, 0, 2, 59, 0, 51
         ]

    assert(len(p) == len(r))
    m = 4
    nb_m = range(m)
    tasks = range(len(p))
    BigM = sum(p[j] for j in tasks) + max(r[j] for j in tasks)
    mapping = [(i, j) for i in tasks for j in tasks if i < j]
    mapping_2 = [(i, j) for i in tasks for j in tasks if i != j]
    mapping_3 = [(i, j, k) for i in tasks for j in tasks for k in tasks if i < j and j < k and i < k]
    T_horizen = range(sum(p[j] for j in tasks) + max(r[j] for j in tasks))
    return p, r, m, nb_m, tasks, BigM, mapping, mapping_2, mapping_3, T_horizen

p, r, m, nb_m, tasks, BigM, mapping, mapping_2, mapping_3, T_horizen = data_()

# =============================================================================
# The variables declaration 
# =============================================================================

def models_variables():
    model = Model()
    
    finish = {j: model.addVar(vtype="C", lb=0.0, name=f"finish[{j}]") for j in tasks}
    start  = {j: model.addVar(vtype="C", lb=0.0, name=f"start[{j}]")  for j in tasks}
[...]

Issue #1240, opened 2026-08-06
The strange behavior of the disjunction constraint


abb-omidi:
@Joao-Dionisio,

Hi,
I hope you are doing well. I tried to analyze the logical capability of the SCIP solver through the PySCIPOpt wrapper to solve a scheduling problem. Also, as an extension of question #1055. The logical constraint I was trying to run is of the form:

$$ (x_{i,m} \land x_{j,m}) \implies (start_{j} \geq finish_{i} \lor start_{i} \geq finish_{j}) $$

It says that if jobs $i$ and $j$ are being assigned to the same machine, they do not overlap. However, it is a series of $\lor$ clauses of the form:

$$ (\lnot x_{i,m}) \lor (\lnot x_{j,m}) \lor (start_{j} \geq finish_{i}) \lor (start_{i} \geq finish_{j}) $$

```
    for i, j in mapping:
        for m in nb_m:
            model.addConsDisjunction(
                [
                    z[i, m] == 0,
                    z[j, m] == 0,
                    finish[i] <= start[j],
                    finish[j] <= start[i],
                ],
                name=f"disj_{i}_{j}_{m}",
            )
```
As long as I was running the example with a few tasks/jobs, the model works well without throwing any error, while I extend the jobs, the model solves to optimality with the wrong solution!!!

```
# The optimal solution
SCIP Status        : problem is solved [optimal solution found]
Solving Time (sec) : 54.66
Solving Nodes      : 729
Primal Bound       : +7.40000000000000e+01 (20 solutions)
Dual Bound         : +7.40000000000000e+01
Gap                : 0.00 %
```

```
# The wrong solution
SCIP Status        : problem is solved [optimal solution found]
Solving Time (sec) : 2.06
Solving Nodes      : 2659 (total of 3555 nodes in 2 runs)
Primal Bound       : +8.10000000000000e+01 (1 solutions)
Dual Bound         : +8.10000000000000e+01
Gap                : 0.00 %
```
I am running `scip = 10.0.2`, and `pyscipopt = 6.2.1`.

If you need any further information, just let me know.
All the best
Abbas

[PMSP_Logical.txt](https://github.com/user-attachments/files/30775526/PMSP_Logical.txt)

abb-omidi:
Dear @Joao-Dionisio, 

A strange observation is that when I set the disjunction constraint attribute 'local' to 'True' the solver works with the normal behavior and every thing goes well. 

[...]

PR #1253, opened 2026-09-08
Add heuristic plugin example

Joao-Dionisio:
This supersedes #1245 . @abb-omidi , would you mind taking a look, please?


Joao-Dionisio:
The disjunction example suffers from the bug mentioned in #1240, so an eventual fix would need to be merged first.

Issue #977, opened 2025-04-22
Add support for all constraint types


Joao-Dionisio:
There are many constraint handlers that haven't been brought over to PySCIPOpt. It can be useful to have them when trying to query/alter transformed constraints, as mentioned in #941.

This issue will track the progress of interfacing all these methods. Some handlers may be left out, and it's not reasonable to expect all methods to be interfaced, let alone tested.

- [ ] AND
- [ ] benders
- [ ] benderslp
- [ ] bounddisjunction
- [ ] cardinality
- [ ] conjunction
- [ ] countsols
- [ ] cumulative
- [ ] disjunction
- [ ] indicator
- [ ] integral
- [ ] knapsack
- [ ] linear
- [ ] linking
- [ ] logicor
- [ ] nonlinear
- [ ] or
- [ ] orbisack
- [ ] orbitope
- [ ] pseudoboolean
- [ ] setppc
- [ ] sos1
- [ ] sos2
- [ ] superindicator
- [ ] symresack
- [ ] varbound
- [ ] xor
- [ ] rpa
- [ ] optcumulative

Not marking anything yet because there are some methods that can be easily wrapped (e.g., `SCIPgetNVarsAnd`, even though there already is some partial support for AND constraints).

Please let us know in case you have some feedback or would like to contribute. Most of the work should be easy, there's just a lot of it.

Joao-Dionisio:
Right now, for many things I'm mostly interfacing the method from the corresponding handler. In the future, it would be nice to have general methods, when applicable.

PR #1255, opened 2026-09-10
Interface SCIP logicor constraints

mohitt31:
Adds the logicor constraint handler to the Python interface. It is one of the unchecked items in #977.

Methods added:

- `addConsLogicor(vars, ...)` creates the constraint
- `addCoefLogicor(cons, var)` adds a variable to an existing one
- `getNVarsLogicor(cons)` and `getVarsLogicor(cons)`
- `getDualsolLogicor(cons)` and `getDualfarkasLogicor(cons)`

They follow the same shape as the existing AND and OR handlers, so the keyword arguments and their defaults match `addConsAnd`.

Tested with SCIP 10.0.3 on macOS (arm64). The suite goes from 449 to 450 passed with the same skips, and `stubs/test.sh` reports no issues against the stubs I added.

One thing I ran into while writing the handler check in `getVarsLogicor`. The equivalent line in `getVarsAnd` (`scip.pxi:6956`) is

```python
assert(constype == 'and', "The constraint handler %s does not have this functionality." % constype)
```

That asserts a two element tuple, which is always truthy, so the guard never actually fires. I wrote mine without the parentheses. I left the AND one alone to keep this PR to one thing, but happy to send a separate fix if you want it.


Joao-Dionisio:
Hello @mohitt31 , thank you so much! I'll only be able to take a look at this over the next week, I believe, but it's looking good!

src/pyscipopt/scip.pxi
```diff
@@ -7013,6 +7013,103 @@ cdef class Model:
...
+        logicor_cons : Constraint
+            logicor constraint to get the dual solution from.
+
+        Returns
+        -------
+        float
+
+        """
+
+        return SCIPgetDualsolLogicor(self._scip, logicor_cons.scip_cons)
```
Joao-Dionisio: I believe this needs the transformed constraint, same with the farkas below.

Joao-Dionisio:
> happy to send a separate fix if you want it.

I would also be happy if you did it, if you don't mind 😄 
We have another issue in `getDualSolKnapsack()`:

```python
        if cons.isOriginal():
            transcons = cons
        else:
            transcons = <Constraint>self.getTransformedCons(cons)
```

This is the wrong way round, of course, and is also present in the other farkas function. Would you mind also fixing this in that separate fix?

src/pyscipopt/scip.pxi
mohitt31: You're right, thanks. Fixed in 559e836. Both now look up the transformed constraint when they're given an original one, the same way getDualsolLinear does, and they raise if the constraint isn't a logicor. I also added a test that calls them on the original constraint after solving.

mohitt31:
Sure, I'll do it. I checked and getDualfarkasKnapsack has the same problem, it passes the original constraint straight to SCIP. So the separate PR will fix both knapsack functions and the getVarsAnd assert, with tests that call them on original constraints. I'll open it in the next day or two.

Joao-Dionisio (approved):
Thank you @mohitt31 ! I'll take a look at your other PR later, maybe this week.

PR #1257, opened 2026-09-16
Fix SCIP warnings caused by parallel test execution

adamj34:
## Summary
Fixes SCIP warnings that appear when running the test suite with parallel execution (`pytest -nauto`).
Updates CHANGELOG.md.

## Before:
When using `pytest -nauto`:
par-testing-nauto-before

It can be seen on the screenshot above that warnings originate from `flp-subprob`, which is defined inside `testBenders` in `tests/test_customizedbenders.py`. `testBenders` is later used by two tests.

However, the same warnings are not present when just using `pytest`:
par-testin-seq-before

## Fix
Manually free the subproblem instance of `testBenders`.

## After:
After the fix,  `pytest -nauto`  no longer produces any warnings:
par-testing-nauto-after


Joao-Dionisio:
Hey @adamj34 , thank you! The end result is pretty useful, but it feels strange to ask a Python user to free a problem. This might be an issue on the SCIP side, I'll take a look later.

adamj34:
Sure @Joao-Dionisio, take your time, I don't think it's very urgent.

It's pretty weird that these warnings only appear when the tests are executed in parallel but they are absent when the tests are executed sequentially, so I guess that the real issue lies somewhere else than the tests themselves.

